PR #3, opened 2017-01-25
Separate stainless-core from front-ends

gsps:
This is a proposal of how to factor the various compiler-specific front-ends out of stainless. The goal is to allow more fine-grained dependency management, and, in particular, make "stainless-core" independent of scalac and dotty. It will also make it easier to interface with stainless in the context of implementing predicate-qualified types in dotty.

The proposed solution is to split stainless up into multiple sbt subprojects. There is "stainless-core", which essentially contains everything but /frontends/* and Main. Since Main must refer to a concrete front-end, we generate it from sbt for each of the two frontend-specific subprojects: "stainless-scalac" and "stainless-dotty". The integration test suite is another places where we depend on this entry-point. As an additional goodie of the refactoring we now have a simple way of executing the integration tests with either front-end.

Some things left to be done:
- [ ] Move LibraryLookupSuite into integration tests, since they depend on a front-end (currently commented out)
- [x] Customize generation of the `stainless` script in build.sbt.
- [ ] Resolve TODO about reenabling site.settings / doc support in build.sbt.
- [ ] Resolve FIXME about possible override in build.sbt
- [x] Clean up build.sbt. There might be a few more settings that can be moved from commonSettings to commonFrontendSettings.
- [ ] Fix the Dotty extraction. The integration tests currently fail with it.


samarion:
Could we maybe get the integration tests running (without the dotty ones for now) and merge this? I have some other features/fixes that are ready and I'd like to have the new directory structure before pushing.

Thanks!

PR #15, opened 2017-03-15
Fixes to compile (but not extract) the stainless standard library using stainless-dotty

gsps:
This commit introduces Dotty as a proper maven dependency, rather than a source dependency, and bumps the version to one that has some concrete issues with the stainless standard library fixed. We also introduce a workaround to provide Dotty with a classpath that includes dotty-library and scala-library, which it requires to run.


samarion:
Merged, thanks!

PR #22, opened 2017-04-04
Re-adding  missing non-negativity check for measure. Not sure  why it was removed.

ravimad:


samarion:
It's discharged through an Assertion injection in the lowering phase of the TerminationComponent

ravimad:
I see. The positivity check didn't work on the examples I tried. So I re-added it in the termination phase.

samarion:
It only gets checked during the verification phase.

ravimad:
Looks like then it is needed in the termination check phase as well. Conceptually, its a part of checking measures.

samarion:
Yes but termination checking is only valid if all verification checks pass as well...

ravimad:
Yeah but checking contracts separately is a reasonable separation. But deferring a part of measures check to verification phase is tricky

PR #33, opened 2017-05-17
Improvements for BitVectors + Tuples

mantognini:
Requires epfl-lara/inox#26.

 - The scalac frontend should be decently stable and capable of extracting operations on `Byte`, `Short`, `Int` and `Long` as well as the bitvector arithmetic and the explicit/implicit conversion/promotion between those types.

 - The codegen module should be capable of handle correctly `BitVector` now as a few inconsistencies were fixed and support for more operations was added. The `Modulo` operator was made consistant with smt solvers semantics.

 - The extraction of Tuple should now be general enough to extract tuple of arbitrary arity, with both frontends.

 - Extraction can now emit debug information when `--debug=extraction` is present on the command line.

 - This PR also makes stainless exit gracefully on a few exceptional scenarios.

A few things are not 100% perfect yet, feel free to point out where things look not to your taste. ;-)

What is missing/can be improved:
 - [x] Support for single-bit vector (aka `Boolean`) and the `& | ^` operation (from an extraction point of view).
 - [x] Fix for the dotty frontend: it currently fails to understand expression such as `b.toInt`.
 - [x] Unit tests for frontends are currently duplicated, which is not nice to say the least.
 - [x] The `TupleExtrationSuite` for dotty seems extremely slow, not sure which component is responsible for that. Will investigate.
 - [x] Positions should be set as often as possible.
 - [ ] When Inox is updated, bump sha in `build.sbt`.


mantognini:
I think now only `build.sbt` needs to be updated once epfl-lara/inox#26 is merged and this PR should then be ready too.

samarion:
Merged, thanks!

PR #43, opened 2017-06-06
export reports to JSON

mantognini:
Add a `--json` option. This is useful to produce HTML rendering of the reports.


samarion:
Merged, thanks!

PR #44, opened 2017-06-07
Dontcheck construct and Inlined functions

jad-hamza:
This commit adds a Dontcheck class. Expressions inside this class are ignored by the generation of verification conditions. When inlining a function, we add an assertion that the precondition holds, which will generate a verification condition. Then, we inline the code using the Dontcheck class, so that the code of the inlined function does not generate duplicate verification conditions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ tmp
 .idea
 Leon.eml
 Leon.iml
+*.iml
 /.idea_modules
 
 #scripts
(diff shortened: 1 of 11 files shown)


jad-hamza:
Some Dontcheck cases were missing in CodeGeneration.scala, should be ok now

jad-hamza:
A case was missing in the RecursiveEvaluator, the tests go through now.

samarion:
Merged, thanks!

PR #48, opened 2017-06-09
Add --strictarithmetic

mantognini:
Import epfl-lara/leon#273, except for C-specific check (i.e. `MinValue % -1`).

Changes to have a `Context` available for calling `weakestPrecondition` should be minimal I think. This best as is, if we want to change things more fundamentally later.

### DIFF
--- a/core/src/main/scala/stainless/verification/AssertionInjector.scala
+++ b/core/src/main/scala/stainless/verification/AssertionInjector.scala
@@ -3,11 +3,18 @@
 package stainless
 package verification
 
+/**
+ * Transform trees by inserting assertions. Those verify that all array access are valid,
+ * casts are legal, no division by zero occur and, when using the [[strictArithmetic]] mode,
+ * that the program is exempt of integer overflow and unexpected behaviour.
+ */
 trait AssertionInjector extends ast.TreeTransformer {
   val s: ast.Trees
   val t: ast.Trees
 
   implicit val symbols: s.Symbols
+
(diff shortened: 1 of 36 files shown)


core/src/main/scala/stainless/verification/AssertionInjector.scala
```diff
@@ -69,11 +144,62 @@ trait AssertionInjector extends ast.TreeTransformer {
...
 }
 
 object AssertionInjector {
+
+  /**
+   * Strict Arithmetmetic Mode:
+   *
+   * Add assertion for integer overflow checking and other unexpected behaviour (e.g. x << 65).
+   */
+  val optStrictkArithmetic = inox.FlagOptionDef("strictarithmetic", false)
```
samarion: I'd prefer if this was defined outside the AssertionInjector object since it's conceptually a generic property of verification (and not really linked to assertion injection). You can put it in VerificationComponent.scala or leave it in this file, whichever is fine by me.

Also, I think it should be `optStrictArithmetic` (without the k at the end of strict), no?

mantognini: I agree, I'll make the changes. And yes, this is a typo. Thanks for spotting it! :-)

mantognini:
Edited and rebased. ;-)

samarion:
LGTM, can you push to master?

mantognini:
Yes, it works! :-)

PR #49, opened 2017-06-16
Performance improvement for extraction

mantognini:
Continuation of epfl-lara/inox#28 and epfl-lara/inox#29. The speedup this time is much more modest as the real hotspot was in inox's extractors.

Before merging this:
 - [ ] epfl-lara/inox#29 should be merged;
 - [ ] a new version of inox should be published;
 - [ ] `build.sbt` should reflect the change.

### DIFF
--- a/core/src/main/scala/stainless/ast/Extractors.scala
+++ b/core/src/main/scala/stainless/ast/Extractors.scala
@@ -34,7 +34,7 @@ trait TreeDeconstructor extends inox.ast.TreeDeconstructor {
       })
   }
 
-  override def deconstruct(expr: s.Expr): (Seq[s.Variable], Seq[s.Expr], Seq[s.Type], (Seq[t.Variable], Seq[t.Expr], Seq[t.Type]) => t.Expr) = expr match {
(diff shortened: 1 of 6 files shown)


mantognini:
The more I think about these PRs, the more I think we should simplify things and go back to only re-ordering the pattern matching like it was [then](https://github.com/epfl-lara/inox/pull/28/commits/1bf0fe3a0a70b9297a56fd22261d90e87a57cf52), and that for mainly two reasons: maintainability is easier with the simpler implementation and the fact that with the introduction of caching and all the other changes the speedup might not be that relevant to motivate such an intrusive optimisation. We can always re-study the need for such (ugly) tricks if needed.

samarion:
I mostly agree, yes. I think we can keep the version in Inox as is since we don't touch the Inox trees much anymore and the optimization is self-contained (doesn't affect the API). If you feel strongly enough about it, feel free to open a PR on Inox that reverts the optimization.

mantognini:
No it's not worth spending too much time on this right now. Maybe later I'll open a PR. ;-)

PR #50, opened 2017-06-18
Typo

y-taka-23:


samarion:
This was fixed in https://github.com/epfl-lara/stainless/pull/42, but thanks for the contribution!

PR #51, opened 2017-06-26
Fix the termination metrics of the 91 function

y-taka-23:
Stainless rejects the `decreases` contract because `101 - n` can be negative, like:
```
...
[  Info  ]  - Now considering 'body assertion' VC for M @14:30...
[Warning ]  => INVALID
[Warning ] Found counter-example:
[Warning ]   n: BigInt -> 102
...
```
or, with `--termination`,
```
...
[Warning ]  ==> INVALID: measure is not well-founded in M
[  Info  ] Result for M
[  Info  ]  => BROKEN (Decreases Processor) -> DecreasesFailed
...
```
So I'm afraid that we need to ensure the metrics non-negative, with `stainless.math.max`.

(stainless revision: f9aff1428512669ccd431b827a501094a45a62aa)

### DIFF
--- a/core/src/sphinx/intro.rst
+++ b/core/src/sphinx/intro.rst
@@ -126,6 +126,6 @@ can be shown terminating as follows:
 .. code-block:: scala
 
   def M(n: BigInt): BigInt = {
-    decreases(max(101 - n, 0))
+    decreases(stainless.math.max(101 - n, 0))
     if (n > 100) n - 10 else M(M(n + 11))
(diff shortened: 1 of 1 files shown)


samarion:
Awesome, nice catch!
Could you please make the call to `stainless.math.max` explicit in the
example (or add a mention in the doc) so that users will know where the
`max` function comes from?

Thanks!

y-taka-23:
Sure, done.

PR #52, opened 2017-07-05
Fixed Windows issues for c:\User repository

MikaelMayer:
The generation of two sources was buggy if the path contained \U (character escape in the program).
This pull request fixes it.


samarion:
LGTM, thanks!

PR #59, opened 2017-07-21
--watch & incremental verification

mantognini:
This PR is not yet fully ready but feedback/testing is welcome.

Here are the headlines:
 - [x] `--watch` is now also available for stainless;
 - [x] redesign of the interface with compiler for extraction (1)
 - [x] the extraction is optimised for multi-core computers (2, 3);
 - [x] warnings are emitted on non-sealed abstract class (4);
 - [x] when watching files, only updated functions should be re-verified (5)
   + [x] implemented for the scalac frontend (6),
   + [x] implemented for the dotty frontend (7);
 - [x] the termination callback should be implemented.
 - [x] additional test suites.
 - [x] some general overview of Stainless' pipeline in the sphinx doc.

Notes:

 1. A new frontend interface is defined. It allows plugging in an arbitrary compiler with a simple interface: [see `Frontend` & `FrontendFactory`](https://github.com/epfl-lara/stainless/blob/watch/core/src/main/scala/stainless/frontend/Frontend.scala). Through a system of [`CallBack`](https://github.com/epfl-lara/stainless/blob/watch/core/src/main/scala/stainless/frontend/CallBack.scala#L31-L41) the ASTs are processed and forwarded to the component.

 2. The [`CallBackWithRegistry`](https://github.com/epfl-lara/stainless/blob/watch/core/src/main/scala/stainless/frontend/CallBack.scala#L68-L179), which is [implemented by the verification callback](https://github.com/epfl-lara/stainless/blob/watch/core/src/main/scala/stainless/verification/VerificationCallBack.scala#L9-L29), manages through a [`Registry`](https://github.com/epfl-lara/stainless/blob/watch/core/src/main/scala/stainless/utils/Registry.scala) to send to components small, self-contained programs as soon as they are available from extraction. Those programs are then processed in parallel.

 3. The main algorithm used by the `Registry` to create self-contained sub-programs is based on [`IncrementalComputationalGraph`](https://github.com/epfl-lara/stainless/blob/watch/core/src/main/scala/stainless/utils/IncrementalComputationalGraph.scala): this general utility basically works as a dependency graph that can be in a partial state when not all nodes are defined. Additionally, when all the dependencies for a given node are available, the node is "computed" -- for the `Registry` it means that the sub-programs are generated and send to the `VerificationComponent`'s usual pipeline.

 1. Such warning are emitted because it can hinder the performance of the new system: for un-sealed class hierarchy, the system wait for the end of the extraction in order to build the proper dependencies for each class, dealing therefore with potentially overridden methods and the lot of subtleties that come with inheritance.

 4. In order to decide whether a function should be re-verified, the `Registry` uses a [canonical representation of types and functions](https://github.com/epfl-lara/stainless/blob/watch/core/src/main/scala/stainless/utils/CanonicalFormBuilder.scala) as illustrated [here](https://github.com/epfl-lara/stainless/blob/watch/core/src/main/scala/stainless/utils/Registry.scala#L120-L154).

 5. The mapping from scalac `Global#Symbol` to inox/stainless Identifier is often tricky when watching files, assuming we want to have the same `Identifier` for the same function across several run of the extraction. In most cases, it should work but don't be surprised if adding a class, import or function triggers the verification of unmodified functions.

 6. Because this is still work in progress, `--watch` with the dotty frontend will always re-verify everything.

Currently, this doesn't touch the actual tree transformation pipeline. Further optimisation can be carried out later to cache the result of transformation and the results of verification. Caching the latter is interesting to avoid re-verifying dependencies of a function that was updated while watching the file -- we just need to keep track of what was verified or not, and probably that the canonical form can be used at this stage too.

One can try to verify `frontends/benchmarks/verification/valid/*.scala` and export the verification report with `--json` to compare the results (and the speed) of both the master version and this branch. There should not be significant difference between the two outputs. Because many things are changed in this PR, I encourage you try a few programs and make sure the same VCs are verified with both branches. :-)


jad-hamza:
Exciting! 

I'm getting an exception when trying to run this branch on any file (e.g. an empty file containing only `object Empty { }`).

```java
Exception in thread "main" java.io.IOException: User limit of inotify watches reached
        at sun.nio.fs.LinuxWatchService$Poller.implRegister(LinuxWatchService.java:264)
        at sun.nio.fs.AbstractPoller.processRequests(AbstractPoller.java:260)
        at sun.nio.fs.LinuxWatchService$Poller.run(LinuxWatchService.java:329)
        at java.lang.Thread.run(Thread.java:748)
```

mantognini:
Thanks Jad for trying it out. Sadly the error you get is due to the underlying implementation of `java.nio.file.WatchService`. If you can, have a look at https://unix.stackexchange.com/questions/13751/kernel-inotify-watch-limit-reached. On mine, `max_user_watches` is 524288. Or maybe just reboot your machine? A process might be leaking resources..

jad-hamza:
Thanks, I don't get the exception anymore after reboot.

mantognini:
This PR should be stable now. More things can be done, of course, but they can be added in a new PR (or this one, I don't mind -- it's just a question of making this reviewable as it's already a big one).

Specifically, we can get inspiration from *Fine-Grained Caching of Verification Results* by Rustan, Leino and Wüstholz, although stainless/inox differ quite a bit from Boogie:

 - To measure performance improvement, we could implement a "session recorder" for `--watch`, to collect the required data but this would probably go at the bottom of the TODO list.
 - In order to further refine what gets sent to the pipeline on update, the `IncrementalComputationalGraph` could be updated with an additional "switch" allowing us to re-compute nodes only when some dependencies' postcondition are weakened (or put in other words: to not recompute a node whose dependencies' postconditions are identical or stronger).
 - This generalisation of the `ICG` might also pay off when we introduce an additional, fine-grained cache right before sending VCs to the solvers.

Orthogonally from that:

 - The reports system can be improved to display a full status between each run under `--watch`.
 - IDE support (e.g. VS Code / Scastie) can be worked on.

I'm also wondering if some tools (e.g. `IncrementalComputationalGraph`) should be moved out of stainless as they are generic enough to be used to solve completely unrelated problems.

mantognini:
Because the system is already complex enough I've added a `doc/` directory. Let me know if [the general view](https://github.com/epfl-lara/stainless/blob/8d4b4fd2e6f2a44c75d142b5507b4ec907c28120/doc/Stainless%20Pipeline.svg) is understandable for you guys.

vkuncak:
Does larabot check stainless pull requests? If yes, is larabot actually up after recent outage?

mantognini:
Larabot hasn't been restarted I think.

core/src/main/scala/stainless/utils/CanonicalFormBuilder.scala
```diff
@@ -0,0 +1,497 @@
...
+  private val mapping = MutableMap[Identifier, UID]()
+
+  /** This one gets filled as we traverse the AST. */
+  private val buffer = ArrayBuffer[Byte]()
+
+
+  /******************* Internal Helpers ***********************************************************/
+
+  /** Register a new, unique & fresh UID for the given [[id]]. */
+  private def registerId(id: Identifier): Unit = {
```
samarion: You can define this as:
```scala
val registerId: Identifier => Unit = {
  var nextId: UID = 0
  ...
}
```
to make sure that `nextId` is not accessible (just a note on style).

core/src/main/scala/stainless/utils/CanonicalFormBuilder.scala
```diff
@@ -0,0 +1,497 @@
...
+    case xt.Forall(vds, _) => vds foreach registerVD
+    case xt.Choose(vd, _) => registerVD(vd)
+    case xt.FunctionInvocation(id, _, _) => storeId(id)
+    case xt.CharLiteral(l) => storeChar(l)
+    case l @ xt.BVLiteral(_, size) => storeInt(size); storeBigInt(l.toBigInt)
+    case xt.IntegerLiteral(l) => storeBigInt(l)
+    case xt.FractionLiteral(n, d) => storeBigInt(n); storeBigInt(d)
+    case xt.BooleanLiteral(l) => storeBool(l)
+    case xt.StringLiteral(l) => storeString(l)
+    case xt.GenericValue(_, _) => ??? // FIXME what's this "id"?
```
samarion: Store the type parameter type and the BigInt value I'd say.

core/src/main/scala/stainless/utils/Registry.scala
```diff
@@ -0,0 +1,314 @@
...
+  /**
+   * To be called once every compilation unit were extracted.
+   */
+  def checkpoint(): Option[xt.Symbols] = {
+    val defaultRes = process(knownClasses.values.toSeq, Seq.empty)
+    val res = if (frozen) {
+      assert(defaultRes.isEmpty)
+      graph.unfreeze() map { case (cls, funs) => xt.NoSymbols.withClasses(cls.toSeq).withFunctions(funs.toSeq) }
+    } else {
+      frozen = true
```
samarion: When is frozen set to false?

doc/Stainless Pipeline.txt
```diff
@@ -0,0 +1,73 @@
+# http://sequencediagram.org/
```
samarion: Could you maybe format this into Sphinx documentation and move it to core/src/sphinx? Thanks!

frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
```diff
@@ -314,10 +305,18 @@ class CodeExtraction(inoxCtx: inox.Context, symbols: SymbolsContext)(implicit va
...
+      annotationsOf(sym) ++ isAbstract ++ isSealed
+    }
+
     val optInv = if (invariants.isEmpty) None else Some {
-      new xt.FunDef(SymbolIdentifier(invSymbol), Seq.empty, Seq.empty, xt.BooleanType,
+      val invId = cache fetchInvIdForClass sym
+      new xt.FunDef(invId, Seq.empty, Seq.empty, xt.BooleanType,
         if (invariants.size == 1) invariants.head else xt.And(invariants),
-        Set(xt.IsInvariant) ++ flags
+        Set(xt.IsInvariant) ++ flags - xt.IsSealed // FIXME if IsSealed is not removed, crash. Why???
```
samarion: I guess dotty adds sealed to the set of annotations of a sealed class. This should be fixed in ASTExtractors.getAnnotations I think (or annotationsOf maybe).

frontends/dotty/src/main/scala/stainless/frontends/dotc/SymbolsContext.scala
```diff
@@ -8,36 +8,48 @@ import dotty.tools.dotc.core.Symbols._
...
+  /** Get the identifier for the class invariant of [[sym]]. */
+  def fetchInvIdForClass(sym: Symbol)(implicit ctx: Context): SymbolIdentifier =
+    invs.getOrElseUpdate(fetch(sym), {
+      SymbolIdentifier(invSymbol)
+    })
+
+  private def getPath(sym: Symbol)(implicit ctx: Context): String = sym.fullName + sym.id.toString
+
+  /** Mapping from [[Symbol]] (or rather: its path) and the stainless identifier. */
+  private val s2i = MutableMap[String, SymbolIdentifier]()
```
samarion: This seems a bit dangerous to me since dotc might produce different symbols that share the same path. Do the symbols change that much? Maybe we should check with the LAMP guys about using something else than the dotc symbol as map key if it changes to often.

frontends/scalac/src/main/scala/stainless/frontends/scalac/ScalaCompiler.scala
```diff
@@ -3,11 +3,71 @@
...
+    SymbolIdentifier(symbol)
+  })
+
+  /** Get the identifier for the class invariant of [[sym]]. */
+  def fetchInvIdForClass(sym: Global#Symbol): SymbolIdentifier = invs.getOrElseUpdate(fetch(sym), {
+    SymbolIdentifier(invSymbol)
+  })
+
+  /** Mapping from [[Global#Symbol]] (or rather: its path) and the stainless identifier. */
+  private val s2i = MutableMap[String, SymbolIdentifier]()
```
samarion: Same remark as for the dotty compiler.

core/src/main/scala/stainless/utils/CanonicalFormBuilder.scala
mantognini: Ha! this is a nice trick. :-)

core/src/main/scala/stainless/utils/CanonicalFormBuilder.scala
mantognini: Will do.

core/src/main/scala/stainless/utils/Registry.scala
mantognini: When the `Registry` is initialised.

What happens is that during the first "updates-checkpoint" cycle the underlying graph is (expected to be) in a possibly incomplete but valid state. Then, upon calling `checkpoint` the graph can enter an invalid state (e.g. a function/type is removed from the user program but is still present for a while in the registry). In other words, the registry is unfrozen and directly re-frozen when calling `checkpoint` because it's only at this point we know for sure a definition was removed.

I'm still pondering whether this "frozen" property of the `Registry` should be made explicit (and public). It works for the current workflow but it could be made more flexible if this utility class is required somewhere else too.

doc/Stainless Pipeline.txt
mantognini: Definitely!

frontends/dotty/src/main/scala/stainless/frontends/dotc/SymbolsContext.scala
mantognini: The current implementation takes a safe approach by appending the (supposedly) unique id of the symbol at the end of the path to disambiguate things (e.g. overloaded functions). So I think it is safe as is.

However, I think they have a project for better incremental compilation.. I'll ask around because it might be a way to further reduce the "false positive" changes detected upon recompilation.

frontends/scalac/src/main/scala/stainless/frontends/scalac/ScalaCompiler.scala
mantognini: Here it's slightly more hacky, but the same applies: whenever two symbols have the same path they are unambiguously identified through their unique id.

core/src/main/scala/stainless/utils/Registry.scala
samarion: I don't understand, shouldn't `checkpoint` only be called at times when the state should be valid (i.e. when all compilation units have been seen)? Why is the first checkpoint different?

frontends/dotty/src/main/scala/stainless/frontends/dotc/SymbolsContext.scala
mantognini: The name is [sbt/zinc](https://github.com/sbt/zinc). Dunno yet if/how it can be integrated in stainless' frontend system easily.

core/src/main/scala/stainless/utils/CanonicalFormBuilder.scala
```diff
@@ -0,0 +1,497 @@
...
+
+  def apply(cd: xt.ClassDef): CanonicalForm = {
+    val impl = new CanonicalFormBuilderImpl
+    impl.build(cd)
+  }
+}
+
+/** Canonical form for functions, expressions, ... */
+class CanonicalForm(val bytes: Array[Byte]) {
+  override def hashCode: Int = bytes.hashCode
```
samarion: Are you sure this works? Strange that hashCode is valid but not equals...

frontends/dotty/src/main/scala/stainless/frontends/dotc/SymbolsContext.scala
samarion: Unfortunately, we might not be able to rely on Scala-style incremental compilation... Scala only needs to recompile if the interface of a function has changed (the types) but we have to run the checks as soon as any part of the implementation has been modified. Maybe zinc provides a good API to recompile only certain files, and we can maintain the dependency map ourselves and request new compilation runs.

core/src/main/scala/stainless/utils/Registry.scala
mantognini: `checkpoint` is called when the state is valid, yes. It then triggers the computation of what should be (re-)verified. The first time it is mainly used to trigger computation for things that require the whole program (e.g. class hierarchies for function overriding). But in every case (at the end of each compilation cycle), right after the call to `checkpoint` the `Registry` assumes it enters an invalid state (that's also why there's no "freeze" function in the API).

During the first cycle, nothing is removed from the graph so it is never invalid but only incomplete.

core/src/main/scala/stainless/utils/CanonicalFormBuilder.scala
mantognini: You're absolutely right! [ref](https://stackoverflow.com/a/744753/520217) I'll replace that by `Arrays.hashCode(bytes)`.

frontends/dotty/src/main/scala/stainless/frontends/dotc/SymbolsContext.scala
mantognini: yes, and it's not clear if we can plug into the compiler pipeline (I haven't got round to look into it properly TBH).

core/src/main/scala/stainless/utils/Registry.scala
samarion: Ohhh, I see, you're guarding the call to graph.unfreeze(). I was having trouble understanding what that `frozen` variable was doing...

samarion:
Simple (or not so simple in certain cases) caches should do the trick, but we should change the transformers API to make it obvious they're stateful and associated to a compiler run. This can happen in a later PR though.

About the normalization, I agree that your solution is probably more optimal but I'm concerned with how complicated it's going to be to add new trees without creating strange and unexpected cache misses... I don't really have a better design in mind except for directly normalizing the trees though.

mantognini:
Adding trees will make the CFB crash: the mapping in `CanonicalFormBuilderImpl` would be incomplete. Hence, we will get a good "warning" when running stainless that something is no longer in sync. :)

samarion:
I'm more worried about people not understanding the API and forgetting to
complete the store*Specifics methods.

mantognini:
Ah, I see. I'll add a quick note about that.

mantognini:
I've moved the doc to the sphinx one, updating some old `leon` instances.

mantognini:
Regarding the `CanonicalForm`: to be integrated easily with Scastie (which uses a load balancer in a way that two successive runs of stainless might be on two different server instance), it will be useful to save the cache (and therefore the CF) to disk for persistence. This feature can also be useful for stainless directly so I'll implement something for that (but in a later PR). This means that the byte format will actually be very practical for that as well. To make sense the verification reports should also be cached, so expect some change to the Report API as well.

samarion:
Okay, looks good to me! Have you run the regression tests and pushed everything you want before I merge? We need to restart larabot...

mantognini:
Everything should be in here. I'll re-run the tests once more, but if someone knows how to reboot larabot...

mantognini:
For some reasons I haven't yet understood CodeGenVerificationSuite fails on some test with a `inox.ast.TypeOps$TypeErrorException`. I let you know when I've identified and fixed the bug.

mantognini:
I'm getting a crazy on that one. Larabot was restarted and reports no failure. When running the full test suite on my local machine everything goes green. **BUT** if I run only part of the test suites, it fails with a really strange error. The selection of tests also impacts the outcome of the suite, for some reason. I set up a VM with a clean environment (some stable Ubuntu) and couldn't reproduce the issue at all. I tried cleaning all sbt/ivy caches, clean my java install.. no luck. Yet, I'm not able to rule out that the issue is related to my working environment.

A second pair of eyes on that one would help a lot. I've gathered here a few logs that might help: https://gist.github.com/mantognini/f1b11745fed0dfc8ceaaba00e9eaad32 My raw understanding is that the type parameter of the forall's inner lambda is not correct. Maybe there's something wrong with the jvm <-> stainless values conversion, but I couldn't find it. I tried also to look for potential global objects and caches that could go wrong during the evaluation of different and unrelated programs, but it looks like every evaluator is used once only per program.

First of all, can anyone reproduce the issue? And has any of you the slightest clue what's going on?

samarion:
Unfortunately, I can't reproduce the issue. The reason the issue is intermittent is probably due to randomness in the models returned by the solvers. In order to make it reproducible, you should output the model that leads to the evaluation failure so that we can reproduce the issue. I assume we have some issue with type parameter substitutions in codegen but these take place at many different places so it's not clear which substitution is at fault. If you can give me a model that breaks the evaluator, I'll try to fix the issue.

vkuncak:
What if one uses interpreter instead?
I feel this issue is orthogonal to what Marco did and he shouikd not be
responsible for fixing it?

samarion:
Indeed, I'll merge this as soon as my regression has passed.

However, since Marco is the only one able to reproduce the issue, it will be hard for anyone else to fix it without some extra information (such as the model I asked for, for example).

samarion:
Merged, thanks!

mantognini:
I updated the debug/codegen branch to reflect the state of stainless/inox which triggers the error on the evalutation of this [expression/model](https://gist.github.com/mantognini/f1b11745fed0dfc8ceaaba00e9eaad32#file-model-lists4) on my machine. With the master versions of inox/stainless, I can no longer reproduce the issue.

mantognini:
Quick note: this codegen bug still happens from time to time on larabot. With #68 I could restart the build enough times to make it pass the integration tests but obviously something wrong is going on. [The current build](http://laraquad4.epfl.ch:9000/epfl-lara/stainless/build/1869/12614) for #67 fails just the same way. This time I can only very rarely reproduce the bug on my machine, making the debugging process a total nightmare.

samarion:
After fixing a few unrelated issues in codegen, I believe I've finally found the one that was at fault here. The fix is in https://github.com/epfl-lara/stainless/commit/903040371139ddfc321d523e5ba0c83114127b68 and is related to string and integer sorting being different when the power of 10 changes (so pretty rare!)...

mantognini:
Seems to be working indeed! :+1:

PR #65, opened 2017-08-20
cache for verification conditions

jad-hamza:
This commit adds a verification cache to store VCs which have already been verified.
The cache is implemented in VerificationCache.scala, and extends the VerificationChecker trait.

Each VC is stored with its dependencies (a Program) as a String in the global `vccache` map.
The cache is also saved on disk in the file `.vccache.bin`, which is loaded at startup.

The dependencies are built with `buildDependencies` using a tree traversal.
The program is then transformed in canonical form, in the file Canonization.scala, using a tree 
transformer. The transformation starts by transforming the VC (by renaming all identifiers to `x`
and changing the global id to a new identifier, local to this transformation). It then transforms the 
function and ADT definitions encountered along the way.

I still need to add some documentation, and investigate why there are still some cache misses.
This pull request depends on https://github.com/epfl-lara/inox/pull/35 and https://github.com/epfl-lara/inox/pull/36.

### DIFF
--- a/core/src/main/scala/stainless/extraction/oo/MethodLifting.scala
+++ b/core/src/main/scala/stainless/extraction/oo/MethodLifting.scala
@@ -20,19 +20,14 @@ trait MethodLifting extends inox.ast.SymbolTransformer { self =>
       }
     }
 
-    val children: Map[Identifier, Set[Identifier]] =
+    val children: Map[Identifier, Seq[Identifier]] =
       symbols.classes.values
         .flatMap(cd => cd.parents.map(_ -> cd))
         .groupBy(_._1.id)
-        .mapValues(_.map(_._2.id).toSet)
-
-    val descendants: Map[Identifier, Set[Identifier]] =
-      inox.utils.fixpoint { (map: Map[Identifier, Set[Identifier]]) =>
-        map.map { case (p, desc) => p -> (desc ++ desc.flatMap(map.getOrElse(_, Set.empty))) }
-      } (children)
+        .mapValues(_.map(_._2.id).toSeq.sortBy(_.name))
 
     sealed trait Override { val cid: Identifier }
-    case class FunOverride(cid: Identifier, fid: Option[Identifier], children: Set[Override]) extends Override
+    case class FunOverride(cid: Identifier, fid: Option[Identifier], children: Seq[Override]) extends Override
     case class ValOverride(cid: Identifier, vd: ValDef) extends Override
 
     val (newSymbols, functionToOverrides): (Symbols, Map[Identifier, FunOverride]) = {
@@ -49,9 +44,9 @@ trait MethodLifting extends inox.ast.SymbolTransformer { self =>
 
       def rec(id: Identifier): Map[Symbol, Override] = {
         val cd = symbols.getClass(id)
-        val cids = children.getOrElse(id, Set.empty)
+        val cids = children.getOrElse(id, Seq.empty)
         val ctrees = if (cids.isEmpty) {
-          Set(cd.fields.flatMap(vd => firstSymbol(id, vd).map(_ -> ValOverride(id, vd))).toMap)
+          Seq(cd.fields.flatMap(vd => firstSymbol(id, vd).map(_ -> ValOverride(id, vd))).toMap)
         } else {
           cids.map(rec)
         }
@@ -60,7 +55,7 @@ trait MethodLifting extends inox.ast.SymbolTransformer { self =>
           fid.symbol -> FunOverride(id, Some(fid), ctrees.flatMap(_.get(fid.symbol)))
         }.toMap
 
-        val noOverrides = (ctrees.flatMap(_.keys.toSet) -- newOverrides.keys).map {
+        val noOverrides = ctrees.flatMap(_.keys.toSet).filterNot(newOverrides contains _).map {
           sym => sym -> FunOverride(id, None, ctrees.flatMap(_.get(sym)))
         }
 
@@ -130,7 +125,7 @@ trait MethodLifting extends inox.ast.SymbolTransformer { self =>
 
     object default extends BaseTransformer
 
-    def makeFunction(cid: Identifier, fid: Identifier, cos: Set[Override]): t.FunDef = {
+    def makeFunction(cid: Identifier, fid: Identifier, cos: Seq[Override]): t.FunDef = {
       val cd = newSymbols.getClass(cid)
       val fd = newSymbols.getFunction(fid)
       val tpSeq = newSymbols.freshenTypeParams(cd.typeArgs)
@@ -151,8 +146,8 @@ trait MethodLifting extends inox.ast.SymbolTransformer { self =>
         }
       }
 
-      val subCalls = (for (co <- cos.toSeq.sortBy(_.cid.name)) yield {
-        firstOverrides(co).toSeq.sortBy(_._1.name).map { case (cid, either) =>
+      val subCalls = (for (co <- cos) yield {
+        firstOverrides(co).map { case (cid, either) =>
           val descType = default.transform(tcd.descendants.find(_.id == cid).get.toType).asInstanceOf[t.ClassType]
           val thiss = t.AsInstanceOf(arg.toVariable, descType).copiedFrom(arg)
           (t.IsInstanceOf(arg.toVariable, descType).copiedFrom(arg), either match {
--- a/core/src/main/scala/stainless/extraction/oo/MethodLifting.scala
+++ b/core/src/main/scala/stainless/extraction/oo/MethodLifting.scala
@@ -151,8 +151,8 @@ trait MethodLifting extends inox.ast.SymbolTransformer { self =>
         }
       }
 
-      val subCalls = (for (co <- cos.toSeq) yield {
-        firstOverrides(co).toSeq.map { case (cid, either) =>
+      val subCalls = (for (co <- cos.toSeq.sortBy(_.cid.name)) yield {
+        firstOverrides(co).toSeq.sortBy(_._1.name).map { case (cid, either) =>
           val descType = default.transform(tcd.descendants.find(_.id == cid).get.toType).asInstanceOf[t.ClassType]
           val thiss = t.AsInstanceOf(arg.toVariable, descType).copiedFrom(arg)
           (t.IsInstanceOf(arg.toVariable, descType).copiedFrom(arg), either match {
--- a/core/src/main/scala/stainless/transformers/Canonization.scala
+++ b/core/src/main/scala/stainless/transformers/Canonization.scala
@@ -24,15 +24,17 @@ trait Canonization { self =>
 
   def transform(syms: s.Symbols, vc: VC): (t.Symbols, Expr) = {
 
-    object idTransformer extends inox.ast.TreeTransformer {
+    // Stores the transformed function and ADT definitions
+    var functions = Seq[FunDef]()
+    var adts = Seq[ADTDefinition]()
+
+    object IdTransformer extends inox.ast.TreeTransformer {
       val s: self.trees.type = self.trees
       val t: self.trees.type = self.trees
 
       var localCounter = 0
       // Maps an original identifier to a normalized identifier
       val ids: mutable.Map[Identifier, Identifier] = mutable.Map.empty
-      var functions = Seq[FunDef]()
-      var adts = Seq[ADTDefinition]()
 
       def freshId(): Identifier = {
         localCounter = localCounter + 1
@@ -44,17 +46,16 @@ trait Canonization { self =>
         val nid = ids.getOrElseUpdate(id, freshId())
 
         if ((syms.functions contains id) && !visited) {
-          functions :+= transform(syms.functions(id))
+          functions = transform(syms.functions(id)) +: functions
         } else if ((syms.adts contains id) && !visited) {
-          adts :+= transform(syms.adts(id))
+          adts = transform(syms.adts(id)) +: adts
         }
 
         nid
       }
     }
 
-    val newVCBody = idTransformer.transform(vc.condition)
-
+    val newVCBody = IdTransformer.transform(vc.condition)
     val newSyms = NoSymbols.withFunctions(functions).withADTs(adts)
 
     (newSyms, newVCBody)
--- a/core/src/main/scala/stainless/transformers/Canonization.scala
+++ b/core/src/main/scala/stainless/transformers/Canonization.scala
@@ -59,7 +59,7 @@ trait Canonization { self =>
 
 
 object Canonization {
-  def canonize(p: inox.Program { val trees: stainless.ast.Trees }, vc: verification.VC[p.trees.type]): 
+  def canonize(p: inox.Program { val trees: stainless.ast.Trees })(vc: verification.VC[p.trees.type]): 
                 (p.trees.Symbols, p.trees.Expr)  = {
     object Canonizer extends Canonization {
       override val trees: p.trees.type = p.trees
--- a/core/src/main/scala/stainless/transformers/Canonization.scala
+++ b/core/src/main/scala/stainless/transformers/Canonization.scala
@@ -5,6 +5,8 @@ package transformers
 
 import stainless.extraction.inlining.Trees
 
+import scala.collection._
+
 trait Canonization { self =>
 
   val trees: stainless.ast.Trees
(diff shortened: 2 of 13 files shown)


core/src/main/scala/stainless/verification/VerificationCache.scala
```diff
@@ -0,0 +1,158 @@
...
+    p._1.adts.values.map(adt => adt.asString(uniq)).toList.sorted.mkString("\n\n") + 
+    "\n#\n" + 
+    p._2.asString(uniq)
+  }
+
+  def addVCToPersistentCache(tt: inox.ast.Trees)(p: (tt. Symbols, tt. Expr), descr: String, ctx: inox.Context): Unit = {
+
+    val task = new java.util.concurrent.Callable[String] {
+      override def call(): String = {
+        MainHelpers.synchronized {
```
samarion: Maybe synchronize on VerificationCache?

core/src/main/scala/stainless/verification/VerificationCache.scala
```diff
@@ -0,0 +1,158 @@
...
+    
+  def add(tt: inox.ast.Trees)(p: (tt.Symbols, tt.Expr)) = {
+    vccache += ((serialize(tt)(p), ()))
+  }
+
+  /**
+    * Transforms the dependencies of a VC and a VC to a String
+    * The functions and ADTs representations are sorted to avoid non-determinism
+    */
+  def serialize(tt: inox.ast.Trees)(p: (tt. Symbols, tt. Expr)): String = {
```
samarion: Please don't put spaces between between path and type.

core/src/main/scala/stainless/verification/VerificationCache.scala
```diff
@@ -0,0 +1,158 @@
...
+    val uniq = new tt.PrinterOptions(printUniqueIds = true)
+    p._1.functions.values.map(fd => fd.asString(uniq)).toList.sorted.mkString("\n\n") + 
+    "\n#\n" + 
+    p._1.adts.values.map(adt => adt.asString(uniq)).toList.sorted.mkString("\n\n") + 
+    "\n#\n" + 
+    p._2.asString(uniq)
+  }
+
+  def addVCToPersistentCache(tt: inox.ast.Trees)(p: (tt. Symbols, tt. Expr), descr: String, ctx: inox.Context): Unit = {
```
samarion: addVCToPersistentCache -> persist? (I like short names :) )

core/src/main/scala/stainless/verification/VerificationCache.scala
```diff
@@ -0,0 +1,158 @@
...
+      val result = super.checkVC(vc,sf)
+      VerificationCache.add(sp.trees)(canonic)
+      VerificationCache.addVCToPersistentCache(sp.trees)(canonic, vc.toString, ctx)
+      result
+    }
+  }
+
+}
+
+object VerificationCache {
```
samarion: It feels to me like it would make more sense to have different instances of the verification cache depending on the options with which Stainless has been called. Maybe this should be a trait and the path of the cache file should depend on the set of options in the context to make sure we don't get inconsistent results, no?

core/src/main/scala/stainless/verification/VerificationCache.scala
```diff
@@ -0,0 +1,158 @@
...
+        MainHelpers.synchronized {
+          oos.writeObject(serialize(tt)(p))
+          descr
+        }
+      }
+    }
+    MainHelpers.executor.submit(task)
+  }
+  
+  def loadPersistentCache(): Unit = {
```
samarion: It seems to me that this piece of code should only be called once on cache load, so I'd make it private to avoid any confusion. If you change this to a trait, you could have this be a static method that returns a new instance of the verification cache populated with the loaded VCs.

core/src/main/scala/stainless/verification/VerificationChecker.scala
```diff
@@ -178,29 +161,26 @@ trait VerificationChecker { self =>
...
-      val inductionTactic = InductionTactic(p)
+    val vccache = opts.findOptionOrDefault(optVCCache)
 
-      protected def getTactic(fd: p.trees.FunDef) =
-        if (fd.flags contains "induct") {
-          inductionTactic
-        } else {
-          defaultTactic
-        }
+    trait checkerInterface extends VerificationChecker {
```
samarion: Please use a capital first letter for trait names.

core/src/main/scala/stainless/transformers/Canonization.scala
```diff
@@ -0,0 +1,106 @@
...
+
+  val trees: stainless.ast.Trees
+  lazy val s: self.trees.type = self.trees
+  lazy val t: self.trees.type = self.trees
+
+  import self.trees._
+
+  type VC = verification.VC[trees.type]
+
+  def transform(syms: s.Symbols, vc: VC): (t.Symbols, Expr) = {
```
samarion: Can't you just use something like:
```scala
val transformer = new inox.ast.TreeTransformer {
  override def transform(id: Identifier): Identifier = {
    // check in renaming cache and create new if necessary
  }
}
(syms.transform(transformer), transformer.transform(vc.condition))
```

core/src/main/scala/stainless/verification/VerificationCache.scala
```diff
@@ -0,0 +1,158 @@
...
+
+    val traverser = new TreeTraverser {
+      override def traverse(id: Identifier): Unit = {
+        if (!traversed.contains(id)) {
+          traversed += id
+          if (program.symbols.functions.contains(id)) {
+            val fd = program.symbols.functions(id)
+            fundefs += fd
+            traverse(fd)
+          }
```
samarion: -- \n

core/src/main/scala/stainless/transformers/Canonization.scala
```diff
@@ -0,0 +1,106 @@
...
+          traversed += id
+          val fd = syms.functions(id)
+          val newFD = transform(fd)
+          transformedFunctions += ((id, newFD))
+        }
+      }
+
+      def transformADT(adt: ADTDefinition): ADTDefinition = {
+        val newInvariant = adt.invariant(syms) map transform
+        transform(adt).changeInvariant(newInvariant)
```
samarion: I don't think you need this. The invariant is just another function within your symbol table. The id in the flag should automatically get updated and so should the function's id, so I don't think anything needs to be done here.

core/src/main/scala/stainless/verification/VerificationCache.scala
jad-hamza: The fact that we store all dependencies with a VC should allow us to safely share the verified VCs across completely different programs right? Are you thinking about an option with the file name? (e.g. --vccache=file.bin instead of just --vccache).

core/src/main/scala/stainless/transformers/Canonization.scala
jad-hamza: I also want to transform the FunDef's (and ADTs) in the order in which they appear in the vc.condition. So I want to trigger `transform(fd)` when I encounter `fd.id` in `vc.condition`. I thought this would be 
important to avoid non-determinism in the transformation.

samarion: Yes, this is important, but the tree transformer traverses the expression in a deterministic order so you should get this for free. However, it may be useful to make the traversal order of flags within definitions deterministic!

core/src/main/scala/stainless/verification/VerificationCache.scala
samarion: More like options that potentially change certain semantics or solver behaviors. I'm not sure we actually have those yet but I'm thinking of adding an option that assumes first-class functions are total in Stainless and this could change the validity of certain VCs.

samarion: So I would make sure that either the cache name depends on the options or have some metadata within the cache itself that maintains this information.

jad-hamza: Yes, one way would be to store the options alongside the dependencies of a VC. The options can be encoded in the String representing the VC + dependencies. Or we could store a pairs of the form (options, String) instead of just a String.

core/src/main/scala/stainless/transformers/Canonization.scala
jad-hamza: But the transformation of vc.condition won't trigger the transformation of the FunDef's it refers to, right?
(It will only transform the Id's, but not the bodies of the FunDef's).

core/src/main/scala/stainless/verification/VerificationCache.scala
samarion: Sure, either of those sound good to me. You may wish to filter out some of the known options (such as optFiles) since these could lead to stupid many cache misses.

jad-hamza: Is there any option that we need to store for now?

core/src/main/scala/stainless/transformers/Canonization.scala
samarion: Ah yes, that's right. You need your transform(id) to be similar to your dependency builder (i.e. transform the functions and adts immediately when you encounter their ids). Actually, you could even merge both functionalities since you need to perform the exact same traversal both times.

jad-hamza: Yes good idea. So to make the canonical form, I would first do an "enhanced" traversal to gather the renamings, and then do a normal transformation using the computed renamings.

core/src/main/scala/stainless/verification/VerificationCache.scala
samarion: Possibly optCheckModels, optFeelingLucky and optStrictArithmetic? It seems you don't actually store the result of verification, shouldn't you report the previous output to the user?

jad-hamza: Yes correct. For now I only store the VC which were successfully verified. Ideally, I should 
also store the INVALID ones, with the corresponding model. I wanted to do this in the 
beginning, but it was a bit tedious since a Model depends on a program, so it would be 
hard to reconstruct a concrete Model from the serialized version on disk (unless we have
a way to serialize/deserialize whole programs as well).

Perhaps I should just save the String representation of the model, and not mind about 
reconstructing the concrete model from the cache?

Also, I'm not sure whether it's necessary to store these particular options? If a VC has
been checked to be VALID (or INVALID), we should be able to trust the result regardless
of these options, right?

samarion: Where do you check whether the result was VALID before persisting it? It seems to me that you store VALID for any VC that went through the checker.

About INVALID VCs, I think that we can ignore them for now. We are interested in building proofs after all. I think that as long as a VC was valid, then the options make no difference (for now). For invalid VCs, removing optCheckModel can have the VC go from INVALID to UNKNOWN, for example. That direction doesn't matter much, but if we cached UNKNOWN, then we would want to rerun the checker if the option changed.

Let's only cache valid results for now and ignore this issue. I'll try to remember to change this when I add the support for total functions.

core/src/main/scala/stainless/transformers/Canonization.scala
samarion: You can do them both at once I think:
```scala
val ids: MutableMap[Identifier, Identifier] = MutableMap.empty
val functions: MutableMap[Identifier, FunDef] = MutableMap.empty
val adts: MutableMap[Identifier, ADTDefinition] = MutableMap.empty

object transformer extends inox.ast.TeeTransformer {
  val s: self.trees.type = self.trees
  val t: self.trees.type = self.trees

  override def transform(id: Identifier): Identifier = {
    val nid = ids.getOrElseUpdate(id, /* Freshen identifier */)

    if ((syms.functions contains id) && !(functions contains id)) {
      functions(id) = transform(syms.functions(id))
    } else if ((syms.adts contains id) && !(adts contains id)) {
      adts(id) = transform(syms.adts(id))
    }

    nid
  }
}

val normalizedCond = transformer.transform(vc.condition)
val normalizedProgram = new Program {
  val trees: self.trees.type = self.trees
  val symbols = NoSymbols.withFunctions(functions.values.toSeq).withADTs(adts.values.toSeq)
}
// do something with the normalized expression and program
```

jad-hamza: Right thanks. I shouldn't even need to use buildDependencies anymore using this, right?

samarion: Yes, this should (normally) build a unique "key" for each VC (including the dependencies).

core/src/main/scala/stainless/verification/VerificationCache.scala
jad-hamza: Woops, corrrected, thanks.

core/src/main/scala/stainless/transformers/Canonization.scala
jad-hamza: I did these changes, thanks! It's working fine for now. I had to remove some more non-determinism
due to Sets. I also added the "printType" option, otherwise I don't think the cache is safe.
I had to change the executor.shutdownNow() in the MainHelpers.scala to execute.shutdown(),
otherwise the tasks writing to disk were not being executed.

core/src/main/scala/stainless/transformers/Canonization.scala
```diff
@@ -0,0 +1,78 @@
...
+  lazy val t: self.trees.type = self.trees
+
+  import self.trees._
+
+  type VC = verification.VC[trees.type]
+
+  // Sequence of transformed function definitions
+  var functions = Seq[FunDef]()
+  // Sequence of transformed ADT definitions
+  var adts = Seq[ADTDefinition]()
```
samarion: I think these are not used.

core/src/main/scala/stainless/transformers/Canonization.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  def transform(syms: s.Symbols, vc: VC): (t.Symbols, Expr) = {
```
samarion: Maybe a more precise name, like normalize?

core/src/main/scala/stainless/transformers/Canonization.scala
```diff
@@ -0,0 +1,78 @@
...
+    val newVCBody = IdTransformer.transform(vc.condition)
+    val newSyms = NoSymbols.withFunctions(functions).withADTs(adts)
+
+    (newSyms, newVCBody)
+  }
+}
+
+
+object Canonization {
+  def canonize(trs: stainless.ast.Trees)
```
samarion: Why do you take the trees as argument?

core/src/main/scala/stainless/verification/VerificationCache.scala
```diff
@@ -0,0 +1,137 @@
...
+  import program.symbols._
+  import program.trees._
+
+  val uniq = new PrinterOptions(printUniqueIds = true)
+
+  override def checkVC(vc: VC, sf: SolverFactory { val program: self.program.type }) = {
+    import VerificationCache._
+
+    // FIXME: can we simplify this?
+    val p: inox.Program { val trees: program.trees.type } = new inox.Program {
```
samarion: The type system is a bit stupid sometimes but if you remove the `trs` argument from `canonize`, you might be able to simply use `program` instead of having to build `p`.

core/src/main/scala/stainless/transformers/Canonization.scala
jad-hamza: Not sure why I added that; I tried to remove it but for some reason that doesn't typecheck (I pushed
the version that does not typecheck). If I define:
> def canonize(p: inox.Program { val trees: stainless.ast.Trees }, vc: verification.VC[p.trees.type])

Then when calling canonize (with p or with program) there is a typing error:

```scala
[error]  found   : VerificationCache.this.VC
[error]     (which expands to)  stainless.verification.VC[VerificationCache.this.program.trees.type]
[error]  required: stainless.verification.VC[p.trees.type]
[error]     val canonic = transformers.Canonization.canonize(p, vc)
[error]                                                         ^
```

samarion: I don't see the pushed version.
You probably need to have separate argument lists unless Scala made some new changes I'm unaware of:
```scala
def canonize(p: inox.Program { val trees: stainless.ast.Trees })(vc: verification.VC[p.trees.type])
```

You probably shouldn't create `p` anymore for the call to `canonize`. You probably want something more like `transformers.Canonization.canonize(program)(vc)`.

jad-hamza: Fixed, thanks!

samarion:
Could you maybe add a `reporter.info` line somewhere stating how many VCs were cache hits so that users won't be surprised by an empty verification output in VerificationCache?

jad-hamza:
I added ValidFromCache as a VCStatus, is that ok?

core/src/main/scala/stainless/extraction/oo/MethodLifting.scala
```diff
@@ -151,8 +151,8 @@ trait MethodLifting extends inox.ast.SymbolTransformer { self =>
         }
       }
 
-      val subCalls = (for (co <- cos.toSeq) yield {
-        firstOverrides(co).toSeq.map { case (cid, either) =>
+      val subCalls = (for (co <- cos.toSeq.sortBy(_.cid.name)) yield {
+        firstOverrides(co).toSeq.sortBy(_._1.name).map { case (cid, either) =>
```
samarion: I think these all have the same name (since they're overrides). The non-determinism comes from `FunOverride` storing a Set of `Override` children instead of a Seq. I think you can simply change the field.

jad-hamza: Oh I see. I thought this fixed one of the cache miss I had (when reordering class constructors in the source code). Which field should I change?

samarion: `FunOverride.children`.

jad-hamza: Lots of Sets are used in this code. Should I leave them as is, and replace Set to Seq only
in `FunOverride.children`? Then I would have to decide on an ordering for the Seq when
creating a `FunOverride`. Which order should we use?

samarion: The attached patch should work.
[MethodLifting.txt](https://github.com/epfl-lara/stainless/files/1255893/MethodLifting.txt)

jad-hamza: Seems to be working well, thanks!

samarion:
Merged, thanks!

core/src/main/scala/stainless/verification/VerificationChecker.scala
```diff
@@ -8,6 +8,8 @@ import inox.solvers._
 // TODO this should probably be removed as it is superseded by the flexible pipeline...
 object optParallelVCs extends inox.FlagOptionDef("parallelvcs", false)
 object optFailEarly extends inox.FlagOptionDef("failearly", false)
+object optFailInvalid extends inox.FlagOptionDef("failinvalid", false)
```
mantognini: This option is not exposed to the user from the command line. Is that intentional?

jad-hamza: Fixed in https://github.com/epfl-lara/stainless/pull/69, thanks

PR #67, opened 2017-08-25
Persistent cache

mantognini:
This adds the `--registry-cache=<optional director>` option. The idea is to be able to run stainless (regardless of `--watch`) based on the results of the previous run. The cache is about the extracted AST, not the VCs, as opposed to #65. This should also be a step forward toward a possible integration in Scastie.

This PR doesn't update the report system -- another PR will be dedicated to cache the components' results and restore them. It will also be useful for `--watch` to get a full report between two verification cycles.

### DIFF
--- a/core/src/main/scala/stainless/Analysis.scala
+++ b/core/src/main/scala/stainless/Analysis.scala
@@ -1,0 +1,19 @@
+/* Copyright 2009-2016 EPFL, Lausanne */
+
+package stainless
+
+/**
+ * Analyses hold all the information about a [[SimpleComponent]] results.
+ *
+ * They can be converted to basic, text-only report, which can be easily serialized or displayed to the user,
+ * or they can be used, when their concrete type is known, to explore in detail the results of components,
+ * such as the source program itself.
+ */
+trait AbstractAnalysis {
+  val name: String
+
+  type Report <: AbstractReport[Report]
+
+  def toReport: Report
+}
+
--- a/core/src/main/scala/stainless/Analysis.scala
+++ b/core/src/main/scala/stainless/Analysis.scala
@@ -3,7 +3,7 @@
 package stainless
 
 /**
- * Analyses hold all the information about a [[SimpleComponent]] results.
+ * Analyses hold all the information about a [[Component]] results.
  *
  * They can be converted to basic, text-only report, which can be easily serialized or displayed to the user,
  * or they can be used, when their concrete type is known, to explore in detail the results of components,
--- a/core/src/main/scala/stainless/MainHelpers.scala
+++ b/core/src/main/scala/stainless/MainHelpers.scala
@@ -54,7 +54,8 @@ trait MainHelpers extends inox.MainHelpers {
     extraction.inlining.optInlinePosts -> Description(General, "Inline postconditions at call-sites"),
     termination.optIgnorePosts -> Description(Termination, "Ignore existing postconditions during strengthening"),
     optJson -> Description(General, "Output verification and termination reports to a JSON file"),
-    optWatch -> Description(General, "Re-run stainless upon file changes")
+    optWatch -> Description(General, "Re-run stainless upon file changes"),
+    frontend.optPersistentRegistryCache -> Description(General, "Load/save program verification status on disk")
   ) ++ MainHelpers.components.map { component =>
     val option = inox.FlagOptionDef(component.name, default = false)
     option -> Description(Pipelines, component.description)
--- a/core/src/main/scala/stainless/MainHelpers.scala
+++ b/core/src/main/scala/stainless/MainHelpers.scala
@@ -55,7 +55,7 @@ trait MainHelpers extends inox.MainHelpers {
     termination.optIgnorePosts -> Description(Termination, "Ignore existing postconditions during strengthening"),
     optJson -> Description(General, "Output verification and termination reports to a JSON file"),
     optWatch -> Description(General, "Re-run stainless upon file changes"),
-    frontend.optPersistentRegistryCache -> Description(General, "Enable caching of program extraction"),
+    frontend.optPersistentCache -> Description(General, "Enable caching of program extraction & analysis"),
     utils.Caches.optCacheDir -> Description(General, "Specify the directory in which cache files should be stored")
   ) ++ MainHelpers.components.map { component =>
     val option = inox.FlagOptionDef(component.name, default = false)
--- a/core/src/main/scala/stainless/MainHelpers.scala
+++ b/core/src/main/scala/stainless/MainHelpers.scala
@@ -55,7 +55,8 @@ trait MainHelpers extends inox.MainHelpers {
     termination.optIgnorePosts -> Description(Termination, "Ignore existing postconditions during strengthening"),
     optJson -> Description(General, "Output verification and termination reports to a JSON file"),
     optWatch -> Description(General, "Re-run stainless upon file changes"),
-    frontend.optPersistentRegistryCache -> Description(General, "Load/save program verification status on disk")
+    frontend.optPersistentRegistryCache -> Description(General, "Enable caching of program extraction"),
+    utils.Caches.optCacheDir -> Description(General, "Specify the directory in which cache files should be stored")
   ) ++ MainHelpers.components.map { component =>
     val option = inox.FlagOptionDef(component.name, default = false)
     option -> Description(Pipelines, component.description)
--- a/core/src/main/scala/stainless/MainHelpers.scala
+++ b/core/src/main/scala/stainless/MainHelpers.scala
@@ -69,7 +69,8 @@ trait MainHelpers extends inox.MainHelpers {
     verification.DebugSectionCacheMiss,
     termination.DebugSectionTermination,
     DebugSectionExtraction,
-    frontend.DebugSectionFrontend
+    frontend.DebugSectionFrontend,
+    utils.DebugSectionRegistry
   )
 
   override protected def displayVersion(reporter: inox.Reporter): Unit = {
--- a/core/src/main/scala/stainless/MainHelpers.scala
+++ b/core/src/main/scala/stainless/MainHelpers.scala
@@ -91,33 +91,32 @@ trait MainHelpers extends inox.MainHelpers {
     val ctx = setup(args)
     val compilerArgs = args.toList filterNot { _.startsWith("--") }
 
-    val compiler = frontend.run(ctx, compilerArgs, factory)
+    val compiler = frontend.build(ctx, compilerArgs, factory)
 
-    // Passively wait until the compiler has finished
-    compiler.join()
+    // For each cylce, passively wait until the compiler has finished
+    // & print summary of reports for each component
+    def runCycle() = {
+      compiler.run()
+      compiler.join()
 
-    // When in "watch" mode, no final report is printed as there is no proper end.
-    // In fact, we might not even have a full list of VCs to be checked...
-    val watch = ctx.options.findOptionOrDefault(optWatch)
+      compiler.getReports foreach { _.emit(ctx) }
+    }
 
-    if (watch) {
+    // Run the first cycle
+    runCycle()
+
+    val watchMode = ctx.options.findOptionOrDefault(optWatch)
+    if (watchMode) {
       val files: Set[File] = compiler.sources.toSet map { file: String => new File(file).getAbsoluteFile }
-      def action() = {
-        compiler.run()
-        compiler.join()
-      }
-      val watcher = new utils.FileWatcher(ctx, files, action)
+      val watcher = new utils.FileWatcher(ctx, files, action = runCycle)
       watcher.run()
-    } else {
-      // Process reports: print summary/export to JSON
-      val reports: Seq[AbstractReport] = compiler.getReports
-      reports foreach { _.emit(ctx) }
+    }
 
-      ctx.options.findOption(optJson) foreach { file =>
-        val output = if (file.isEmpty) optJson.default else file
-        ctx.reporter.info(s"Printing JSON summary to $output")
-        exportJson(reports, output)
-      }
+    // Export final results to JSON if asked to.
+    ctx.options.findOption(optJson) foreach { file =>
+      val output = if (file.isEmpty) optJson.default else file
+      ctx.reporter.info(s"Printing JSON summary to $output")
+      exportJson(compiler.getReports, output)
     }
 
     ctx.reporter.whenDebug(inox.utils.DebugSectionTimers) { debug =>
@@ -132,7 +131,7 @@ trait MainHelpers extends inox.MainHelpers {
   }
 
   /** Exports the reports to the given file in JSON format. */
-  private def exportJson(reports: Seq[AbstractReport], file: String): Unit = {
+  private def exportJson(reports: Seq[AbstractReport[_]], file: String): Unit = {
     val subs = reports map { r => JObject(r.name -> r.emitJson) }
     val json = if (subs.isEmpty) JObject() else subs reduce { _ ~ _ }
     val string = pretty(render(json))
--- a/core/src/main/scala/stainless/MainHelpers.scala
+++ b/core/src/main/scala/stainless/MainHelpers.scala
@@ -7,8 +7,7 @@ import utils.JsonUtils
 import java.io.File
 import java.util.concurrent.{ ExecutorService, Executors }
 
-import org.json4s.JsonAST.JObject
-import org.json4s.JsonDSL._
+import io.circe.Json
 
 object MainHelpers {
 
@@ -133,8 +132,7 @@ trait MainHelpers extends inox.MainHelpers {
 
   /** Exports the reports to the given file in JSON format. */
   private def exportJson(reports: Seq[AbstractReport[_]], file: String): Unit = {
-    val subs = reports map { r => JObject(r.name -> r.emitJson) }
-    val json = if (subs.isEmpty) JObject() else subs reduce { _ ~ _ }
+    val json = Json.fromFields(reports map { r => (r.name -> r.emitJson) })
     JsonUtils.writeFile(new File(file), json)
   }
 
--- a/core/src/main/scala/stainless/MainHelpers.scala
+++ b/core/src/main/scala/stainless/MainHelpers.scala
@@ -2,12 +2,13 @@
 
 package stainless
 
-import java.io.{ File, PrintWriter }
+import utils.JsonUtils
+
+import java.io.File
 import java.util.concurrent.{ ExecutorService, Executors }
 
 import org.json4s.JsonAST.JObject
 import org.json4s.JsonDSL._
-import org.json4s.native.JsonMethods._
 
 object MainHelpers {
 
@@ -134,9 +135,7 @@ trait MainHelpers extends inox.MainHelpers {
   private def exportJson(reports: Seq[AbstractReport[_]], file: String): Unit = {
     val subs = reports map { r => JObject(r.name -> r.emitJson) }
     val json = if (subs.isEmpty) JObject() else subs reduce { _ ~ _ }
-    val string = pretty(render(json))
-    val pw = new PrintWriter(new File(file))
-    try pw.write(string) finally pw.close()
+    JsonUtils.writeFile(new File(file), json)
   }
 
 }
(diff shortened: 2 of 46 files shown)


mantognini:
I'll modify this PR a tiny bit: `--registry-cache` will be a simple on/off switch option like `--vccache`, and I'll add a `--cache-dir=<default value>` option, which defines in which directory the cache files (including `vccache.bin`) should be stored on disk, to let the user have some "session" directories (useful when using stainless in Scasite, for example).

mantognini:
Here, the option was added and `VerificationCache` updated to reflect the new setting.

mantognini:
Just posting this here because we all love when things work fast and nicely:

Running stainless on `verification/valid` benchmarks takes roughly 240s of CPU on my machine. When using @jad-hamza's `--vccache` option we get down to ~70s! (BTW, using one the latest inox version we have 100% cache hit.) And using `--registry-cache` we go down to ~30s of CPU time (< 10 wall-clock seconds on my machine).

mantognini:
(I've changed the base of this PR to make the review process easier. When #73 is merged, it should be changed again to `master`.)

core/src/main/scala/stainless/utils/Registry.scala
```diff
@@ -83,17 +89,137 @@ trait Registry {
...
+      case Left(cd) => cd.id
+      case Right(fd) => fd.id
+    }
+
+    def isChecked(node: NodeValue) = node match {
+      case Left(cd) => shouldBeChecked(cd)
+      case Right(fd) => shouldBeChecked(fd)
+    }
+
+    val localCFCache = MutableMap[Identifier, CanonicalForm]() // cache CanonicalForm and their hash.
```
samarion: Private?

core/src/main/scala/stainless/utils/Registry.scala
```diff
@@ -24,9 +28,25 @@ import scala.collection.mutable.{ Map => MutableMap, Set => MutableSet }
  * Specific implementations of this trait have to provide a context and facilities to compute
  * direct dependencies for functions and classes, as well as filters to identify data that
  * is of interest.
+ *
+ * Regarding the persistent cache:
```
samarion: Please add 1-2 lines saying what the persistent cache is actually doing.

core/src/main/scala/stainless/utils/Registry.scala
```diff
@@ -24,9 +28,25 @@ import scala.collection.mutable.{ Map => MutableMap, Set => MutableSet }
...
+ *  - It is expected that the cache is loaded before the first cycle starts. The behaviour is undefined
+ *    if it is loaded during any cycle or between two cycles.
+ *  - In order to re-compute nodes whose dependencies have changed, we need the knowledge of the
+ *    full universe, i.e. all function and class definitions, in order to known which nodes should
+ *    be (re)computed by the underlying ICG. To do that, when the cache is loaded, the [[update]] and
+ *    [[checkpoint]] method will switch to a specific mode which defer all updates to [[checkpoint]].
+ *  - The [[isOfInterest]] method is also switching mode whenever the persistent cache is loaded.
+ *
+ * NOTE If stainless is interrupted, the cache will most probably be invalid, or at least not represent
+ *      the state of processed/unprocessed nodes.
```
samarion: Maybe make the cache interruptible and register it to the `InterruptManager` if this is a possibility? You can clean the invalid parts of the cache in that case.

core/src/main/scala/stainless/utils/Registry.scala
mantognini: It's local to the `saveCache` method. ;-)

core/src/main/scala/stainless/utils/Registry.scala
mantognini: will do

core/src/main/scala/stainless/utils/Registry.scala
mantognini: The only sane thing I could do is delete the cache file -- to delete the invalid bits, it would mean adding a progression state to the registry and complexify things for such a rare incident.

mantognini:
> Could you coordinate with Jad about using a single normalization function between the VC cache and this one?

We can investigate that, yes. I'll do some experiments to see which canonical form is more compact than the other.

core/src/main/scala/stainless/utils/Registry.scala
samarion: I'm not sure that interrupting Stainless is a rare occurrence (like cancelling a VC). Can't you defer the cache writing decision to the components and let them make sure to only write to the cache when the component has ended in a valid state (e.g. not cancelled)?

I guess it'll be pretty much impossible to keep a valid cache when receiving a SIGTERM... I believe that an invalid cache state shouldn't create unsoundness during verification, if I understand correctly?

samarion:
It'm pretty sure yours will be more compact (and possibly more sound as well), since Jad's relies on `asString` to provide an unambiguous canonical form. However, Jad normalizes global (function and class) identifiers, which should give better caching results. Also, now that extractors also return identifiers, you may be able to simplify your canonical form builder a bit.

core/src/main/scala/stainless/utils/Registry.scala
mantognini: Taking into account the program can be ungracefully interrupted/killed, I think it's simpler to add a check when loading the cache to make sure it's complete -- this check is already there in `loadCache` in form of assertions: every time we read something we know its size and therefore can detect something is missing. If the cache is corrupted any other way, well..., we can't do something about that but it would be a user mistake, not one from stainless. I'll update the assertions to print a decent message to let the user know their cache is invalid.

mantognini:
> However, Jad normalizes global (function and class) identifiers, which should give better caching results.

Yes, I'm definitely going to take this into account when doing my experiments. ;-)

> Also, now that extractors also return identifiers, you may be able to simplify your canonical form builder a bit.

Only one or two lines of code would change I think for the current usage.

core/src/main/scala/stainless/utils/Registry.scala
samarion: This sounds good. But this won't cover the case of VC cancellation, will it? In that case, the cache would be complete since the cancellation is supported by the verification component, but the cache will assume all subsequent runs will have a cancelled result, no?

mantognini: No, it won't. The basic issue is that the Registry doesn't get feedback from component to know what is valid, unknown, invalid, ... It's simply care whether it was processed or not.

Currently this information (the reports) is not available at all to the Registry. Independently of this PR, I want to be able to cache the components' reports as well. Maybe we can use that to remove things from the registry's cache when saving it to the disk.

samarion: Okay, we can leave this for a future PR. But this is another reason for https://github.com/epfl-lara/stainless/pull/72 to have a more principled relation between input functions and reports to make sure the symbols uniquely determine the functions that will exist within the report.

samarion:
Ok, lets go ahead and merge these (this PR and https://github.com/epfl-lara/stainless/pull/72). We'll see about programs implying VCs and caching transformers at some later point.

Please rebase and merge. Note that I've changed the program API to not contain the context anymore (shouldn't affect your rebases much I think).

mantognini:
I've been working on the persistent report cache. It doesn't change much to this PR but I think it makes sense to wait until I've put all the piece together and merge everything at once. I'll let you know when I've something stable.

mantognini:
This is still somewhat work in progress, but the main changes are:

 - `Report`s are now `Analysis` and contain all the data relevant to the component process (i.e. the program, the checker, ...)
 - From an `Analysis` we can derive a simple text `Report` that is easily serialisable and provides an API to combine new `Report` together or remove old data (basically).
 - In `--watch` mode, a full report is printed after each compilation cycle.
 - The persistent cache now includes the `Report`s too and the `--watch` mode is (more) compatible with the persistent cache.
 - Some refactoring, mostly moving classes to their own files -- in some case it should improve recompilation time (e.g. nested trait usually trigger a lot of recompilation, now some are extracted to top level).
 - I also switch the JSON library; instead of `json4s` it now uses `circe` mainly because it should be easier to maintain code with it's (semi-auto) generic code interface.

mantognini:
Latest addition, specifically designed for `--watch` mode: new VC/TC are clearly marked as "NEW" in the user reports. The registry is not yet perfect and will get confused when the frontend compiler generates different `Identifier`s, but overall it should be good 

Feedback welcome if you guys end up checking this out.

BTW, regarding the command line options, there is currently some inconsistency: some options use `-` to separate words (e.g. `--registry-cache`) while others simply consist of words being concatenated (e.g. `--inlineposts`). I think the former is *much* more readable but I'm happy to modify those options to the other format for consistency reasons. What's the general feeling about this?

core/src/main/scala/stainless/Analysis.scala
```diff
@@ -0,0 +1,19 @@
+/* Copyright 2009-2016 EPFL, Lausanne */
+
+package stainless
+
+/**
+ * Analyses hold all the information about a [[SimpleComponent]] results.
```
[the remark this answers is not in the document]
mantognini: Yes indeed!

core/src/main/scala/stainless/MainHelpers.scala
```diff
@@ -54,7 +54,9 @@ trait MainHelpers extends inox.MainHelpers {
     extraction.inlining.optInlinePosts -> Description(General, "Inline postconditions at call-sites"),
     termination.optIgnorePosts -> Description(Termination, "Ignore existing postconditions during strengthening"),
     optJson -> Description(General, "Output verification and termination reports to a JSON file"),
-    optWatch -> Description(General, "Re-run stainless upon file changes")
+    optWatch -> Description(General, "Re-run stainless upon file changes"),
+    frontend.optPersistentRegistryCache -> Description(General, "Enable caching of program extraction"),
```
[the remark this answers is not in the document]
mantognini: Analyses' report are indeed cached. Will change the associated text.

core/src/main/scala/stainless/Report.scala
```diff
@@ -16,9 +16,26 @@ case class ReportStats(total: Int, time: Long, valid: Int, validFromCache: Int,
...
-  val name: String
-  def emitJson: JArray
+/**
+ * Text version of [[AbstractAnalysis]] that holds the basic information a user might be interested in.
+ *
+ * Provide facilities for I/O serialisation (through a JSON interface), generating a human-readable view
+ * of the results (through an ASCCI table) and the ability to maintain an up-to-date view of the results
+ * by means of concatenation of reports and invalidation of some part of the report itself.
+ *
+ * [[SelfType]] is used for typechecking purposed: it should denote the subclass itself. E.g.:
```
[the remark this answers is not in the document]
mantognini: Fixed

core/src/main/scala/stainless/Report.scala
```diff
@@ -16,9 +16,26 @@ case class ReportStats(total: Int, time: Long, valid: Int, validFromCache: Int,
...
+ */
+trait AbstractReport[SelfType <: AbstractReport[SelfType]] { self: SelfType =>
+  val name: String // The same name as the [[AbstractAnalysis]] this report was derived from.
+
+  def emitJson: Json
+
+  /** Create a new report with *only* the information about the given functions/classes/... */
+  def filter(ids: Set[Identifier]): SelfType
+
+  /** Merge two reports, considering [[other]] to contain the last information in case of update. */
```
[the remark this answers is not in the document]
mantognini: Yes, thanks.

core/src/main/scala/stainless/verification/VerificationReport.scala
```diff
@@ -0,0 +1,116 @@
...
+        Cell(pos.fullString),
+        Cell(status.name),
+        Cell(solverName getOrElse ""),
+        Cell(f"${time / 1000d}%3.3f")
+      ))
+    },
+    ReportStats(totalConditions, totalTime, totalValid, totalValidFromCache, totalInvalid, totalUnknown)
+  ))
+
+  override def ~(other: VerificationReport): VerificationReport = {
```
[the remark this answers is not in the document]
mantognini: I'll try to refactor things to DRY code further. :)

mantognini:
> IIUC, you're not yet performing program normalizations to improve cache hit rates, right? Does this mean we have a huge amount of cache misses because of non-determinism in the frontend compilers? Once we introduce these normalizations, will we have an issue with the serialized identifiers in the reports not corresponding to the identifiers in the actual programs?

I'm indeed not doing such normalisations. Cache misses are expected to be zero if nothing changes at all, but can indeed be significant, yet probably not "huge", when recompiling an edited program (although modifying the body of functions shouldn't have an impact).

Introducing such normalisations, or even unification of top level functions (e.g. two identical functions could be "merged" into one), will most certainly require changing the Report API a bit, but I think this doesn't impact the serialised identifiers in the reports -- the serialised data might be different though.

> Is the cache still at the program level (and not the function level)? 

The report cache is at the function level, kind of -- the report is fragmented into record which can be grouped by functions. The `Registry` cache (the canonical forms) is also at the function/class level.

> About the option names, I agree the names should be consistent. I don't really care either way about having dashes or not, please do as you prefer (but please make sure the doc is also up to date with whichever option you choose).

Cool, I'll do that in a separate PR. I've just modified the `--registry-cache` into `--cache` as it's more general that extraction, as you correctly noted.

mantognini:
I DRY'ed the code a bit, but not using an inherited common interface for the Report/Record as it was too heavily imposed on other parts (e.g. tests).

core/src/main/scala/stainless/Report.scala
```diff
@@ -62,6 +62,40 @@ trait AbstractReport[SelfType <: AbstractReport[SelfType]] { self: SelfType =>
   }
 }
 
+object AbstractReportHelper {
+  type Record = { val fid: Identifier }
```
samarion: Argh! I think I prefer code duplication over invoking functions through reflection...
What is wrong with introducing an `AbstractRecord` class with this signature?

mantognini: I guess there was one alternative I've overlooked. Initially I tried to put the generic code inside `AbstractRecord` itself, which ended up making things really ugly -- it was no longer abstract reports, but specialised report for records such as `VerificationReport` and `TerminationReport` and therefore would prohibits future extensions.

The last commit should be decent enough, although solutions that aren't based on inheritance should be possible I guess.

core/src/main/scala/stainless/Report.scala
```diff
@@ -63,23 +63,22 @@ trait AbstractReport[SelfType <: AbstractReport[SelfType]] { self: SelfType =>
 }
 
 object AbstractReportHelper {
-  type Record = { val fid: Identifier }
-  import scala.language.reflectiveCalls
+  trait Record { val id: Identifier }
```
mantognini: Let me know if this alternative works better for you. ;-)

mantognini: I thought it was more generic, because those ids could also be for classes from the `AbstractReport` point of view.

samarion: You could rename the `fid` fields to `id`, it should be clear from the report class whether it's targeting a function or a class. But this version is fine as well, I just tend to prefer shorter code.

mantognini: I don't mind changing it. As you prefer. ;-)

samarion: Well I (slightly) prefer the shorter version with `id` :)

mantognini: Done. ;-)

mantognini:
I've added a few fixes. Also, now stainless will stop if something wasn't properly extracted (i.e. a dependency is missing at the end of an extraction cycle). Some additional tests are also here to ensure it is more robust against such bugs.

frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
```diff
@@ -1320,6 +1320,7 @@ class CodeExtraction(inoxCtx: inox.Context, cache: SymbolsContext)(implicit val
     case tpe if tpe.typeSymbol == defn.LongClass    => xt.Int64Type()
     case tpe if tpe.typeSymbol == defn.BooleanClass => xt.BooleanType()
     case tpe if tpe.typeSymbol == defn.UnitClass    => xt.UnitType()
+    case tpe if tpe.typeSymbol == defn.NothingClass => xt.NothingType()
```
mantognini: See comment here: https://github.com/epfl-lara/stainless/commit/929c301b8a53993477f82123478891ed6eec7e1e#commitcomment-24778023

frontends/scalac/src/main/scala/stainless/frontends/scalac/ASTExtractors.scala
```diff
@@ -748,15 +748,101 @@ trait ASTExtractors {
...
+             => Some((array, index, value))
+
+        // There's no `updated` method in the Array class itself, only though implicit conversion.
         case _ => None
       }
     }
 
+    object ExArrayUpdate {
+      def unapply(tree: Apply): Option[(Tree, Tree, Tree)] = tree match {
+        // This implicit conversion to ArrayOps is shadowed. We therefore skip it here.
```
samarion: Why is this comment here?

frontends/scalac/src/main/scala/stainless/frontends/scalac/ASTExtractors.scala
```diff
@@ -748,15 +748,101 @@ trait ASTExtractors {
...
+        case Apply(ExArraySelect(array, "update"), Seq(index, newValue)) => Some((array, index, newValue))
+        case _ => None
+      }
+    }
+
+    object ExArrayApply {
+      def unapply(tree: Apply): Option[(Tree, Tree)] = tree match {
+        // This implicit conversion to ArrayOps is shadowed. We therefore skip it here.
```
samarion: Same as above

core/src/main/scala/stainless/utils/Registry.scala
```diff
@@ -365,7 +365,7 @@ trait Registry {
       case Right(fd) if !(recentFunctions contains fd) => fd.id
     }
 
-    reporter.debug(s"Removing <${toRemove mkString ", "}> from graph")
+    reporter.debug(s"Removing <${toRemove map { _.uniqueName } mkString ", "}> from graph")
```
samarion: Ideally, use `asString` for identifiers (if you have a context available), that way the command-line options determine what is shown. Same comment for other calls to `uniqueName` in reporting.

samarion:
One final comment I have (it's in the review but on an outdated diff it seems, so not very visible) relates to using `id.uniqueName` inside debug/output strings. In cases where we have an available `inox.Context`, it would be better to use `id.asString` and let the user determine whether he wants a unique name or not through the `--print-ids` command-line option.

frontends/scalac/src/main/scala/stainless/frontends/scalac/ASTExtractors.scala
mantognini: It goes in the same direction as [the note on line 761](https://github.com/epfl-lara/stainless/pull/67/files/c072af1f25467611795a9cb6fb90d046d5948904..33c2b6d98e76baf61d97d7181b833647db463ece#diff-fdd435b04b358dd3755d6e03a7bfb113R761) and explain why it wasn't implemented.

mantognini:
You're right about the `asString`, but it doesn't apply only to `Identifier`s. I'll work later on a general PR to update the codebase to use as much as possible `asString` instead of (the often implicit) `toString`.

frontends/scalac/src/main/scala/stainless/frontends/scalac/ASTExtractors.scala
samarion: You mean that `ArrayOps.update()` can never happen because the method already exists in `Array` itself?

samarion:
Yes, you're right. It's indeed not a blocker for this PR.

frontends/scalac/src/main/scala/stainless/frontends/scalac/ASTExtractors.scala
mantognini: You can use it, but you need to use something like that: `(array: 
scala.collection.mutable.ArrayOps[T]).update(i, v)`, i.e. the conversion needs to be explicitly written down and there's probably zero use case for that so I felt it wasn't useful to spend time and implement support for that.

samarion: Ok, that's fine.

mantognini:
I think, besides the newly opened issues, this is stable and ready to be merged. Further (implementation) improvement can be done later depending on our priorities. Do you agree?

samarion:
Yep

PR #70, opened 2017-08-31
missing positions for lambda precondition

jad-hamza:
Some positions were missing in the verification reports ("lambda precondition" checks).

### DIFF
--- a/core/src/main/scala/stainless/extraction/PreconditionInference.scala
+++ b/core/src/main/scala/stainless/extraction/PreconditionInference.scala
@@ -33,7 +33,7 @@ class PreconditionInference(override val sourceProgram: Program { val trees: ext
       def apply(caller: Expr, args: Seq[Expr]): Expr = caller.getType match {
         case FunctionType(from, to) if args.size >= from.size =>
           val (currArgs, restArgs) = args.splitAt(from.size)
-          apply(Application(caller, currArgs), restArgs)
+          apply(Application(caller, currArgs).copiedFrom(caller), restArgs)
         case _ => caller
       }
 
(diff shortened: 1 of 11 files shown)


core/src/main/scala/stainless/extraction/PreconditionInference.scala
```diff
@@ -40,7 +40,10 @@ class PreconditionInference(override val sourceProgram: Program { val trees: ext
       def pre(caller: Expr, args: Seq[Expr]): Expr = caller.getType match {
         case FunctionType(from, to) =>
           val (currArgs, restArgs) = args.splitAt(from.size)
-          and(Application(Pre(caller), currArgs), pre(Application(caller, currArgs), restArgs))
+          and(
+            Application(Pre(caller), currArgs).copiedFrom(caller),
```
samarion: Please add the position to the Pre() tree as well.

core/src/main/scala/stainless/extraction/PreconditionInference.scala
```diff
@@ -40,7 +40,10 @@ class PreconditionInference(override val sourceProgram: Program { val trees: ext
       def pre(caller: Expr, args: Seq[Expr]): Expr = caller.getType match {
         case FunctionType(from, to) =>
           val (currArgs, restArgs) = args.splitAt(from.size)
-          and(Application(Pre(caller), currArgs), pre(Application(caller, currArgs), restArgs))
+          and(
+            Application(Pre(caller), currArgs).copiedFrom(caller), 
+            pre(Application(caller, currArgs).copiedFrom(caller), restArgs)
+          )
         case _ => BooleanLiteral(true)
```
samarion: Set the position as well.

jad-hamza:
I added some more missing positions, but I just noticed that there are still VCs with no positions (e.g. in verification/valid/FlatMap.scala). Investigating!

jad-hamza:
I added some more. And also some in Inox (https://github.com/epfl-lara/inox/pull/38)

I think many are still missing (dozens), but at least now all positions are defined in the verification report for verification/valid/*.scala

samarion:
Merged

PR #72, opened 2017-09-01
VC filter

mantognini:
This filter avoids computing the same VCs twice during the same extraction cycle, which can yield several overlapping sub-programs.

### DIFF
--- a/core/src/main/scala/stainless/Component.scala
+++ b/core/src/main/scala/stainless/Component.scala
@@ -41,13 +41,16 @@ trait SimpleComponent extends Component { self =>
     extraction.extract(program, ctx).transform(lowering)
   }
 
+  private val marks = new utils.AtomicMarks[Identifier]
+  def onCycleBegin(): Unit = marks.clear()
+
   def apply(program: Program { val trees: xt.type }, ctx: inox.Context): Report = {
     val extracted = extract(program, ctx)
     import extracted._
 
     val filter = new utils.CheckFilter { override val context = ctx }
     val relevant = symbols.functions.values.toSeq filter { fd =>
-      filter.shouldBeChecked(fd.id, fd.flags)
+      filter.shouldBeChecked(fd.id, fd.flags) && marks.compareAndSet(fd.id)
     } map { _.id }
 
     apply(relevant, extracted, ctx)
--- a/core/src/main/scala/stainless/frontend/CallBackWithRegistry.scala
+++ b/core/src/main/scala/stainless/frontend/CallBackWithRegistry.scala
@@ -15,7 +15,7 @@ trait CallBackWithRegistry extends CallBack { self =>
 
   /******************* Public Interface: Override CallBack ***************************************/
 
-  final override def beginExtractions(): Unit = { /* nothing */ }
+  override def beginExtractions(): Unit = { /* nothing */ }
 
   final override def apply(file: String, unit: xt.UnitDef,
                            classes: Seq[xt.ClassDef], functions: Seq[xt.FunDef]): Unit = {
@@ -63,8 +63,6 @@ trait CallBackWithRegistry extends CallBack { self =>
   protected def shouldBeChecked(fd: xt.FunDef): Boolean
   protected def shouldBeChecked(cd: xt.ClassDef): Boolean
 
-  /** Mark the end of a callback cycle. */
-  protected def onCycleEnd(): Unit = () // Nothing by default.
 
   /******************* Internal State *************************************************************/
 
--- a/core/src/main/scala/stainless/frontend/CallBackWithRegistry.scala
+++ b/core/src/main/scala/stainless/frontend/CallBackWithRegistry.scala
@@ -15,8 +15,6 @@ trait CallBackWithRegistry extends CallBack { self =>
 
   /******************* Public Interface: Override CallBack ***************************************/
 
-  override def beginExtractions(): Unit = { /* nothing */ }
-
   final override def apply(file: String, unit: xt.UnitDef,
                            classes: Seq[xt.ClassDef], functions: Seq[xt.FunDef]): Unit = {
     context.reporter.debug(s"Got a unit for $file: ${unit.id} with:")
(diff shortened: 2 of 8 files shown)


mantognini:
(I've changed the base of this PR to make the review process easier. When #73 is merged, it should be changed again to `master`.)

mantognini:
Re-working on this one.. not as good as I thought.

mantognini:
Compared to the previous version, now this PR filter functions (before VCs are generated) and not the generated VCs themselves. It should therefore be more efficient in addition to be a simpler PR.

You might want to look at `AtomicMarks` as it's quite general and could be useful in other part of stainless.

samarion:
This looks good to me but I'm not sure I understand the use case. Shouldn't the extraction cycle only yield non-overlapping verification tasks (even when the programs are overlapping)?

mantognini:
> Shouldn't the extraction cycle only yield non-overlapping verification tasks (even when the programs are overlapping)?

Not currently: say you verify `A.scala`  and `B.scala` and that those files contain functions in `B` depending on functions defined in `A`. The ICG/Registry will generate `Symbols` for self-contained programs as soon as one is extracted, meaning that if the compiler extract `A` first, a sub-program with functions of `A` is sent to component pipeline. Then, when `B` is extracted a second sub-program is sent to the pipeline but it contains (part of) `A` as well to satisfy dependencies. This is were this new filter comes into play and filters out functions at the solver level, yet keeping all definitions available.

samarion:
Hmm, couldn't you keep track of which core functions (excluding the transitive dependencies) were used to generate the symbols and then only use these functions to generate VCs? I guess it's fairly similar in functionality, it just seems easier to support more components this way.

mantognini:
I'm not sure.. Wouldn't that require changing the `Symbols` (or `Program`) so that `VerificationGenerator` and other can extract the proper information from the transformed program? Seems more complex than the current solution, no?

samarion:
You can simply carry around a set of identifiers alongside the program to determine which functions need to be verified. We already have the `Derived` annotation that can be used to find which functions were generated from others. I'm not sure how consistently it's handled though.

It may be a little more complex, I'm not sure. But the current implementation requires shared state whereas this would work even across different verification backends.

samarion:
Actually, we already take the `funs` as argument in `check`. Wouldn't it be easy to make sure these are disjoint?

mantognini:
I've moved things to `SimpleComponent`. Now the termination component should also benefit from this filter. I've decided not to refactor further the Verification and Termination component into `CallBackWithRegistry` because it would put a lot of things in this class. Instead, subclasses are forced to `override def beginExtractions() `. 

Does this change suit you? :) I think it's more lightweight in terms of implementation than modifying other classes or extracting information from trees.

samarion:
Well, it'll do for now :) If we use programs as keys for reports, I would still like for the set of functions that will belong to the report to be uniquely determined by the program, but this can be a feature for the future.

mantognini:
> If we use programs as keys for reports

I'm not sure it'll work, especially under `--watch`. Say a program p1 is generated for foo and bar and the report is collected. Then the user modifies foo only, a program p2 gets generated and a new report is created. The old report is partially valid because bar hasn't changed and the information relative to bar should be presented to the user (again -- this is not the case ATM). So I think it would be better to associate (sub-)reports to functions instead.

samarion:
So are we assuming bar doesn't call foo here? In that case, there should exist a program with only bar for which verification will be run, and its report should not change when foo is changed. So the info relative to bar will be loaded from cache, but from the report associated to the program containing only bar.

If bar does call foo, then we need to rerun verification of bar since verification depends on the implementation of foo.

With the current implementation, if you re-run Stainless from scratch and the report for bar was presented based on its verification result in the program containing both bar and foo, the verifier should discard that report since the program's normal form has changed and re-verify both foo and bar, which constitutes a cache miss of validly cached data.

In essence, the solution I'm proposing associates sub-reports to strongly connected components in the call-graph, which is the minimal unit that needs to be re-verified when there is a function change. Maybe this will require some changes in the incremental symbol builder to make sure it returns one program per component.

mantognini:
> So are we assuming bar doesn't call foo here? In that case, there should exist a program with only bar for which verification will be run, and its report should not change when foo is changed.

In practice this makes things slow (+30s CPU / =~ +40% on `verification/valid` *when* using a cache for VCs, so just doing some tree transformations). That's why I've reverted the *Yield several smaller programs from Registry* commit, which should just do what you want.

For the end user, both solutions yield the same results. So I'm wondering if we want to modify the internal structure significantly (to take care of the performance penalty and so on) or rather use function as the base "unit" for sub-reports and keep most things as is.

samarion:
I'm not surprised the current tree transformations are slow, but these should be cachable at the SCC level (in the transformers). Transformers with caches should be even more efficient than the current version since there will be no transformation overlap at all!

It's probably my love for clean design that's speaking though...

Also, won't you be seeing many cache misses with the current version of the registry cache? If the non-minimal program that is output by extraction is slightly different from the cached one (e.g. one unrelated function more or less), you'll see a cache miss for the full report, no?

mantognini:
> It's probably my love for clean design that's speaking though...

Which is perfectly understandable. ;-)

> Also, won't you be seeing many cache misses with the current version of the registry cache? If the non-minimal program that is output by extraction is slightly different from the cached one (e.g. one unrelated function more or less), you'll see a cache miss for the full report, no?

The current registry cache is highly sensible to changes in identifiers produced by the underlying compiler -- inserting or removing a definition will result in many definitions to be re-verified. Thankfully, the VC cache solve this particular issue but after all transformations are applied.

I keep in a corner of my head the idea of performing a more agressive unification of definitions (*) but since this is not a trivial matter I'm focusing on other things: Scastie integration and (hopefully) GenC.

(*) Say we have two identical functions in the user program. We should properly verify them (well, only one of them) and display report for both of them. That's one of the challenges that goes with this unification. It shouldn't be hard to find a proper solution but time flies so.. ;-)

mantognini:
patch included in #67.

PR #74, opened 2017-09-06
Increase memory limits when launching stainless

mantognini:
This is a modest PR to fix issues I've been having when running stainless on Mac. Thanks @gsps for the tip! :-)

This fixes (or at least reduces) errors such as

    java(9672,0x70010ae84000) malloc: *** error for object 0x7f8e72294c00: double free

The main limitation was the stack size. Other parameters were also increased however.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -179,7 +179,7 @@ val scriptSettings: Seq[Setting[_]] = Seq(
                                |
                                |SCALACLASSPATH="$paths"
                                |
-                               |java -Xmx2G -Xms512M -Xss64M -classpath "$${SCALACLASSPATH}" -Dscala.usejavacp=true stainless.Main $$@ 2>&1 | tee -i last.log
+                               |java -Xmx8G -Xms2G -Xss1G -classpath "$${SCALACLASSPATH}" -Dscala.usejavacp=true stainless.Main $$@ 2>&1 | tee -i last.log
                                |""".stripMargin)
       scriptFile.setExecutable(true)
     } catch {
--- a/core/src/main/scala/stainless/verification/VerificationChecker.scala
+++ b/core/src/main/scala/stainless/verification/VerificationChecker.scala
@@ -92,6 +92,7 @@ trait VerificationChecker { self =>
   protected def checkVC(vc: VC, sf: SolverFactory { val program: self.program.type }): VCResult = {
     import SolverResponses._
     val s = sf.getNewSolver
+    val solverName = Some(s.name)
 
     try {
       val cond = simplifyLets(vc.condition)
@@ -112,7 +113,7 @@ trait VerificationChecker { self =>
 
         res match {
           case _ if interruptManager.isInterrupted =>
-            VCResult(VCStatus.Cancelled, Some(s), Some(time))
+            VCResult(VCStatus.Cancelled, solverName, Some(time))
 
           case Unknown =>
             VCResult(s match {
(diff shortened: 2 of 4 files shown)


samarion:
LGTM

mantognini:
Note: this helps when working on "small" programs, but it still crashes (usually with a different error) when verifying `verification/valid`... Apparently too many tasks are spawned, so I'll try to reduce the number of parallel verification tasks as well.

mantognini:
So.. even with only one solver running at a time (tried: two threads with binary semaphore for solver), it still crashes in native code (e.g. z3). Problem seems related to memory leaks. No clue *yet* where they come from.

mantognini:
It's not strictly memory leaks, but the current design keeps references all over the place so we end up never releasing most things (e.g. the report have VC results which have a reference to the solver which has one to the programs and therefore the trees & symbols). I guess without changing significantly the design we cannot truly solve this issue.

samarion:
I don't think programs, trees and symbols should use so much memory... they're relatively small objects and there aren't so many of them around. However, the memory footprint of a solver can grow quite large due to unrolling. Can't we get rid of the reference to the solver within the VCResult? Is it used for anything else than getting the solver's name?

mantognini:
It was indeed its only usage. It seems to reduce memory usage a bit, but since JVM monitoring tools are quite unstable (and crash) I cannot give an exact number; it reduces memory usage by around 50-100 MB according to `time -v` "Maximum resident set size". I'll test this again on my Mac when I've access to it but I don't expect the changes to make much improvements.

mantognini:
Sadly this doesn't solve a thing. I retried the patch on my laptop and even verifying a small program results in a crash.

When verifying the `verification/valid` benchmark suite I get something along those lines:

```
java(54599,0x7002085a1000) malloc: *** error for object 0x7f817de68400: double free
*** set a breakpoint in malloc_error_break to debug
/somepath/stainless/bin/stainless: line 9: 54599 Abort trap: 6           java -Xmx8G -Xms2G -Xss1G -classpath "${SCALACLASSPATH}" -Dscala.usejavacp=true stainless.Main $@ 2>&1
     54600 Done                    | tee -i last.log
```
or something more complete such as:

```
$ stainless frontends/benchmarks/verification/valid/*.scala --timeout=1
[ stainless starts verifying some VC, some finish, and then this: ]
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00000001318d77c4, pid=54723, tid=0x0000000000005903
#
# JRE version: Java(TM) SE Runtime Environment (8.0_144-b01) (build 1.8.0_144-b01)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.144-b01 mixed mode bsd-amd64 compressed oops)
# Problematic frame:
# C  [libz3.dylib+0x8217c4]  void mpz_manager<true>::set<0>(mpz&, int, unsigned int)+0x20
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /somepath/stainless/hs_err_pid54723.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
/somepath/stainless/bin/stainless: line 9: 54723 Abort trap: 6           java -Xmx8G -Xms2G -Xss1G -classpath "${SCALACLASSPATH}" -Dscala.usejavacp=true stainless.Main $@ 2>&1
     54724 Done                    | tee -i last.log
```

Looking at the report file generated by java, it seems two versions of z3 are loaded:

```
0x000000012fc68000 	/private/var/folders/2w/r3qm7hw96z1807mg5hc95t8c0000gn/T/SCALAZ3_2753beb29c1a569725a4ada678cafa/lib-bin/libz3.dylib
0x00000001310b6000 	/usr/local/lib/libz3.dylib
```

So maybe all that is related to a mismatch in z3 binaries... I need to investigate more but in the meantime this PR should be ignored.

Unrelated to that, changes to `VCResult` might not be needed because of serialisation design: I need to change how we interact with reports anyway. I'll explain the details in a follow up of #67.

mantognini:
Exploring this further, I modified locally ScalaZ3 to have only one compiled version of z3 on my system, therefore excluding binary mismatch. It didn't help. I also toyed with the interpreter settings to limit the memory available from some ridiculously small 4MB to much larger heap size. It didn't help. I tried the head version of z3. It didn't help.

I noticed however that we might not be the only one facing this issue: https://github.com/Z3Prover/z3/issues/1080 (at least it has the same symptoms). Their current workaround:

> The master branch of batfish disables this parallelism only in OSX.

We could do the same temporarily (it works™).

mantognini:
I could reproduce this with Inox (*) & z3 only. I close this PR and open a different one to workaround  this by disabling //ism on Mac.

(*) I modified Inox's main slightly to add //ism and used as input the `.tip` files Stainless produces with `--debug=solver`.

PR #75, opened 2017-09-26
Update for new Timer API

mantognini:
Adapt stainless to changes introduced in epfl-lara/inox#39. Doesn't compile as-is: `build.sbt` needs to be updated and refer to a newest version of inox.

`TerminationComponent` will be updated after #67 is merged (merge conflict anyway).

### DIFF
--- a/core/src/main/scala/stainless/verification/VerificationChecker.scala
+++ b/core/src/main/scala/stainless/verification/VerificationChecker.scala
@@ -131,10 +131,11 @@ trait VerificationChecker { self =>
         }
       } catch {
         case u: Unsupported =>
-          val t = timer.selfTimer.get
-          val time = if (t.isRunning) t.stop else t.runs.last
+          val time = if (timer.isRunning) timer.stop else timer.time
           reporter.warning(u.getMessage)
           VCResult(VCStatus.Unsupported, Some(s), Some(time))
+      } finally {
+        if (timer.isRunning) timer.stop()
       }
 
       reporter.synchronized {
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,7 @@ val isMac     = osInf.indexOf("Mac") >= 0
 val osName = if (isWindows) "win" else if (isMac) "mac" else "unix"
 val osArch = System.getProperty("sun.arch.data.model")
 
-val inoxVersion = "1.0.2-182-g6bb5560"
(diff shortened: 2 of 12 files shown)


core/src/main/scala/stainless/verification/VerificationChecker.scala
```diff
@@ -131,10 +131,11 @@ trait VerificationChecker { self =>
         }
       } catch {
         case u: Unsupported =>
-          val t = timer.selfTimer.get
-          val time = if (t.isRunning) t.stop else t.runs.last
+          val time = if (timer.isRunning) timer.stop else timer.time
```
samarion: Why don't we use the new API here?

mantognini: The `run` facility unfortunately doesn't give access to the elapsed time and we need it here for the `VCResult`.

samarion: Ah I see, that's too bad. This is fine then.

mantognini: TBH I hesitated to introduce some `tryRun[T](b: => T): (Long, Try[T])` in the `TimerStorage` API. We could still do it, although its usage would be limited to 1 or 2 places. What do you think?

samarion: As you say, the use cases are so rare (for now) that I don't think there's any need for such a method. If in the future we find out we need it more, we can rethink this.

samarion:
I published a new inox version 1.0.2-191-g3b196ab with the latest timer API version.

mantognini:
Thanks, it seems the integration suite went fine!

mantognini:
Unless some one has any concern I'll merge this.

samarion:
I thought you were waiting on https://github.com/epfl-lara/stainless/pull/67?

mantognini:
It's not a big deal, but I'd rather merge this one first because #76 depends on #67, *and* requires the newest inox version too.

samarion:
Ok, sounds good. Just don't forget the termination component :)

mantognini:
Done in https://github.com/epfl-lara/stainless/pull/67/commits/32b0843687f6ccdce9e33e1b89f4fd2f42ce1ad5 ;-)

PR #76, opened 2017-09-26
Improve consistency of command line names & more

mantognini:
Linked to epfl-lara/inox#41. Require #67 (easier than to deal with merge conflict at this point).

Removed `--inlineposts`, which wasn't used in the code at all.

Removed `--parallelvcs`, as it is now always parallel.

Fix internal links in the documentation. The doc is also updated to reflect the new option names and includes some of the missing ones (not the `--print*` options).


core/src/sphinx/gettingstarted.rst
```diff
@@ -5,7 +5,7 @@ Getting Started
 
 Stainless is currently only available as a command line tool,
 which exposes most of the functionality. See the
-`installation documentation <installation.rst>`_
+:doc:`installation documentation <installation>`
```
samarion: Did the previous syntax not work with sphinx? Its main advantage was that it was compatible with Github markdown, so users could navigate through some basic doc without having to go to the website.

frontends/scalac/src/it/scala/stainless/verification/VerificationSuite.scala
```diff
@@ -62,8 +62,7 @@ class InlinedPostsVerificationSuite extends VerificationSuite {
   override def configurations = super.configurations.map {
     seq => Seq(
       inox.optSelectedSolvers(Set("smt-z3")),
-      inox.solvers.optCheckModels(true),
-      extraction.inlining.optInlinePosts(true)
```
samarion: We ended up running into some conceptual issues when discussing this option so it was never fully implemented. Ok with removing it for now, you can also remove the `InlinedPostsVerificationSuite` while we're at it.

core/src/sphinx/gettingstarted.rst
mantognini: Unfortunately, they resulted in 404 error: the HTML link was against the rst file, not the generated html one. To mitigate the issue we could host a proper HTML version of the documentation somewhere.

frontends/scalac/src/it/scala/stainless/verification/VerificationSuite.scala
mantognini: Done.

core/src/sphinx/gettingstarted.rst
samarion: This is very unfortunate but shouldn't be a blocker for this PR. There seems to be [some work](https://gist.github.com/shimizukawa/d62d947509a24f816168) on getting sphinx and markdown to play nice, we should open an issue and investigate.

mantognini:
Just double checking, is everything okay here?

samarion:
It looks fine to me.

mantognini:
Merged in 3e8aa4485761908531f49cab606806bc21f4a777.

PR #78, opened 2017-10-05
Fix bug where class constructor's arguments were not boxed

### LINKED ISSUE #77: Constructor arguments are not boxed properly during type encoding
I'm currently trying to model a typed actor system with Stainless which, as it stands now, requires me to do work with what I would express in plain Scala as a `List[Foo[_]]` (an existential type).

Here's a simple testcase demonstrating the few ways I tried to express it with Stainless, where methods are annotated with the errors being thrown by Stainless:

```scala
object existential {
  case class A()
  case class B()

  sealed abstract class AnyFoo
  case class Foo[T](x: T) extends AnyFoo

  def a: Foo[A] = Foo(A())
  def b: Foo[B] = Foo(B())

  def list0: List[Foo[_]] = List(a, b)
  // Exception in thread "main" stainless.frontend.UnsupportedCodeException: Could not extract type: existential.Foo[_ >: existential.B with existential.A <: Product with Serializable] (class scala.reflect.internal.Types$ExistentialType)

  def list1: List[AnyFoo] = List(a, b)
  // Exception in thread "main" stainless.frontend.UnsupportedCodeException: Could not extract type: existential.Foo[_ >: existential.B with existential.A <: Product with Serializable] (class scala.reflect.internal.Types$ExistentialType)

  def list2: List[AnyFoo] = List(a.asInstanceOf[AnyFoo], b.asInstanceOf[AnyFoo])
  // Insert lenghty inox.ast.TypeOps$TypeErrorException here
}
```

I get why `list0` and `list1` cannot be extracted successfully as we currently lack support for existential types, but I had expected `list2`to succeed.

As another datapoint, here's a testcase which does not involve existential types at all, where methods are again annotated with the errors being thrown:

```scala
object no_existential {
  sealed abstract class A
  case class B() extends A
  case class C() extends A

  case class Foo[+T](x: T)

  def b: Foo[B] = Foo(B())
  def c: Foo[C] = Foo(C())

  def list0: List[Foo[A]] = List(b, c)
  // Insert lengthy inox.ast.TypeOps$TypeErrorException here

  def list1: List[Foo[Any]] = List(b, c)
  // Insert lengthy inox.ast.TypeOps$TypeErrorException here

  def list3: List[Foo[A]] = List(b.asInstanceOf[Foo[A]], c.asInstanceOf[Foo[A]])
  // Insert lengthy inox.ast.TypeOps$TypeErrorException here

  def list4: List[Foo[Any]] = List(b.asInstanceOf[Foo[Any]], c.asInstanceOf[Foo[Any]])
  // Insert lengthy inox.ast.TypeOps$TypeErrorException here
}
```

Is there indeed a problem with the type system that makes the second testcase not type-check successfully or is that that scenerio just unsupported at the moment?

romac:
As discussed in https://github.com/epfl-lara/stainless/issues/77#issuecomment-334435059.

Thank you @jad-hamza and @samarion for taking the time to look into this :)

Closes #77


mantognini:
I didn't look into it but it seems `stainless.ExtractionSuite` fails.

romac:
Damn. Sadly I was not able to run the integration test suite before submitting the PR and thus relied on Larabot to run it for me. I'm getting [compile errors](https://gist.github.com/romac/5df0a6f8fe985877c9659ce1e6f0f145) when compiling the integration suite, that I haven't been able to fix (even after nuking my Ivy cache as well as the coursier cache, and even after disabling coursier). I am the only one with this problem?

romac:
In the meantime, I'll run stainless manually on the extraction benchmarks to see what is going wrong there.

mantognini:
I couldn't reproduce your build issue with your branch on my machine. Maybe try sbt cleanFiles?Sometimes it helps me for weird sbt-related issues.

romac:
Thanks for taking a look! Are tests still failing though?

I have the same issue with a clean checkout of stainless, will try cleanFiles and report back. Thanks!

mantognini:
Yes, unfortunately they fail on my machine like on larabot.

romac:
cleanFiles fixed it :) Thanks, I didn’t know about that command. Will now look into the tests themselves.

romac:
#93 is what makes the test suite fail at the moment.

romac:
@mantognini I have rebased this PR on top of `master` now that #93 has been merged, and tests now  pass :)

mantognini:
Nice!

I've tested the added benchmark. They generate a lot of VC, but I guess it's okay as they are related to casts and they all check. But it seems termination checking doesn't terminate on those programs. This should be investigated (now or later, no objection from me) -- the fact that stainless cannot check termination of `isSubtypeOf`, not the fact that it doesn't terminate itself (for that I've my idea how to make it more robust against timeout violations).

I think the tests would better fit in `stainless/frontends/benchmarks/verification/valid/`.

samarion:
`isSubtypeOf` in the regression tests is HUGE because it's proportional to the number of type definitions in the program. It may be a useful feature to decompose it based on some sort of SCC analysis on the types.

In the mean time, `isSubtypeOf` is also annotated as `@unchecked` so we should have the termination checker filter out termination problems where all functions in the SCC are marked as `@unchecked`, as is the case of the verification checker.

mantognini:
I tried to hack something in the termination checker to ignore unchecked functions but was unsuccessful. I've checked that the `@unchecked` functions are not part of `toVerify` as defined in `TerminationComponent.apply`. I guess those functions are considered later on in the pipeline. So I tried to hack `ProcessingPipeline` in a few places. However, 

 1. Adding `!(fd.flags contains "unchecked") ||` to `isProblem` makes the termination finish instantaneously but with "No Guarantee", and that's probably not what we want.

 2. Adding unchecked functions from the whole `program` to `clearedMap` in the constructor doesn't help.

 3. Updating `terminates` to this doesn't change a thing.

```scala
  def terminates(funDef: FunDef): TerminationGuarantee = {
    val guarantee = {
      {
        if (funDef.flags contains "unchecked") Some(Terminates("unchecked"))
        else None
      } orElse
```

@samarion  Any hints? Because I've no idea how the termination checker work globally...

samarion:
In `ProcessingPipeline.generateProblems`, replace the line
```scala
newProblems.map(fds => Problem(fds.toSeq))
```
by
```scala
val (uncheckedProblems, toCheck) = newProblems.partition(_.forall(_.flags contains "unchecked"))
for (fd <- uncheckedProblems.toSet.flatten) clearedMap(fd) = "Unchecked"
toCheck.map(fds => Problem(fds.toSeq))
```

However, this also means we need to make sure `isSubtypeOf` terminates which might not be the case depending on the type definitions within the program... I'll have to think about this a bit, but I believe that `isSubtypeOf` termination by construction is a feature we want so we can go ahead with ignoring `@unchecked` functions.

mantognini:
@samarion Your patch did the trick, thanks. Beside the regression tests no being in the proper directory, do you have comment on this PR? I think it's pretty straightforward but still, three pairs of eyes are better than two. :)

samarion:
This looks good to me.

Actually I think that we shouldn't put these benchmarks into `verification/valid` for now due to the blowup of `isSubtypeOf`. I have implemented some simplification procedures to alleviate this issue but having to handle such a large function is still troublesome in some cases. The clean fix for this is making the transformation pipeline modular at the function level. We should perform transformations on demand and cache results and provide a final API from which a minimal program can be generated given a function identifier or SCC. This should help avoid the blowup in `isSubtypeOf`. However, this is an involved change and will probably require a significant amount of work.

In the mean time, I would recommend either
1. just keep the benchmarks in extraction and hope that the number of benchmarks there won't have grown too much by the time the above fix is implemented, or
2. create a new verification suite that extracts programs one-by-one (with your `extractOne`, for example). The OO benchmarks should then live in this suite.

mantognini:
@romac If you choose to go with 1/, just move the tests into the `valid` subdirectory.

romac:
> create a new verification suite that extracts programs one-by-one (with your extractOne, for example). The OO benchmarks should then live in this suite.

@samarion Shouldn't we do that for all the extraction benchmarks? Or do some of those split across multiple files?
IIRC there wasn't any problem when I last tried to extract them one-by-one?

> @romac If you choose to go with 1/, just move the tests into the valid subdirectory.

@mantognini I think I will go for this option for now. We can always move there elsewhere later on, and potentially update the test suite as well.

samarion:
Extracting tests one by one implies starting a new instance of the Scala compiler for each file, which used to lead to a major slowdown of the whole suite. However, given that we perform some program transformations on the whole extracted program in individual test cases (in the solvers and evaluators, not the pipeline), the slowdown is probably less noticeable nowadays.

Once the splitting into small programs I mentioned in my previous post happens, we should see the benefits of the speedup again.

samarion:
LGTM, please rebase so we can merge.

romac:
The build failed, but it seems to be one of those random termination test suite failures described in #126.

romac:
Can we get this merged since the build failure is only related to #126 and not this PR?

mantognini:
Merged in 3fc74fc37739f6075b5336b94982a0e70c3c2b5b.

PR #82, opened 2017-10-11
Parallel off for macOS

mantognini:
Built on top of #76 for testing purposes (the crash are much more frequent with the newest changes). This PR disables parallelism on Apple's OS to workaround z3 crashes (as far as I can tell). See Z3Prover/z3#1080 & #74.

### DIFF
--- a/core/src/main/scala/stainless/MainHelpers.scala
+++ b/core/src/main/scala/stainless/MainHelpers.scala
@@ -4,12 +4,11 @@ package stainless
 
 import utils.JsonUtils
 
-import scala.collection.Parallelizable
-import scala.collection.parallel.{ ExecutionContextTaskSupport, ForkJoinTasks, ParIterable }
+import scala.collection.parallel.{ ExecutionContextTaskSupport, ForkJoinTasks }
 import scala.concurrent.ExecutionContext
 
 import java.io.File
-import java.util.concurrent.ExecutorService
+import java.util.concurrent.{ Executors, ExecutorService }
 
 import io.circe.Json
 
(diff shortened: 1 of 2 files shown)


samarion:
I wouldn't be surprised if this issue also occurred on Linux. We had to build a special non-parallel sbt configuration for ScalaZ3 regressions in Leon because it wasn't stable enough when run in parallel.

Since this issue seems pretty common, I think we might as well solve it within Inox. We can add a global synchronization block around all native Z3 solver calls (hopefully just `checkAssumptions` should be enough) and allow parallelism through a flag (no parallelism by default). WDYT?

romac:
This PR has been working fine for me for a while? Can we merge it already?

mantognini:
I'd like to merge this as well. I've been using stainless on linux on and on for a while now and haven't encountered the same bug as on Mac. Therefore, I think it's unnecessary to add additional synchronisation points. Also, diving into the z3 bug itself would be quite time consuming, so this workaround seems good enough.

samarion:
This feature looks fine to me.
I'd also like parallelism to be off when there is an active debug section, but this can be done in a different PR.

Btw I have a few commits coming to Inox soon that remove some lazy vals that create deadlocks in Stainless, so hopefully the regressions will be a bit more stable.

mantognini:
> Btw I have a few commits coming to Inox soon that remove some lazy vals that create deadlocks in Stainless

Nice! :)

> I'd also like parallelism to be off when there is an active debug section, but this can be done in a different PR.

Disabling it based on command line option would require a bit of change in the design, especially if we want to avoid mutable states in `MainHelpers`, namely to move `executor` & `par` to the `Context`. Do you have concerns about that, or maybe a different design in mind?

samarion:
I unfortunately don't see a better way of doing this. The API of Context might need to change a bit in Inox, I'll let you see what you can do :)

mantognini:
Alright. In the meantime I'll merge this one and open an issue to track the desired feature.

PR #84, opened 2017-10-12
Return non-zero exit code if verification fails

romac:
This could be useful when running stainless as part of a CI process.


mantognini:
I think this is indeed something good to have. I've include changes based on yours into #76 mainly because the `MainHelper` class has change a bit, see https://github.com/epfl-lara/stainless/pull/76/commits/7d5a9d5e0a53612068ec3099daf56f322f1cc878.

romac:
Much better, cool :)

PR #89, opened 2017-10-16
Ensure well formed

mantognini:
This requires and extends epfl-lara/inox#44 for stainless usage. This is still WIP but it should fix the traversal of trees and generate proper variable bounds.

### DIFF
--- a/core/src/main/scala/stainless/ast/SymbolOps.scala
+++ b/core/src/main/scala/stainless/ast/SymbolOps.scala
@@ -45,8 +45,9 @@ trait SymbolOps extends inox.ast.SymbolOps { self: TypeOps =>
     *
     * @see [[purescala.Expressions.Pattern]]
     */
-  def conditionForPattern[P <: PathLike[P]](in: Expr, pattern: Pattern, includeBinders: Boolean = false)
-                                           (implicit pp: PathProvider[P]): P = {
+  final def conditionForPattern[P <: PathLike[P]]
+                               (in: Expr, pattern: Pattern, includeBinders: Boolean = false)
+                               (implicit pp: PathProvider[P]): P = {
 
     def bind(ob: Option[ValDef], to: Expr): P = {
       if (!includeBinders) {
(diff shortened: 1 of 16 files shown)


core/src/main/scala/stainless/transformers/TransformerWithPC.scala
```diff
@@ -45,7 +45,7 @@ trait TransformerWithPC extends inox.transformers.TransformerWithPC {
         val patternPathNeg = conditionForPattern[Env](rs, c.pattern, includeBinders = false)
 
         val map = mapForPattern(rs, c.pattern)
-        val sguard = c.optGuard.map(rec(_, soFar))
+        val sguard = c.optGuard.map(rec(_, soFar withBounds c.pattern.binders))
```
samarion: Isn't it something more like `soFar merge patternPathPos`?

core/src/main/scala/stainless/ast/SymbolOps.scala
```diff
@@ -56,42 +57,47 @@ trait SymbolOps extends inox.ast.SymbolOps { self: TypeOps =>
       }
     }
 
-    def rec(in: Expr, pattern: Pattern): P = {
-      pattern match {
-        case WildcardPattern(ob) =>
+    conditionForPatternRecImpl(in, pattern)(pp, bind)
+  }
+
+  protected def conditionForPatternRecImpl[P <: PathLike[P]]
```
samarion: Although I agree this enables better code deduplication, I'd prefer if we just overrode `conditionForPattern` directly, especially because `bind` is not really a parameter to `conditionForPatternRecImpl`. It's already quite hard to navigate the code in these tree refinements, I think it's easier to understand if the override targets are consistent with the original function names. Given how rare the overrides of this method should be, I think it's fine if we duplicate the `bind` function in these cases.

mantognini: I tend to disagree, because it also depends on `Path` and other facilities (and not only on potential future override of this function). Duplicating the code would make it a time bomb.

samarion: But both `conditionForPattern` and `conditionForPatternRecImpl` are parametric in the path. The rec version simply replaces the `includeBinders` parameter with the `bind` implicit parameter. But since the value of `bind` should only ever be the exact argument we provide in `conditionForPattern`, taking it as an argument makes me a bit uneasy. I'd prefer having an auxiliary function 
```scala
protected final def maybeBind[P <: PathLike[P]]
                             (ob: Option[ValDef], to: Expr, includeBinders: Boolean)
                             (implicit pp: PathProvider[P]): P
```
that we call from `conditionForPattern` and all overrides.

mantognini: That would be a better design indeed.

mantognini:
While working on this I've noticed this incorrect translation:

source: `frontends/benchmarks/verification/valid/AssociativityProperties.scala`
```
  def isAssociative[A](f: (A,A) => A): Boolean = {
    forall((x: A, y: A, z: A) => f(f(x, y), z) == f(x, f(y, z)))
  }
```
error triggered on transformed function because variables in `require` are not defined.
```
def isAssociative$0[A$30](f$26: (A$30, A$30) => A$30): Boolean = {
  require(f$26.pre(x$56, y$6) && f$26.pre(y$6, z$9) && f$26.pre(f$26(x$56, y$6), z$9) && f$26.pre(x$56, f$26(y$6, z$9)))
  ∀x$56: A$30, y$6: A$30, z$9: A$30. (f$26(f$26(x$56, y$6), z$9) == f$26(x$56, f$26(y$6, z$9)))
} 
```

It should probably be more something along those lines:
```
def isAssociative[A](f: (A,A) => A): Boolean = {
  require(forall((x: A, y: A) => f.pre(x, y))
  forall((x: A, y: A, z: A) => f(f(x, y), z) == f(x, f(y, z)))
}
```

Am I right?

mantognini:
Overseeded by `latest_inox` branch.

PR #90, opened 2017-10-19
Import missing static checks from Leon's library

mantognini:
See https://github.com/epfl-lara/stainless/commit/142e0e663d8184a84876fdbaa3025ea6b0b59ab6#commitcomment-25072830

### DIFF
--- a/frontends/benchmarks/verification/invalid/StaticChecks2.scala
+++ b/frontends/benchmarks/verification/invalid/StaticChecks2.scala
@@ -1,0 +1,9 @@
+import stainless.lang.StaticChecks._
+
+object StaticChecks2 {
+
+  def foo(n: BigInt, m: BigInt) = {
+    assert(n + m > n)
+  }
+
+}
--- a/frontends/benchmarks/verification/invalid/StaticChecks3.scala
+++ b/frontends/benchmarks/verification/invalid/StaticChecks3.scala
@@ -1,0 +1,10 @@
+import stainless.lang.StaticChecks._
+
+object StaticChecks3 {
+
+  def add(n: BigInt, m: BigInt): BigInt = {
+    require(n >= 0)
+    n + m
+  } ensuring { _ >= 0 }
+
+}
--- a/frontends/dotty/src/main/scala/stainless/frontends/dotc/ASTExtractors.scala
+++ b/frontends/dotty/src/main/scala/stainless/frontends/dotc/ASTExtractors.scala
@@ -501,7 +501,11 @@ trait ASTExtractors {
 
     object ExRequire {
       def unapply(tree: tpd.Apply): Option[tpd.Tree] = tree match {
-        case Apply(ExSymbol("scala", "Predef$", "require"), Seq(body)) => Some(body)
+        case Apply(
+            ExSymbol("scala", "Predef$", "require") | ExSymbol("stainless", "lang", "StaticChecks$", "require"),
+            Seq(body)) =>
+          Some(body)
+
         case _ => None
       }
     }
@@ -515,10 +519,16 @@ trait ASTExtractors {
 
     object ExAssert {
       def unapply(tree: tpd.Apply): Option[(tpd.Tree, Option[String])] = tree match {
-        case Apply(ExSymbol("scala", "Predef$", "assert"), Seq(body)) =>
+        case Apply(
+            ExSymbol("scala", "Predef$", "assert") | ExSymbol("stainless", "lang", "StaticChecks$", "assert"),
+            Seq(body)) =>
           Some((body, None))
-        case Apply(ExSymbol("scala", "Predef$", "assert"), Seq(body, Literal(cnst: Constant))) =>
+
+        case Apply(
+            ExSymbol("scala", "Predef$", "assert") | ExSymbol("stainless", "lang", "StaticChecks$", "assert"),
+            Seq(body, Literal(cnst: Constant))) =>
           Some((body, Some(cnst.stringValue)))
+
         case _ => None
       }
     }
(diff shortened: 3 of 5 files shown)


mantognini:
Also related, the documentation still mentions `setForall`, `setExists` and `setToList` but they weren't imported into stainless. Should they be added to the library or removed from the documentation?

samarion:
Sure, why not. However, I would maybe redefine `setForall` and `setExists` with wrappers and quantifiers. Something like
```scala
implicit class SetWrapper[T](val set: Set[T]) {
  @inline
  def forall(p: T => Boolean): Boolean = forall((e: T) => set.contains(e) ==> p(e))
  @inline
  def exists(p: T => Boolean): Boolean = !set.forall((e: T) => !p(e))
}
```
WDYT? @jad-hamza, you have more experience writing proofs, does this sound useful?

For the static checks, you'll have to implement `require` and `assert` extraction for the `StaticChecks` versions for regressions to pass. I wonder if these shouldn't be `@ignore`ed too...

jad-hamza:
Yes I think this would work We could try writing some examples using Sets and forall/exists to see if proofs go through easily using these notations. (e.g. if an element that satisfies property p is added to a set, then the resulting set should expose this with exists(p)).

romac:
+1

FYI the following verifies on latest master:

```
import stainless.lang._
import stainless.collection._
import stainless.annotation._

object s {

  implicit class SetWrapper[T](val set: Set[T]) {

    @inline
    def forall(p: T => Boolean): Boolean = stainless.lang.forall { (e: T) =>
      set.contains(e) ==> p(e)
    }

    @inline
    def exists(p: T => Boolean): Boolean =
      !set.forall((e: T) => !p(e))

    @extern
    def toList: List[T] = {
      ???.asInstanceOf[List[T]]
    } ensuring { (xs: List[T]) =>
      (xs.content == set && forall((x: T) => set.contains(x) ==> (xs.filter(_ == x).size == 1)))
    }

  }

  def add_exists[A](set: Set[A], x: A, p: A => Boolean): Boolean = {
    require(p(x))
    (set + x) exists p
  } holds

  def add_toList[A](set: Set[A], x: A): Boolean = {
    require(!set.toList.contains(x))
    (set + x).toList.contains(x)
  } holds

  def toList_nodup[A](set: Set[A], x: A): Boolean = {
    (set + x + x).toList.filter(_ == x).size == 1
  } holds

}
```

While this doesn't (`choose satisfiability` for `toList` times out):

```scala
import stainless.lang._
import stainless.collection._

object s {

  implicit class SetWrapper[T](val set: Set[T]) {

    @inline
    def forall(p: T => Boolean): Boolean = stainless.lang.forall { (e: T) =>
      set.contains(e) ==> p(e)
    }

    @inline
    def exists(p: T => Boolean): Boolean =
      !set.forall((e: T) => !p(e))

    @inline
    def toList: List[T] = choose { (xs: List[T]) =>
      xs.content == set && forall { (x: T) =>
        set.contains(x) ==> (xs.filter(_ == x).size == 1)
      }
    }
  }

  def add_exists[A](set: Set[A], x: A, p: A => Boolean): Boolean = {
    require(p(x))
    (set + x) exists p
  } holds

  def add_toList[A](set: Set[A], x: A): Boolean = {
    require(!set.toList.contains(x))
    (set + x).toList.contains(x)
  } holds

  def toList_nodup[A](set: Set[A], x: A): Boolean = {
    (set + x + x).toList.filter(_ == x).size == 1
  } holds

}
```

Is there another way of writing `toList` without invoking `@extern` that would still verify? It seems to me that Stainless's `Set`s are intensional, so I currently don't see how to write `toList` without either `choose` or `@extern`.

romac:
Also, having `def keys: List[A]`, `def values: List[B]`, and `def toList: List[(A, B)]` on `Map[A, B]` would be similarly useful.

samarion:
@romac: due to some soundness issues with chooses - namely you can prove anything by using
```scala
val proofForFalse = choose(_ => false)
```
we require evidence that a choose is indeed satisfiable. However, this means that the user basically has to give an implementation for the choose, or we need support for general synthesis in Stainless (Leon synthesis hasn't (yet?) been ported).

However in this particular case, even with Leon synthesis we couldn't avoid the `@extern` annotation. The `Set.toList` method cannot be defined in the computable Stainless fragment (basically Stainless without chooses and quantifiers) since Stainless sets don't have iterators or selectors (can't get an element from the set). You could possibly define it somehow using chooses as a set selector:
```scala
def setToList[T](set: Set[T]): List[T] = {
  if (set == Set[T]()) {
    Nil[T]()
  } else {
    val element = choose((x: T) => set contains x) // not sure Stainless will prove this.
    val rest = set - element
    Cons(element, setToList(rest))
  }
}
```
@jad-hamza might have some ideas to get the non-empty set element existence check to go through.

@mantognini: please extend extraction to handle these new contract specifications and we can merge this PR. The extensions to the Set and Map APIs can be implemented in a new PR.

jad-hamza:
Stainless gives counter-examples (not valid with --check-models) for these examples. Should it return unknown instead? How is Set[T] encoded? Is it always as T => Bool or does it depend on the backend?

```scala
  def notEmptyImpliesExists[T](s: Set[T]) = {
    require(s != Set[T]())

    exists((x: T) => s.contains(x))
  } holds

  def notExistsImpliesEmpty[T](s: Set[T]) = {
    require(forall((x: T) => !s.contains(x)))

    s == Set[T]()
  } holds
```

mantognini:
> @mantognini: please extend extraction to handle these new contract specifications and we can merge this PR. The extensions to the Set and Map APIs can be implemented in a new PR.

> For the static checks, you'll have to implement require and assert extraction for the StaticChecks versions for regressions to pass. 

Should be addressed now. :-)

samarion:
Could you please add a `require` and `ensuring` in the test?

Other than that, LGTM, thanks!

PR #91, opened 2017-10-24
Handles exceptions when opening and closing cache file

jad-hamza:
Currently, a corrupt cache file creates an execution when creating the ObjectInputStream.
This commit allows to catch the StreamCorruptedException when creating the ObjectInputStream.
I also added something to catch exceptions when closing the stream (IOException).

### DIFF
--- a/core/src/main/scala/stainless/verification/VerificationCache.scala
+++ b/core/src/main/scala/stainless/verification/VerificationCache.scala
@@ -75,7 +75,7 @@ trait VerificationCache extends VerificationChecker { self =>
   }
 
   private val cacheFile: File = utils.Caches.getCacheFile(context, "vccache.bin")
-  private val vccache: Cache = CacheLoader.get(cacheFile)
+  private val vccache: Cache = CacheLoader.get(context, cacheFile)
 
   private def contains(p: (Symbols, Expr)) = vccache contains serialize(p)
   private def add(p: (Symbols, Expr)) = vccache addPersistently serialize(p)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/stainless/verification/VerificationCache.scala
```diff
@@ -127,26 +127,52 @@ private object CacheLoader {
...
    *      while being loaded!
    */
-  def get(cacheFile: File): Cache = this.synchronized {
+  def get(ctx: inox.Context, cacheFile: File): Cache = this.synchronized {
     db.getOrElseUpdate(cacheFile, {
       val cache = new Cache(cacheFile)
 
       if (cacheFile.exists) {
-        val ois = new ObjectInputStream(new FileInputStream(cacheFile))
+        val ois: Option[ObjectInputStream] = openStream(ctx, cacheFile)
```
mantognini: If this fails, an empty `Cache` will be added to `db`. Is that intentional?

core/src/main/scala/stainless/verification/VerificationCache.scala
```diff
@@ -127,26 +127,52 @@ private object CacheLoader {
...
+
         try {
-          while (true) {
-            val s = ois.readObject.asInstanceOf[String]
+          while (ois != None) {
```
mantognini: I think the `while (true)` was more readable than this non-trivial *if not None, do forever X*. The code could probably be rewritten to pattern match on `openStream` result instead. What do you think?

mantognini:
Merged in 97b64c15f36ea87dc0959bce147d329e7f4b4f60.

PR #93, opened 2017-10-30
Add missing case for type `Any` in `dotc.CodeExtraction`

romac:
The type `Any` was previously not extracted, which would yield the following error, when referencing it:

```
[ Error  ] foo$0 depends on missing dependencies: Any$0.
[Internal] Missing some nodes in Registry: Any$0
[Internal] Please inform the authors of Inox about this message
```

This seems to be what is currently making #77's test suite fail.

### DIFF
--- a/frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
+++ b/frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
@@ -1322,6 +1322,10 @@ class CodeExtraction(inoxCtx: inox.Context, cache: SymbolsContext)(implicit val 
     case tpe if tpe.typeSymbol == defn.UnitClass    => xt.UnitType()
     case tpe if tpe.typeSymbol == defn.NothingClass => xt.NothingType()
 
+    // `isRef` seems to be needed here instead of `==`, as the latter
+    // seems to be too lax, and makes the whole test suite fail. - @romac
+    case tpe if tpe.isRef(defn.AnyClass)            => xt.AnyType()
+
     case tpe if isBigIntSym(tpe.typeSymbol) => xt.IntegerType()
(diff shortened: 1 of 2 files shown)


frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
```diff
@@ -1322,6 +1322,10 @@ class CodeExtraction(inoxCtx: inox.Context, cache: SymbolsContext)(implicit val
     case tpe if tpe.typeSymbol == defn.UnitClass    => xt.UnitType()
     case tpe if tpe.typeSymbol == defn.NothingClass => xt.NothingType()
 
+    // `isRef` seems to be needed here instead of `==`, as the latter
+    // seems to be too lax, and makes the whole test suite fail. - @romac
+    case tpe if tpe.isRef(defn.AnyClass)            => xt.AnyType()
```
mantognini: I've just tested it with `isDirectRef` and it seems to work too. Not sure it should be changed though.

romac: Yeah I tested that as well, but ended up choosing `isRef` since it's more commonly used than `isDirectRef` in the Dotty codebase:

```shell
$ cd dotty/compiler
$ rg '\bisRef\b'       | wc -l                     // 124
$ rg '\bisDirectRef\b' | wc -l                     //  7
```

romac:
> I'm wondering if usages of `def getResolvedTypeSym(sym: Symbol): Symbol` to compare type symbols should be replaced by `isRef/isDirectRef` instead...

I will give it a try and report back.

romac:
Can we merge this already now that the test suite passes? Haven’t had time to look further into the ‘getResolvedTypeSym’ thing yet, so we can maybe go ahead with this PR already?

Issue #96, opened 2017-11-04
Unrolling performance and values


jad-hamza:
```scala
import stainless.lang._ // for the holds keyword
import scala.language.postfixOps // to avoid warnings about postfix holds

object Factorial {
  sealed abstract class Nat
  case object O extends Nat
  case class S(n: Nat) extends Nat

  val one = S(O)
  val two = S(one)
  val three = S(two)
  val four = S(three)
  val five = S(four)
  val six = S(five)
  val seven = S(six) 
  val eight = S(seven)
  val nine = S(eight)
  val ten = S(nine)
  val eleven = S(ten)
  val twelve = S(eleven)

  def plus(n: Nat)(m: Nat): Nat = n match {
    case O => m
    case S(n2) => S(plus(n2)(m))
  }

  def mult(n: Nat)(m: Nat): Nat = n match {
    case O => O
    case S(n2) => plus(m)(mult(n2)(m))
  }

  def factorial(n: Nat): Nat = n match {
    case O => S(O)
    case S(n2) => mult(n)(factorial(n2))
  }
  
  def test_factorial2() = { factorial(five) == mult(ten)(twelve) } holds
}
```

This is an example taken from the software foundations book. By computation, we can easily verify that `test_factorial2` holds, but in Stainless, that takes 2 or 3 minutes to verify due to the SMT encoding. I wonder if we could do better by using the evaluator on expressions where all inputs are known. The problem is that we don't know whether expressions terminate before running the termination checker, so we could end up in an infinite loop.

romac:
With #104, the following testcase verifies in 0.777 seconds on my machine with `stainless --debug=eval Factorial.scala`:

```scala
import stainless.lang._ // for the holds keyword
import stainless.lang.eval._ // for the force keyword
import scala.language.postfixOps // to avoid warnings about postfix holds

object Factorial {
  sealed abstract class Nat
  case object O extends Nat
  case class S(n: Nat) extends Nat

  val one = S(O)
  val two = S(one)
  val three = S(two)
  val four = S(three)
  val five = S(four)
  val six = S(five)
  val seven = S(six)
  val eight = S(seven)
  val nine = S(eight)
  val ten = S(nine)
  val eleven = S(ten)
  val twelve = S(eleven)

  def plus(n: Nat)(m: Nat): Nat = n match {
    case O => m
    case S(n2) => S(plus(n2)(m))
  }

  def mult(n: Nat)(m: Nat): Nat = n match {
    case O => O
[...]

PR #104, opened 2017-11-09
[WIP] Add a construct to force evaluation of an expression

romac:
This PR adds a `force` function in `stainless.lang.eval` that forces the evaluation of its argument  right after the extraction phase.

On the current [`master`](https://github.com/epfl-lara/stainless/commit/3e8aa4485761908531f49cab606806bc21f4a777), the [attached benchmark](https://github.com/epfl-lara/stainless/compare/master...romac:force-eval?expand=1#diff-d6cef4e50e8e468080734b4f73982445R171) timeouts after 10min, while with this commit it verifies in about 2min.

This PR also adds an `eval` debug section which prints the expressions that have been evaluated, and thus can also be useful for debugging purposes.

This is a WIP, and currently only works when there are no <del>free variables</del> "abstract" sub-expressions in the expression being evaluated, but I wonder whether it is possible to use some kind of technique that would evaluate an expression as much as possible, ie. until it gets stuck because it would to need to evaluate a free variable to go further. Perhaps a lazy evaluator that could reduce under lambdas etc.?

I have a hunch that doing so might help us prove typeclass laws etc., in the same vein as this [GHC compiler plugin](https://github.com/nomeata/ghc-proofs).

What do you think?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,7 @@ val isMac     = osInf.indexOf("Mac") >= 0
 val osName = if (isWindows) "win" else if (isMac) "mac" else "unix"
 val osArch = System.getProperty("sun.arch.data.model")
 
-val inoxVersion = "1.0.2-191-g3b196ab"
+val inoxVersion = "1.0-290-gfa92c8a"
 val dottyVersion = "0.1.1-bin-20170429-10a2ce6-NIGHTLY"
 
 lazy val nParallel = {
--- a/core/src/main/scala/stainless/MainHelpers.scala
+++ b/core/src/main/scala/stainless/MainHelpers.scala
@@ -84,6 +84,7 @@ trait MainHelpers extends inox.MainHelpers {
     termination.DebugSectionTermination,
     DebugSectionExtraction,
     frontend.DebugSectionFrontend,
+    forcing.DebugSectionForcing,
     utils.DebugSectionRegistry
   )
 
--- a/core/src/main/scala/stainless/ast/Definitions.scala
+++ b/core/src/main/scala/stainless/ast/Definitions.scala
@@ -8,11 +8,13 @@ import scala.collection.mutable.{Map => MutableMap}
 
 trait Definitions extends inox.ast.Definitions { self: Trees =>
 
+  case object Force extends Flag("force", Seq())
   case object Extern extends Flag("extern", Seq.empty)
   case object Unchecked extends Flag("unchecked", Seq.empty)
   case class Derived(id: Identifier) extends Flag("derived", Seq(id))
 
   override def extractFlag(name: String, args: Seq[Any]): Flag = (name, args) match {
+    case ("force", Seq()) => Force
     case ("extern", Seq()) => Extern
     case ("unchecked", Seq()) => Unchecked
     case _ => super.extractFlag(name, args)
--- a/core/src/main/scala/stainless/ast/Extractors.scala
+++ b/core/src/main/scala/stainless/ast/Extractors.scala
@@ -191,6 +191,7 @@ trait TreeDeconstructor extends inox.ast.TreeDeconstructor {
   }
 
   override def deconstruct(f: s.Flag): DeconstructedFlag = f match {
+    case s.Force => (Seq(), Seq(), Seq(), (_, _, _) => t.Force)
     case s.Extern => (Seq(), Seq(), Seq(), (_, _, _) => t.Extern)
     case s.Unchecked => (Seq(), Seq(), Seq(), (_, _, _) => t.Unchecked)
     case s.Derived(id) => (Seq(id), Seq(), Seq(), (ids, _, _) => t.Derived(ids.head))
--- a/core/src/main/scala/stainless/extraction/inlining/package.scala
+++ b/core/src/main/scala/stainless/extraction/inlining/package.scala
@@ -4,7 +4,7 @@ package stainless
 package extraction
 
 package object inlining {
-  
+
   object trees extends Trees with inox.ast.SimpleSymbols {
     case class Symbols(
       functions: Map[Identifier, FunDef],
--- a/core/src/main/scala/stainless/forcing/Forcing.scala
+++ b/core/src/main/scala/stainless/forcing/Forcing.scala
@@ -1,0 +1,72 @@
+/* Copyright 2009-2017 EPFL, Lausanne */
+
+package stainless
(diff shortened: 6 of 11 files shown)


romac:
@jad-hamza It seems to me that this PR could help with #96, what do you think?

romac:
@jad-hamza Well, I just gave it a try and [it now verifies in < 1 second](https://github.com/epfl-lara/stainless/issues/96#issuecomment-343250986). \o/

romac:
@samarion Where do you think is the right place to run this transformer? AFAICT it needs to be run after extraction as ‘Evaluator’ requires a ‘Program { val trees: ast.Trees }’, which is why I put it within ‘VerificationComponent’, but maybe there’s better way?

samarion:
I was thinking of adding this within Inox but only in certain "safe" cases (see @jad-hamza's comment about soundness). I was thinking of avoiding quantifiers and impure expressions as a rule of thumbs.

I believe it would indeed be useful to have a user annotation, but it's not clear to me exactly what shape it should have. In order to make sure everything stays as sound and consistent as possible, it would be ideal if we could have the annotation at the function level. This would ensure that if C[e] = C[e'], then adding more force annotations won't change that (though removing them could...)

For simplifications of non-ground terms, I also agree with @jad-hamza. Note that Inox actually already does this pretty extensively in [SimplifierWithPC](https://github.com/epfl-lara/inox/blob/master/src/main/scala/inox/transformers/SimplifierWithPC.scala). However, no beta-reduction is performed for named functions, so that could be an interesting new simplification to consider. The challenge is to choose good heuristics about when a function inlining can be considered 'useful'. Possibly basing ourselves off the simp tactic could indeed be useful, but I would prefer automatic this as much as possible.

About the place, you can build a program from symbols simply as `inox.Program(trees)(symbols)`. What you are missing is an instance of `Context`. See `PreconditionInference` and [extraction/package.scala](https://github.com/epfl-lara/stainless/blob/master/core/src/main/scala/stainless/extraction/package.scala#L64) for an example of how you can get context during the transformation pipeline.

romac:
> I was thinking of adding this within Inox but only in certain "safe" cases (see @jad-hamza's comment about soundness). I was thinking of avoiding quantifiers and impure expressions as a rule of thumbs.

Yeah I too was wondering whether this would better fit within Inox. Should I move it there? Seems like it would be a better fit indeed since it comes after extraction anyway.

> I believe it would indeed be useful to have a user annotation, but it's not clear to me exactly what shape it should have. In order to make sure everything stays as sound and consistent as possible, it would be ideal if we could have the annotation at the function level. This would ensure that if C[e] = C[e'], then adding more force annotations won't change that (though removing them could...)

I thought about introducing it as an annotation as well, though I am not sure I understand why doing it this way keeps the semantics intact vs. as a function. Do you have an example in mind?

I understand that reducing under lambdas changes the semantics, but I would think that forcing evaluation using Stainless's `Evaluator` does not change the semantics, right?

> For simplifications of non-ground terms, I also agree with @jad-hamza. Note that Inox actually already does this pretty extensively in SimplifierWithPC. However, no beta-reduction is performed for named functions, so that could be an interesting new simplification to consider. The challenge is to choose good heuristics about when a function inlining can be considered 'useful'. Possibly basing ourselves off the simp tactic could indeed be useful, but I would prefer automatic this as much as possible.

Okay, I am gonna think about this and come back to you hopefully soon with a concrete proposal, most likely inspired by `simpl`. When you say "automate this", you mean integrating this within `SimplifierPC` on all terms, rather than require an annotation/construct?

> About the place, you can build a program from symbols simply as inox.Program(trees)(symbols). What you are missing is an instance of Context. See PreconditionInference and extraction/package.scala for an example of how you can get context during the transformation pipeline.

Yeas but it seems to me that stainless' evaluator requires a `StainlessProgram` == `  type StainlessProgram = Program { val trees: stainless.trees.type }`, which I cannot get until extraction is completed and lowering has been performed, right?

romac:
@jad-hamza Can you expand on that:

> Perhaps it's not a big deal if force is only done when the user asks for it. Maybe force could be interpreted as syntactic sugar for a higher-level let binding.

samarion:
I think I would define a new simplifier in `SymbolOps`, something like:
```scala
def simplifyGround(e: Expr)(implicit sem: symbols.Semantics, ctx: Context): Expr = {
  // 1. Get evaluator from semantics.
  // 2. Traverse e and evaluate all pure expressions.
  // 3. Either check that the pure expressions cannot refer to a quantifier, or add an
  //    option to the evaluator to make it crash when it encounters a quantifier
  //    or choose.
  // 4. Make sure the evaluator will terminate (use inox.evaluators.optMaxCalls)
}
```

You can define the simp simplifier as an other simplifier in `SymbolOps` as well.

The issue with semantics is as follows. Consider the closures
```scala
L1 := (x: Int) => x + {
  assert(something hard to prove but ground)
  1
}
```
and
```scala
L2 := (x: Int) => x + 1
```
The simplifier will be conservative when encountering `assert(something hard to prove but ground)` and will assume this piece of code is possibly impure. The structure of L1 and L2 is therefore different (note that impurity also concerns function calls whose termination isn't known), so the closures are not equal.
Now if you wrap the assertion and `1` constant with a force as
```scala
force({
  assert(something hard to prove but ground)
  1
})
```
your force simplifier will replace this by `1` (assuming the contract actually held) and your closures will become equal all of a sudden.
For the other way around, simply consider L1' the closure L1 where the assertion and `1` constant have been forced. We have L1 = L1 but L1 != L1'.

About automation, yes I meant getting it to work without an annotation.

jad-hamza:
> @jad-hamza Can you expand on that?

I was thinking that:
`(x: Int) => x + force { assert(something hard to prove but ground); 1 }`
could be understood as syntactic sugar for:
`val v = assert(something hard to prove but ground); 1
(x: Int) => x + v`

So it would explain why `(x: Int) => x + force { assert(something hard to prove but ground); 1 }` is different than `(x: Int) => x + { assert(something hard to prove but ground); 1 }` in the Stainless semantics. As long as the user writes the `force` or `simpl` on purpose then we can explain the difference. If we do it automatically we have to be careful not to do it here.

@samarion If we implement an automatic simplifier like simpl, would we make an entirely new simplifier? (which could potentially solve https://github.com/epfl-lara/inox/issues/37)

romac:
See #108 and epfl-lara/inox#48.

Issue #83, opened 2017-10-12
Extraction test suite and verification pipeline yield different results


romac:
With my branch in #78, the following benchmark yields different results with the extraction test suite and the verification pipeline that runs when invoking stainless on its own (either from the command-line or through `sbt runMain`.

```scala
object ConstructorArgsBoxing2 {
  case class A()
  case class B()

  sealed abstract class AnyFoo
  case class Foo[T](x: T) extends AnyFoo

  val a: AnyFoo = Foo(A())
  val b: AnyFoo = Foo(B())

  val list0: List[AnyFoo] = List(a, b)
}
```

I tried [performing the extraction](https://github.com/romac/stainless/commit/b4d5a68bb0c7e583c5b06715a6c82f2646c31836#diff-4dfbdfe43e0434f80afd08b8a1174910R43) in the test suite using [`VerificationComponent.extract`](https://github.com/romac/stainless/blob/2260e06824625743d0a0aea6dda5edab8560adac/core/src/main/scala/stainless/verification/VerificationComponent.scala), but am still getting the same error.

Any idea what could be going wrong there?

#### `stainless-scala/it:testOnly stainless.ExtractionSuite`

```
(…snip…)

[info]   Extraction of 'extraction/ConstructorArgsBoxing2.scala'
[info]   - should be successful (0 milliseconds)
[info]   - should typecheck (12 milliseconds)
[info]     and transformation
====== ERROR ======
java.util.NoSuchElementException: key not found: apply
	at scala.collection.MapLike$class.default(MapLike.scala:228)
	at scala.collection.AbstractMap.default(Map.scala:59)
	at scala.collection.MapLike$class.apply(MapLike.scala:141)
	at scala.collection.AbstractMap.apply(Map.scala:59)
	at stainless.extraction.oo.TypeEncoding$Scope$2.rewrite(TypeEncoding.scala:724)
	at stainless.extraction.oo.TypeEncoding$Scope$2.transform(TypeEncoding.scala:849)
	at stainless.extraction.oo.TypeEncoding$$anonfun$96.apply(TypeEncoding.scala:1074)
	at stainless.extraction.oo.TypeEncoding$$anonfun$96.apply(TypeEncoding.scala:1074)
	at scala.Option.map(Option.scala:146)
	at stainless.extraction.oo.TypeEncoding$class.transformFunction$1(TypeEncoding.scala:1074)
[...]

Issue #103, opened 2017-11-09
Warnings or errors when printing


jad-hamza:
These are various warnings or errors I get when using println.

```scala
object SI_Print1 {
  println("test")
}
```
> [Warning] Could not extract tree in static container: scala.this.Predef.println("test")

```scala
object SI_Print2 {
  def test() = {
    println("test")
  }
}
```
> [Internal] Missing some nodes in Registry: println$4

```scala
object SI_Print3 {
  stainless.io.StdOut.print(5)
}
```
> [Warning] Could not extract tree in static container: stainless.io.StdOut.print(5)

`SI_Print3` actually works fine when the print is not at the top-level.

```scala
import stainless.annotation._

object SI_Print4 {
  @library
  def pp(a: Any) = println(a)
}
```
> [Internal] Missing some nodes in Registry: println$4

```scala
import stainless.annotation._

object SI_Print5 {
  @library
  @extern
  def pp(a: Any) = println(a)
  pp("Hello")
}
```
> [Warning] Could not extract tree in static container: SI_Print4.this.pp("Hello")

 ```scala
import stainless.annotation._

object SI_Print6 {
  @library
  @extern
  def pp(a: Any) = println(a)

  def test() = {
    pp("Hello")
  }
}
```

> pp("Hello") is of type <untyped>
>  "Hello" is of type String because pp was instantiated with  with type (Object) => Unit

jad-hamza:
Just noticed a typo in the last error message: "with with", from Inox:
https://github.com/epfl-lara/inox/blob/master/src/main/scala/inox/ast/SymbolOps.scala#L1265-L1267

samarion:
Well, I can't really call these _bugs_...

1. You can't define any state mutation within an object or class body. The imperative transformations don't support this. You could theoretically call a pure function there, but that doesn't really make much sense to we just don't allow any calls within a body's definition. I guess the warning could be made a bit clearer.
2. `println` is not defined within the Stainless library and therefore can't be extracted unless we add dedicated support.
3. Same problem as 1.
4. Same problem as 2.
5. Same problem as 1.
6. This is indeed a bug, but bugs in Scala's type system support are expected since I've only had time to implement about half of it.

[...]

PR #105, opened 2017-11-10
Reject usage of Scala's standard List

mantognini:
This PR follows from the discussion in #83. With it, this is now rejected:

```scala
//no import of stainless.collection.List
object NoStdList {
  // No usage of List[T]
  def bar0[T](l: List[T]) = {
    require(l.size > 0)
    l.head
  }

  // No usage of Nil
  def bar1 = Nil

  // No usage of ::
  def bar2 = 1 :: Nil

  // No usage of List.apply
  def bar3 = List(1, 2)
}
```

This patch is probably not rejecting all List usage but it should be a
first step in the right direction, especially to help user detect
missing import statements.


romac:
👍 Sweet, been bitten by this too many times!

samarion:
Looks good! Could you add some rejection tests for Stainless frontends and add the fix to dotty too, please?

jad-hamza:
Thanks! In general is it possible to have a white list of things we accept, rather than a black list of things we reject?

mantognini:
@samarion I'll have a look. IIRC we don't have a test suite to check something is rejected at the extraction.

@jad-hamza We kind of have it already, but the error message is not obvious: if the user uses something from the Scala library they will get a missing dependency from the Registry error. It's reported as an internal error, which is not really correct. Maybe this error message could be changed to be more understandable... I'll try something. :-)

samarion:
@mantognini Indeed we don't have such a suite yet, but I believe it's high time we did. Many of @jad-hamza's issues would contribute examples to put in there.

jad-hamza:
This PR doesn't affect the examples in https://github.com/epfl-lara/stainless/issues/103, I thought the `println` issue was related, as it doesn't belong to the "white list"?

mantognini:
This PR is only dedicated to List & cie. I might work on a second PR for that later on. ;-)

PR #108, opened 2017-11-13
Allow forcing the evaluation of ground function calls

romac:
- Add a `@force` annotation which forces the evaluation of the function body.
- Add a `forcing` debug section which prints which functions are being forced, as well as the result of the evaluation.

This PR depends on https://github.com/epfl-lara/inox/pull/48.

### DIFF
--- a/core/src/main/scala/stainless/MainHelpers.scala
+++ b/core/src/main/scala/stainless/MainHelpers.scala
@@ -84,6 +84,7 @@ trait MainHelpers extends inox.MainHelpers {
     termination.DebugSectionTermination,
     DebugSectionExtraction,
     frontend.DebugSectionFrontend,
+    forcing.DebugSectionForcing,
     utils.DebugSectionRegistry
   )
 
--- a/core/src/main/scala/stainless/ast/Definitions.scala
+++ b/core/src/main/scala/stainless/ast/Definitions.scala
@@ -8,11 +8,13 @@ import scala.collection.mutable.{Map => MutableMap}
 
 trait Definitions extends inox.ast.Definitions { self: Trees =>
 
+  case object Force extends Flag("force", Seq())
   case object Extern extends Flag("extern", Seq.empty)
   case object Unchecked extends Flag("unchecked", Seq.empty)
   case class Derived(id: Identifier) extends Flag("derived", Seq(id))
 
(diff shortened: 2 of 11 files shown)


romac:
I am currently doing this in `VerificationComponent.apply` as I am otherwise unable to obtain a valid instance of `inox.Semantics`, which is required to call `inox.SymbolsOps.simplifyGround`.
It does not seem possible to get one without having a `StainlessProgram` on hand (as opposed to a Program { val trees: SOME_PHASE.trees.type }`), which I don't have until that point.

samarion:
Ah, I hadn't thought about the lowering. I think it would be worth building an `EncodingEvaluator` or extending the evaluators to `extraction.trees` so that forcing can be done at the end of the pipeline. It will probably be useful to enable forcing for the termination checker as well in some cases. You can probably use a fold of all lowerings for the encoding evaluator.

On a more general note, this is not exactly how I envisioned this feature. I would add the `simplifyGround` method to `simplifyFormula` in Inox which will run it on all "safe" expressions sent to the solver. The force annotation would signal to Stainless that ground calls to the function should be considered _safe_ to force, even when they contain expressions we would usually consider as "unsafe". I expect the simplification of safe ground expressions to be fairly efficient so I think we can run it on all ground expressions without impacting the baseline much.

romac:
I am getting some strange errors when trying to evaluate certain expressions, such as in [this testcase](https://gist.github.com/romac/b548377193f3264aa561d7a0c221383c). I have poked at both evaluators a bit already without any success, but I will keep trying.

romac:
@jad-hamza  @samarion Do you guys agree that https://gist.github.com/romac/b548377193f3264aa561d7a0c221383c is symptomatic of some kind a bug in the both evaluators (or even in stainless/inox)?

samarion:
Yes, this does seem like a bug. I'd take a look at `SymbolOps.matchToIfThenElse`.

romac:
I updated the PR with the changes discussed above. Note that it now depends on https://github.com/epfl-lara/inox/pull/53/.

romac:
Any idea why the latest build failed? It looks like sbt got killed, but the log does not show why.

romac:
This has been superseded by both the `--eval` flag as well as the upcoming #193.

PR #110, opened 2017-11-16
Adapt code to new Timer API

mantognini:
Taking advantage of epfl-lara/inox#49 to improve code in two places & print runtime of each main cycle.


mantognini:
FYI I've rebased this onto `latest_inox`.

mantognini:
Merged in 7be0c387eb3718bcf3cfeb88beedfda07fdcce8e.

Issue #92, opened 2017-10-26
Stainless yields lambdas with unsatisfiable preconditions (ie. an invalid model)


romac:
@jad-hamza and I ran into the following issue today:

The following testcase yields a counter-example with an unsatisfiable precondition, which is catched by `--check-models` as an invalid model.

```scala
import stainless.lang._

object util {
  @inline
  def total[A, B](f: A => B): Boolean = forall { (a: A) =>
    f.pre(a)
  }
}

case class Cont[R, A](run: (A => R) => R) {

  import util._

  require(total(run))

  def map[B](f: A => B): Cont[R, B] = {
    require(total(f))

    Cont { (k: B => R) =>
      // require(true) // TODO: Report issue
      this.run(x => k(f(x)))
    }
  }
}
```

```
(snip...)

[  Info  ]  - Now solving 'lambda precondition' VC for map @24:21...
[ Debug  ] @unchecked ∀a: A. f.pre(a) ==> k.pre(f(x))
[ Debug  ] Solving with: nativez3
[  Info  ]  - Result for 'lambda precondition' VC for map @24:21:
[Warning ]  => INVALID
[Warning ] Found counter-example:
[Warning ]   x: A        -> A#0
[Warning ]   k: (B) => R -> (x$524: B) => {
[Warning ]     require(if (x$524 == B#0) {
[Warning ]       false
[Warning ]     } else {
[Warning ]       false
[Warning ]     })
[Warning ]     R#0
[Warning ]   }
[Warning ]   f: (A) => B -> (x$541: A) => if (x$541 == A#0) {
[Warning ]     B#0
[Warning ]   } else {
[Warning ]     B#0
[Warning ]   }

(snip...)
```
On the other hand, explicitly specifying `require(true)` as a precondition inside of the lambda makes the testcase validates.

```scala
import stainless.lang._

object util {
  @inline
  def total[A, B](f: A => B): Boolean = forall { (a: A) =>
    f.pre(a)
  }
}

case class Cont[R, A](run: (A => R) => R) {

  import util._

  require(total(run))

  def map[B](f: A => B): Cont[R, B] = {
    require(total(f))

    Cont { (k: B => R) =>
      require(true)
      this.run(x => k(f(x)))
    }
  }
}
```

```
(snip...)

[  Info  ]  - Now solving 'lambda precondition' VC for map @24:23...
[ Debug  ] @unchecked ∀a: A. f.pre(a) ==> f.pre.pre(x)
[...]

PR #111, opened 2017-11-16
WIP: Make lambdas total by default

romac:
Following the discussion in #92.

### DIFF
--- a/core/src/main/scala/stainless/ast/Expressions.scala
+++ b/core/src/main/scala/stainless/ast/Expressions.scala
@@ -88,19 +88,6 @@ trait Expressions extends inox.ast.Expressions with inox.ast.Types { self: Trees
   }
 
 
-  /* First-class function specs */
-
-  /** First class function precondition accessor
-    * 
-    * $encodingof `f.pre` */
-  case class Pre(f: Expr) extends Expr with CachingTyped {
-    protected def computeType(implicit s: Symbols): Type = f.getType match {
-      case FunctionType(from, to) => FunctionType(from, BooleanType()).unveilUntyped
-      case _ => Untyped
-    }
-  }
-
-
   /* Pattern-match expression */
 
   /** $encodingof `... match { ... }`
--- a/core/src/main/scala/stainless/ast/Extractors.scala
+++ b/core/src/main/scala/stainless/ast/Extractors.scala
@@ -51,9 +51,6 @@ trait TreeDeconstructor extends inox.ast.TreeDeconstructor {
       (Seq(), Seq(), Seq(body, pred), Seq(), (_, _, es, _) => t.Ensuring(es(0), es(1).asInstanceOf[t.Lambda]))
     case s.Assert(pred, error, body) =>
       (Seq(), Seq(), Seq(pred, body), Seq(), (_, _, es, _) => t.Assert(es(0), error, es(1)))
-    case s.Pre(f) =>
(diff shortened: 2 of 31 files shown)


jad-hamza:
Thanks for making that branch. [This example](https://gist.github.com/jad-hamza/f09dca5bf54d6d25cbf2160c8d7ec7ff) is still producing an invalid model, I can investigate later.

> [Warning ] - Model leads to runtime error: Reached empty tree in evaluation @46:3

romac:
Yes there's still quite a few problem with those changes. And the [counting example](https://gist.github.com/romac/80ee13f9e4425098e236c7d31b4895f1) we wrote yesterday still yields an invalid counterexample, but with the `require(false)` now at least.

jad-hamza:
Yes the gist is the Counting example, but a bit simplified.

romac:
See https://gist.github.com/romac/7d8ee52865dbd41e66e34164c0c102c4

jad-hamza:
Have you tried the one I linked? It could be due to non-determinism in the solver

romac:
Yes that's the one I posted the log of above. Yeah it could be that or that we don't have the same Z3 version.

jad-hamza:
I updated https://gist.github.com/jad-hamza/f09dca5bf54d6d25cbf2160c8d7ec7ff with some evaluation debug (that leads to the empty tree being evaluated).

jad-hamza:
Here's the debug output with information from the Inox evaluator as well: https://gist.github.com/jad-hamza/a87ef04d594ff6d578c2ce074e424d9c

romac:
This is really weird, especially since I can’t reproduce it. Can you please print the full Stainless and Inox programs as well? How about the original Counting example? Do your run into the same error there?

jad-hamza:
Can you try with --functions=theorem? Without it I cannot reproduce the bug. 
These are the options I'm using `--functions=theorem --vc-cache --debug=verification --check-models`. I'm also able to reproduce on the original version.

This is the full Stainless program: https://gist.github.com/jad-hamza/830e07aad73d0ca206d6295f7fe0c365

romac:
Alright, so it seems the culprit is indeed `--functions=theorem`, I get the same error with it.

It seems that, for some reason that I am currently unable to figure from reading the source, when this option is active `processMsg`'s body is replaced by an empty tree, as evidenced by [this line](https://gist.github.com/jad-hamza/830e07aad73d0ca206d6295f7fe0c365#file-fullprogram-scala-L174).

```scala
def processMsg(thiss: Behavior, msg: Msg, ctx: ActorContext): Behavior = <empty tree>[Behavior]
```

jad-hamza:
Thanks. I opened an issue https://github.com/epfl-lara/stainless/issues/112, because the behavior also appear in the master branch.

romac:
Copy-paste of @jad-hamza's comment on my fork:

> I had some pending changes, It should be OK now for total functions. I put some regression tests on hold for the moment. We should probably discuss with the others on how to implement support for lambdas with preconditions, we already discussed various options.

> To be safe we should also prevent extraction of requires inside lambdas for the moment, until the next pull request that adds higher order contracts.

jad-hamza:
I have made lots of changes in files that I'm not familiar with. Does that look good for total functions?

jad-hamza:
I'm leaving this as is for the moment (anyone welcome to continue one of the approaches we discussed :)). This branch should work ok if you don't use the require keyword for anonymous lambdas

mantognini:
Superseded by #160.

PR #115, opened 2017-11-20
Improved watch mode

mantognini:
This fix some issues with the reports displayed under `--watch` mode (basically, nested functions are no longer removed from the report when no noticeable changes were detected from the input source files and a new cycle was run without re-verifying anything).

This PR also adds some colors to the terminal output using ANSI colors. But it also trimmed down reports under `--watch` mode to show only unknown/invalid VC/TG.

Finally, it sorts VC/TG slightly better according to the function's id **and** the VC position.

### DIFF
--- a/core/src/main/scala/stainless/ast/SymbolOps.scala
+++ b/core/src/main/scala/stainless/ast/SymbolOps.scala
@@ -14,6 +14,17 @@ trait SymbolOps extends inox.ast.SymbolOps { self: TypeOps =>
   import trees.exprOps._
   import symbols._
 
+  /**
+   * Get the source of this function
+   *
+   * i.e. either its identifier or the identifier of its (recursively) outer function.
+   *
+   * NOTE no need to actually recurse here as [[Derived]] already
+   *      holds the requested data.
+   */
+  def source(fd: FunDef): Identifier =
+    fd.flags collectFirst { case Derived(id) => id } getOrElse fd.id
+
   override protected def createSimplifier(opts: inox.solvers.PurityOptions) =
     new SimplifierWithPC(opts) with transformers.SimplifierWithPC
 
--- a/core/src/main/scala/stainless/MainHelpers.scala
+++ b/core/src/main/scala/stainless/MainHelpers.scala
@@ -51,15 +51,6 @@ trait MainHelpers extends inox.MainHelpers {
   case object Verification extends Category
   case object Termination extends Category
 
-  object optJson extends inox.OptionDef[String] {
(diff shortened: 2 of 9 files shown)


mantognini:
To give you an idea, an example of `--watch=false`:

image

Same example but with `--watch=true` (and the patch epfl-lara/inox#51):

image

samarion:
Your first screenshot seems to reveal a bug in the top table border :) Looks like a cool feature otherwise!

mantognini:
> Your first screenshot seems to reveal a bug in the top table border :)

Yes, that's why the second show how epfl-lara/inox#51 is useful. :D

samarion:
> Yes, that's why the second show how epfl-lara/inox#51 is useful. :D

Ah my bad. LGTM then!

core/src/main/scala/stainless/ast/SymbolOps.scala
```diff
@@ -14,6 +14,17 @@ trait SymbolOps extends inox.ast.SymbolOps { self: TypeOps =>
...
 
+  /**
+   * Get the source of this function
+   *
+   * i.e. either its identifier or the identifier of its (recursively) outer function.
+   *
+   * NOTE no need to actually recurse here as [[Derived]] already
+   *      holds the requested data.
+   */
+  def source(fd: FunDef): Identifier =
```
samarion: Wouldn't this make sense in the `StainlessFunDef` implicit class in `ast/Definitions.scala`

mantognini: I guess, yes. Just a question: I've noticed a heavy use of `@inline`. Out of curiosity, why did you use this annotation?

samarion: No reason really, I was hoping for a tiny bit of performance gain when it's just an accessor. Actually I believe the JIT compiler does the inlining of small functions too so it may be useless.

mantognini: Okay. I was asking this because I was told in another context (C++) to leave this to the compiler heuristics. My guess is that using `final` should give stronger hints to the compiler. Anyway, this is not really important for this PR I believe.

mantognini: I've addressed this in `bunch_of_prs`: 5e2b3fc.

mantognini:
Merged in 5e2b3fc8257944f0b3dad10f3ca4112bff1daa29.

PR #116, opened 2017-11-20
command line options: update default values

mantognini:
I suggest to enable *by default* termination *and* verification components. Explicitly writing `--verification` or `--termination` will disable the other component unless both options are present in the command line arguments.

Also, because the VC cache seems stable I think it makes sense to have it on all the time, except when otherwise stated (i.e. using `--vc-cache=false`).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,7 @@ val isMac     = osInf.indexOf("Mac") >= 0
 val osName = if (isWindows) "win" else if (isMac) "mac" else "unix"
 val osArch = System.getProperty("sun.arch.data.model")
 
-val inoxVersion = "1.0.2-238-gfecc51d"
+val inoxVersion = "1.0.2-253-gb56078e"
 val dottyVersion = "0.1.1-bin-20170429-10a2ce6-NIGHTLY"
 
 lazy val nParallel = {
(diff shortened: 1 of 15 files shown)


samarion:
About the defaults: ok but tests __must not__ have `vc-cache=true`!

About verification and termination, I agree but I would go with a dedicated component for two reasons:
1. the default termination checker is quite slow because it tries many different approaches to termination, some of which only succeed very rarely. I think it would make sense to instantiate a restricted version of the termination checker with fewer components (fairly easy) that will be much faster for the default run. I would go with `recursionProcessor`, `selfCallsProcessor`, `decreasesProcessor`, and `lexicographicProcessor` (see `TerminationChecker.scala`).
2. this gives us the opportunity to use the dependency graph between termination and verification to actually output exact validity information about each function. I think we should output a report that provides a function-level VALID/INVALID status that looks at all the relevant VCs and termination checks.

mantognini:
> About the defaults: ok but tests must not have vc-cache=true!

Should be okay now.

> About verification and termination, I agree but I would go with a dedicated component for two reasons:

I removed this bit of the PR as it seems to deviate a lot from my initial intention.

samarion:
I think you still need to set the vc-cache option in the integration tests, no?

mantognini:
Good catch!

Note that this requires epfl-lara/inox#52 (in order to get a context that disables the cache) and can be further improved once #115 is merged (in order to properly test that no cache is used).

I've manually checked this by adding `???` to `def check` in `VerificationCache` in the meantime.

Issue #114, opened 2017-11-20
decreases & require order


mantognini:
The documentation on `decreases` (§ Introduction > Program Termination) is very thin and doesn't specify the order relation of this statement w.r.t. `require`. My guess is that `require` should come first as some assumptions in `decreases` could depends on the precondition, but currently stainless doesn't reject code such as this one:

```scala
  def requireDecreaseWrongOrder[T](l1: List[T], l2: List[T], x: T): Boolean = {
    decreases(l1.size)
    require(l1.content == l2.content)
    (l1-x).content == (l2-x).content
  }.holds
```

Currently stainless emits the following counterexample that doesn't pass `--check-models`.

```
[Warning ] - Model leads to runtime error: Requirement did not hold @19:5
[ Error  ] Something went wrong. The model should have been valid, yet we got this:
[ Error  ]   x: T        -> T#7
[ Error  ]   l1: List[T] -> Cons[T](T#6, Nil[T]())
[ Error  ]   l2: List[T] -> Nil[T]()
[ Error  ] for formula ¬{
[ Error  ]   assert(size[T](l1) >= 0, "Measure not guaranteed positive")
[ Error  ]   require(content[T](l1) == content[T](l2))
[ Error  ]   content[T](-[T](l1, x)) == content[T](-[T](l2, x))
[ Error  ] }
```

Swap the `require` and `decreases` clauses and everything is valid.

Proposed action to resolve this action:
 1. Impose order with `require` first (this can be generalised to reject programs that contain `require` in a position that is not the first statement of a function/lambda).
 2. Update the documentation.
 3. Add regression tests.

(Credit to @czipobence for spotting this.)

samarion:
I noticed this in the context of building new termination benchmarks and have fixed it (require goes first, indeed). I haven't updated the doc though, and I didn't add any errors during extraction to point out the users mistake.

I'm still fighting with parallelism so I have a bunch of pending commits... I'm getting closer but still can't get the regressions to pass on my machine with parallelism activated.

mantognini:
Autoclosed in #119, but doc & proper error message still required.

PR #119, opened 2017-11-23
Many fixes

samarion:
Fixes include:
- (partial) support for well-formedness check on free variables from @mantognini 
- refactored the logic in termination pipeline runner
- fixed and generalized handling of pre/post/measure specs during transformation pipeline
  (fixes https://github.com/epfl-lara/stainless/issues/114 but still needs docs) 
- more stable parallelism (still not done though)

### DIFF
--- a/core/src/main/scala/stainless/extraction/oo/Definitions.scala
+++ b/core/src/main/scala/stainless/extraction/oo/Definitions.scala
@@ -96,7 +96,11 @@ trait Definitions extends imperative.Trees { self: Trees =>
 
   type Symbols >: Null <: AbstractSymbols
 
-  trait AbstractSymbols extends super.AbstractSymbols with TypeOps { self0: Symbols =>
+  trait AbstractSymbols
+    extends super.AbstractSymbols
+       with TypeOps
+       with SymbolOps { self0: Symbols =>
+
     val classes: Map[Identifier, ClassDef]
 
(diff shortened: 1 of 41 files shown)


core/src/main/scala/stainless/extraction/oo/Definitions.scala
```diff
@@ -59,44 +59,72 @@ trait Definitions extends extraction.Trees { self: Trees =>
   }
 
   case class TypedClassDef(cd: ClassDef, tps: Seq[Type])(implicit val symbols: Symbols) extends Tree {
-    lazy val id: Identifier = cd.id
+    @inline def id: Identifier = cd.id
 
-    lazy val typeMap: Map[TypeParameter, Type] = (cd.typeArgs zip tps).toMap
+    private[this] var _typeMap: Map[TypeParameter, Type] = _
```
mantognini: I noticed you introduced `Lazy` in inox but you don't use it here. Any reasons for this choice?

samarion: Forgetfulness...

mantognini: That's a valid reason. ;-)

samarion:
Shall we go ahead and merge this and fix the remaining parallelism issues in another PR?

mantognini:
Sounds good to me.

Issue #126, opened 2017-11-26
Investigate slow termination checking


romac:
I am noticing random tests failing in the termination test suite. For example, the `NNFSimple` benchmark in [this build](http://laraquad4.epfl.ch:9000/epfl-lara/stainless/build/2059), which disappeared when I tried to reproduce it locally, but instead it is the `BinarySearch` test that failed.

mantognini:
With #132 (commit 0feb5107134e298b24ee38ea8065f8e8e52a3ec9) we disabled the NNFSimple test for termination. This commit should be reverted once the cause of termination checking slowness is identified and fixed.

samarion:
I believe this is a broader issue with parallelism during regressions. I don't know why it's happening, but the tests when run sequentially on my machine are individually much faster than when run on larabot (about a factor 3 I'd say). The full regression still takes much longer when run sequentially obviously but I don't see why the tests are so much slower, also when compared to when Stainless wasn't running in parallel.

samarion:
Seems to be mostly fixed by https://github.com/epfl-lara/stainless/commit/74ac842db46899312d19d8abc5a31d3e1b26d95b
We maybe still need to fix the timeout on decreases checks but I'll leave that for another issue if it turns out to be relevant.

PR #129, opened 2017-12-05
Catch `ExecutionException` within the Registry

romac:
This allows to properly shutdown Stainless, whereas it would otherwise just hang and needed to be manually killed.

### DIFF
--- a/core/src/main/scala/stainless/frontend/CallBackWithRegistry.scala
+++ b/core/src/main/scala/stainless/frontend/CallBackWithRegistry.scala
@@ -63,7 +63,10 @@ trait CallBackWithRegistry extends CallBack { self =>
     // Save cache now that we have our report
     saveCaches()
   } catch {
-    case e: ExecutionException => ()
+    case e: ExecutionException =>
+      stop()
+      tasks.clear()
+      endExtractions()
   }
 
   // See assumption/requirements in [[CallBack]]
--- a/core/src/main/scala/stainless/frontend/CallBackWithRegistry.scala
+++ b/core/src/main/scala/stainless/frontend/CallBackWithRegistry.scala
@@ -11,6 +11,7 @@ import scala.collection.mutable.{ ListBuffer, Map => MutableMap, Set => MutableS
 import io.circe.Json
 
 import java.io.File
+import java.util.concurrent.ExecutionException
 
 trait CallBackWithRegistry extends CallBack { self =>
   import context.{ options, reporter }
@@ -53,7 +54,7 @@ trait CallBackWithRegistry extends CallBack { self =>
   final override def stop(): Unit = tasks foreach { _.cancel(true) } // no need to update state, it's a KILL.
 
   // Build the report
-  final override def join(): Unit = {
+  final override def join(): Unit = try {
     val newReports = tasks map { _.get } // blocking! TODO is there a more efficient "get all" version?
     val reports = (report +: newReports) filter { _ != null }
     if (reports.nonEmpty) report = reports reduce { _ ~ _ }
@@ -61,6 +62,8 @@ trait CallBackWithRegistry extends CallBack { self =>
 
     // Save cache now that we have our report
     saveCaches()
+  } catch {
+    case e: ExecutionException => ()
   }
 
   // See assumption/requirements in [[CallBack]]
--- a/core/src/main/scala/stainless/frontend/CallBackWithRegistry.scala
+++ b/core/src/main/scala/stainless/frontend/CallBackWithRegistry.scala
@@ -1,4 +1,4 @@
-/* Copyright 2009-2016 EPFL, Lausanne */
+/* Copyright 2009-2017 EPFL, Lausanne */
 
 package stainless
 package frontend
@@ -63,10 +63,14 @@ trait CallBackWithRegistry extends CallBack { self =>
     // Save cache now that we have our report
     saveCaches()
   } catch {
-    case e: ExecutionException =>
+    case e: ExecutionException if isFatalError(e) =>
       stop()
       tasks.clear()
       endExtractions()
+  }
+  
+  private def isFatalError(e: ExecutionException): Boolean = {
+    e.getCause != null && e.getCause.isInstanceOf[inox.FatalError]
   }
 
   // See assumption/requirements in [[CallBack]]
--- a/core/src/main/scala/stainless/extraction/package.scala
+++ b/core/src/main/scala/stainless/extraction/package.scala
@@ -73,9 +73,11 @@ package object extraction {
       innerfuns.extractor  andThen
       inlining.extractor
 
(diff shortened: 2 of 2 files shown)


romac:
Some more changes are needed, as with the current `catch` block, nothing is executed again when a file changes under `--watch`.

romac:
Again, the CI fails because of a random termination exception:

```scala
[info] -  30: termination/valid/NNFSimple solver=smt-z3 *** FAILED *** (2 minutes, 51 seconds)
[info]   List(def eval(f: Formula): Boolean = f match {
[info]     case And(lhs, rhs) =>
[info]       eval(lhs) && eval(rhs)
[info]     case Or(lhs, rhs) =>
[info]       eval(lhs) || eval(rhs)
[info]     case Implies(lhs, rhs) =>
[info]       ¬eval(lhs) || eval(rhs)
[info]     case Not(f) =>
[info]       ¬eval(f)
[info]     case Literal(id) =>
[info]       evalLit(id)
[info]   }, @induct
[info]   def simplifySemantics(f: Formula): Boolean = {
[info]     eval(f) == eval(simplify(f))
[info]   } ensuring {
[info]     (holds: Boolean) => holds
[info]   }) was not empty Functions List(eval, simplifySemantics) should terminate (TerminationSuite.scala:26)
```

while on my machine with this very same commit, I get:

```
[Warning ] Parallelism is disabled.
[Warning ] warning: there were two feature warnings; re-run with -feature for details
[  Info  ] Result for isSimplified
[  Info  ]  => CLEARED (Relation Processor comparing sum of argument sizes)
[  Info  ] Result for simplify
[  Info  ]  => CLEARED (Relation Processor comparing sum of argument sizes)
[  Info  ] Result for simpleNNF
[  Info  ]  => CLEARED (Chain Processor)
[  Info  ] Result for isNNF
[  Info  ]  => CLEARED (Relation Processor comparing sum of argument sizes)
[  Info  ] Result for eval
[  Info  ]  => CLEARED (Relation Processor comparing sum of argument sizes)
[  Info  ] Result for vars
[  Info  ]  => CLEARED (Relation Processor comparing sum of argument sizes)
[  Info  ]   ┌─────────────────────┐
[  Info  ] ╔═╡ termination summary ╞══════════════════════════════════════════════════════════════════════════╗
[  Info  ] ║ └─────────────────────┘                                                                          ║
[  Info  ] ║ NEW  vars               ✓ Terminates (Relation Processor comparing sum of argument sizes)  0.638 ║
[  Info  ] ║ NEW  isNNF              ✓ Terminates (Relation Processor comparing sum of argument sizes)  0.467 ║
[  Info  ] ║ NEW  eval               ✓ Terminates (Relation Processor comparing sum of argument sizes)  4.155 ║
[  Info  ] ║ NEW  simplify           ✓ Terminates (Relation Processor comparing sum of argument sizes)  2.794 ║
[  Info  ] ║ NEW  simpleNNF          ✓ Terminates (Chain Processor)                                     3.033 ║
[  Info  ] ║ NEW  evalLit            ✓ Terminates (Non-recursive)                                       0.000 ║
[  Info  ] ║ NEW  simplifyIsStable   ✓ Terminates (Non-recursive)                                       0.000 ║
[  Info  ] ║ NEW  simplifySemantics  ✓ Terminates (Non-recursive)                                       0.000 ║
[  Info  ] ║ NEW  isSimplified       ✓ Terminates (Relation Processor comparing sum of argument sizes)  0.000 ║
[  Info  ] ╟┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄╢
[  Info  ] ║ total: 9      valid: 9    (0 from cache)   invalid: 0      unknown: 0     time:  11.087          ║
[  Info  ] ╚══════════════════════════════════════════════════════════════════════════════════════════════════╝
[  Info  ] Shutting down executor service.
```

I opened issue #126 to track this problem.

core/src/main/scala/stainless/frontend/CallBackWithRegistry.scala
```diff
@@ -53,14 +54,23 @@ trait CallBackWithRegistry extends CallBack { self =>
...
     // Save cache now that we have our report
     saveCaches()
+  } catch {
+    case e: ExecutionException if isFatalError(e) =>
+      stop()
+      tasks.clear()
+      endExtractions()
+  }
+  
+  private def isFatalError(e: ExecutionException): Boolean = {
```
mantognini: What made you add this filter?

core/src/main/scala/stainless/frontend/CallBackWithRegistry.scala
```diff
@@ -53,14 +54,23 @@ trait CallBackWithRegistry extends CallBack { self =>
...
     if (reports.nonEmpty) report = reports reduce { _ ~ _ }
     tasks.clear()
 
     // Save cache now that we have our report
     saveCaches()
+  } catch {
+    case e: ExecutionException if isFatalError(e) =>
+      stop()
+      tasks.clear()
+      endExtractions()
```
mantognini: `endExtractions()` is actually [always](https://github.com/epfl-lara/stainless/blob/master/core/src/main/scala/stainless/frontend/ThreadedFrontend.scala#L28-L35) run before reaching `join` or `stop`, as [documented](https://github.com/epfl-lara/stainless/blob/master/core/src/main/scala/stainless/frontend/CallBack.scala#L11-L21). ;-)

Or did you noticed it was required?

romac: Ah sorry, I have to admit I didn’t look at the docs.

core/src/main/scala/stainless/frontend/CallBackWithRegistry.scala
romac: Because if there is another type of exception (eg. StackOverflow or MatchError) I figured we’d like to have it bubble up to the top. FatalError are a special case of errors that are handled below and which are by that point will already have been reported in the console. Does that make sense?

mantognini: Yes, I think so. :)

core/src/main/scala/stainless/frontend/CallBackWithRegistry.scala
mantognini: I'll address this myself when merging.

romac:
Ok, thanks!

mantognini:
Merged in 68a4ff3461400debb9832328ceaf1f83b9795c2e.

PR #131, opened 2017-12-06
Extract `set.isEmpty` as `set == Set()`

romac:


samarion:
Please fix in the dotty extractor as well.

romac:
Done.

romac:
#126 kicked in again (and once again it is`NNFSimple` which fails to verify, haven't been able to reproduce it).

mantognini:
Merged in ddff4c982c7455ec5fc1af87cd4120ef8c046a02.

PR #117, opened 2017-11-21
Consider @unchecked functions as terminating

mantognini:
Following up on https://github.com/epfl-lara/stainless/pull/78#issuecomment-345967833. The patch seems to work. For example, for [ConstructorArgsBoxing3.scala](https://github.com/romac/stainless/blob/88c9016dd095613ffc605e65541ecf1589b76aad/frontends/benchmarks/extraction/ConstructorArgsBoxing3.scala):

```
[  Info  ]   ┌─────────────────────┐
[  Info  ] ╔═╡ termination summary ╞═══════════════════════════════════════════════════════════════════════╗
[  Info  ] ║ └─────────────────────┘                                                                       ║
[  Info  ] ║ NEW     b            ✓ Terminates (Non-recursive)                                 0.010       ║
[  Info  ] ║ NEW     c            ✓ Terminates (Non-recursive)                                 0.001       ║
[  Info  ] ║ NEW     list0        ✓ Terminates (Non-recursive)                                 0.001       ║
[  Info  ] ║ NEW     list1        ✓ Terminates (Non-recursive)                                 0.001       ║
[  Info  ] ╟┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄╢
[  Info  ] ║ total: 4      valid: 4    (0 from cache)   invalid: 0      unknown: 0     time:   0.013       ║
[  Info  ] ╚═══════════════════════════════════════════════════════════════════════════════════════════════╝
```

PR #132, opened 2017-12-07
Disable / move tests

mantognini:
- Workaround #126. This fixes nothing but only silences the timeout. Should be properly fixed in the future but for now it's blocking all (unrelated) PRs.
 - Move `Postcondition` test to `unchecked` because it no longer fits the supported fragment for quantifiers.


samarion:
LGTM, this will have to do for now

mantognini:
Good. :)

I plan to merge `bunch_of_prs` in one go to merge these (in order to avoid re-running the it test many many times):
 - #132 (this PR)
 - #129 
 - #78 
 - #110 
 - #115 
 - #117 
 - #131, assuming you have no additional concern on that one.

samarion:
Sounds good

PR #134, opened 2017-12-08
Fix method dependencies in Registry

### LINKED ISSUE #112: Empty tree when using --functions
(See https://github.com/epfl-lara/stainless/pull/111)

Before verifying the VCs, the program contains an empty tree for the definition `processMsg`.
This only happens when using the option --functions=theorem.

Full example at https://gist.github.com/jad-hamza/8e77b7e2dcb432f347e2842a86ceb0fb (not fully minimized sorry).

@mantognini Can this be related to the dependency checks?

I can investigate further to see where the empty tree appears.

mantognini:
Fixes #112.


mantognini:
aaaand I just noticed the tests failing...

mantognini:
Okay, all is green now.

samarion:
Could we maybe add a minified version of Jad's failing benchmark in the regression suite?

mantognini:
<details>
<summary>Minified version is something like this</summary>

```scala

import stainless.lang._

// Use with --functions=theorem
// Expect invalid result.
object TestFunctionsOption {

  def theorem(foo: Foo): Boolean = {
    foo.method == 42
  }.holds

  /* NON SEALED */ abstract class Foo { def method: BigInt }
  case class Bar() extends Foo { override def method: BigInt = 0 }

}
```
</details>



But the problem is actually testing this. The bug cannot be triggered from `VerificationSuite` directly because the extraction/registry needs the proper filter too. I've therefore added a specific regression suite. Let me know what you think.

samarion:
This is probably fine, but wouldn't it be simpler if you just used something like
```scala
 override def filter(ctx: inox.Context, name: String): FilterStatus = name match {
     case "extraction/registry/TestFunctionsOption.scala" => WithContext(ctx withOpts optFilterFunctions("theorem"))
    case _ => super.filter(ctx, name)
```
instead of handling the filter specifically?

I'm fine with the current version if this doesn't work / isn't ideal for some other reason.

mantognini:
I tried that but unfortunately it doesn't work as easily as that, unless we switch `DEBUG` to `true` [here](https://github.com/epfl-lara/stainless/blob/ef052866ceab36064d0804e534eb9c4f087b9668/frontends/scalac/src/it/scala/stainless/ComponentTestSuite.scala#L87) in order to use the right context. Alternatively, we could make allow subclass of `ComponentSuite` to modify this `DEBUG` flag (and rename it), but it wouldn't change much while making the tests even more complicated IMO.

Also, notice how the tests currently fail because some VC are unknown (I've added this new check).

samarion:
I'm not sure I understand why `DEBUG` would need to be true since I thought each component applies the filter by itself [here](https://github.com/epfl-lara/stainless/blob/ef052866ceab36064d0804e534eb9c4f087b9668/frontends/scalac/src/it/scala/stainless/ComponentTestSuite.scala#L108). But I agree that supporting `DEBUG` will make the tests too complicated.

Tests containing unknown VCs is expected behavior in the invalid regressions due to `check-models` being set to true. The reason for invalid models is that invalid assertions are assumed in VCs that have the assertion in their path condition. For example, in
```scala
assert(false)
assert(false)
```
the first assertion check will fail but the second one will succeed since `false ==> false`.
VCs with invalid assertions in their path can lead to invalid models or non-termination of the verification procedure. So we are only guaranteed to have one invalid VC check with a checked model for an invalid benchmark.

mantognini:
I see, thanks for the explanation about unknowns. I'll revert the change tomorrow.

Regarding the context, the problem is that the bug root cause is in Registry, which output is then feed to the component. The only way to properly test the bugfix is to share the same context from extraction to verification. This is not possible when DEBUG is off.

samarion:
Ah I see, thanks. I think you can put it in its own file then, it seems sufficiently different from the other verification tests.

mantognini:
All right, will do.

mantognini:
any additional comment before I merge this?

samarion:
I think the test might be a bit more stable/clear if you used the input utils to extract the minified program from a string and injected the "functions" option (and maybe also "check-models"?) within the testcase to make adding further registry verification tests possible. However, this is not really necessary currently and you can leave it as is and we'll fix it when the need arises if you prefer.

mantognini:
I think it was a good idea to use `InputUtils` directly. Now it's a regular unit test (and not part of `it:test` anymore).

samarion:
I think you should move `RegistryVerificationSuite.scala` into `frontends/common/src/test/scala/stainless/verification/`. Other than that, LGTM.

Issue #135, opened 2017-12-09
Missing some nodes in Registry: Passes$2, passes$0


rodamber:
The following code comes almost straight from the "Symbolic Input-Output Examples" subsection of the "Pure Scala" section of the documentation (the original code is missing the type parameter for List):

```scala
package test

import stainless._
import stainless.lang._


sealed abstract class List[T] {

  def size: Int = (this match {
    case Nil() => 0
    case Cons(h, t) => 1 + t.size
  }) ensuring { res => (this, res) passes {
    case Nil() => 0
    case Cons(_, Nil()) => 1
    case Cons(_, Cons(_, Nil())) => 2
  }}
}
case class Cons[T](h: T, t: List[T]) extends List[T]
case class Nil[T]() extends List[T]
```

Running `stainless Test.scala` outputs the following:

```
[Warning ] Parallelism is disabled.
[ Error  ] size$1 depends on missing dependencies: Passes$1, passes$0.
[Internal] Missing some nodes in Registry: Passes$1, passes$0
[Internal] Please inform the authors of Inox about this message
[Internal] inox.package$FatalError: Missing some nodes in Registry: Passes$1, passes$0
[Internal] Please inform the authors of Inox about this message
```

The version of stainless I'm using is the one from this commit: ddff4c982c7455ec5fc1af87cd4120ef8c046a02.

mantognini:
@samarion Do you plan to support this in stainless as well, or should it be removed from the documentation, library and extraction ([ExPasses](https://github.com/epfl-lara/stainless/blob/9bec2c50bcfd0b60587c30a0f0ff8cb1e74cb147/frontends/scalac/src/main/scala/stainless/frontends/scalac/ASTExtractors.scala#L371-L393))?

samarion:
I don't plan on supporting it (at least for now). We should remove the references I think and we can re-introduce them later if need be.

mantognini:
Addresses by #136.

PR #136, opened 2017-12-11
Cleanup: remove dead code and unsupported features from the library

mantognini:
Related to #135.

I've noticed some additional dead pieces of code. Commits can be reverted if/when adding support the those features.


samarion:
LGTM

PR #137, opened 2017-12-11
Fix processing of `Old`

mantognini:
Small patch to properly get rid of `old` in the `ImperativeCodeElimination` transformation.


samarion:
LGTM

Issue #98, opened 2017-11-09
Internal error with using ignore annotation


jad-hamza:
It's not a big deal but if `@ignore` is not meant for the user, we should probably change the error message. But then it's perhaps not easy to differentiate between the `@ignore`'s written by the user, and the legitimate ones from the library?

```scala
import stainless.annotation._

object SI_Ignore {
    @ignore
    def f(x: Int) = 0

    def g(x: Int) = f(x)
}
```

mantognini:
It seems `f` is not extracted at all:

```
...
[ Debug  ] Got a unit for /localhome/mantogni/stainless/frontends/scalac/target/scala-2.11/classes/stainless/math/package.scala: math with:
[ Debug  ]      functions -> [abs, max, max, min, min]
[ Debug  ]      classes   -> []
[ Debug  ] Got a unit for /localhome/mantogni/SubProjects/Proofs/Tests/SI_Ignore.scala: SI_Ignore with:
[ Debug  ]      functions -> [g]
[ Debug  ]      classes   -> []
[ Error  ] g$1 depends on missing dependencies: f$26.
[Internal] Missing some nodes in Registry: f$26
[Internal] Please inform the authors of Inox about this message
[ Error  ] Rethrowing exception emitted from within the compiler: Missing some nodes in Registry: f$26
```

This annotation is currently defined as

 > Ignore this definition when extracting Stainless trees. This annotation is useful to define functions that are not in Stainless’s language but will be hard-coded into specialized trees, or to include code written in full Scala which is not verifiable by Stainless.

I don't remember exactly what was done in Leon, but I guess we should be extracting something with some kind of `choose` body instead of the actual function body, right?

samarion:
I believe that not extracting it at all is correct. The ignore annotation is used for functions that will be extracted as something else than a function. For example, the ignored `forall` function calls gets extracted as `Forall` ASTs and the `forall` function shouldn't exist within the program.

In order to emit a warning, we would need to compile a set of all symbols that should indeed be ignored (we don't have this for now). Can we maybe make the annotation private within the stainless library?

mantognini:
Looking at the sphinx doc, it seems not all annotations are documented. I thought `@extern` had been removed but it isn't the case. What's the usage of `@internal`? Maybe this one should be used instead of `@ignore` if the usage is meant to be private.

I don't know if an annotation can be made private or not, but we can check whether a symbols is from the library or not:

```diff
[...]

PR #139, opened 2017-12-11
Addressing some documentation issues with annotations

mantognini:
Related to the discussion in #98.

I used `private[stainless]` to disallow `@ignore` in user code as it seems `@extern` should be used instead. The remaining questions are, should we get rid of `@traceInduct`, `@internal` and `@pure` since they don't seem well tested at best? Or how do you see things, @samarion ?


mantognini:
An additional question: synthesis being not yet implemented in stainless, should I remove this part of the documentation & library as well for now? (No tests rely on these features ATM.)

samarion:
Yes, I believe that removing the synthesis stuff makes sense. We can also remove (or comment?) the `@traceInduct`, `@internal` and `@pure` annotations though I think `@traceInduct` at least should make its way back into Stainless at some point.

mantognini:
Okay, so I've removed the synthesis part (`holes` trees and transformers still remains though) and I also removed the three annotations as they weren't handled anyway. Those separate commits can be reverted in the future if needed.

PR #144, opened 2017-12-15
--eval: evaluator component for parameterless functions

mantognini:
This PR adds support for the `--eval` command line option.

As with the other component, you can use `--watch` or `--functions`. And because this is using the evaluator facilities, you can use `--ignore-contracts` and the other related options. You can combine it with `--verification` and/or `--termination` as well.

Just to showcase the feature, here is an example of report for [this file](https://gist.github.com/mantognini/ccbee3c64f929ddb182907fbb7ed8e3e#file-eval-scala):

```
$ stainless Eval.scala --eval --ignore-contracts
[  Info  ]   ┌──────────────┐
[  Info  ] ╔═╡ eval summary ╞═════════════════════════════════════════════════════════════════════════╗
[  Info  ] ║ └──────────────┘                                                                         ║
[  Info  ] ║ bar        "hello, stainless!"              valid        Eval.scala:11:3         0.009   ║
[  Info  ] ║ foo        43                               invalid      Eval.scala:4:3          0.000   ║
[  Info  ] ║ foobar     "failure in foo's PC"            valid        Eval.scala:13:3         0.000   ║
[  Info  ] ║ goo        1                                             Eval.scala:9:3          0.000   ║
[  Info  ] ║ loop       Stack overflow                   crashed      Eval.scala:15:3         0.071   ║
[  Info  ] ╟┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄╢
[  Info  ] ║ total: 5    valid: 3    (0 from cache) invalid: 2    unknown: 0    time:   0.080         ║
[  Info  ] ╚══════════════════════════════════════════════════════════════════════════════════════════╝
```

As you can see, two main informations are attached to the evaluated functions: the value of their body and the status of their postcondition (if any). And of course, you've got colours in the actual output as per 
#115.


core/src/main/scala/stainless/ast/Definitions.scala
```diff
@@ -68,6 +68,9 @@ trait Definitions extends inox.ast.Definitions { self: Trees =>
       Lambda(Seq(ValDef(FreshIdentifier("res", true), fd.returnType)), BooleanLiteral(true))
     }
 
+    /** Check whether the function has no (generic) parameter. */
+    final def isParameterless = fd.params.isEmpty && fd.tparams.isEmpty
```
samarion: Is it necessary to check that `fd.tparams.isEmpty`? I assume the check is only rarely needed in practice anyway, but it would seem to me that type parameters aren't a problem for the evaluator.

mantognini: I wasn't sure of all the implications it had. I guess it could be removed in the future if people really want to run generic parameterless function.

mantognini:
>  I would assume that precondition failures are more common here. 

Luckily we are well covered here. :)

Say you have `require(false)` in `foo`:
```
[  Info  ] ║ foo     Requirement did not hold @6:5  crashed                        Eval.scala:4:3   0.000 ║
```

samarion:
But won't you get the same result for postconditions without special support?

mantognini:
Yes, you actually do. But then if the PC fails you don't get the body's value.

Issue #146, opened 2017-12-15
Eval table empty with watch


vkuncak:
Using command line  --eval --watch  
results in empty set of evals, which remains empty even if I change the bodies of parameterless functions. See attached file.
Without --watch, the eval values are shown.
[WatchNoEval.scala.txt](https://github.com/epfl-lara/stainless/files/1562950/WatchNoEval.scala.txt)

mantognini:
I think this is due to the changes introduced in https://github.com/epfl-lara/stainless/pull/115#issuecomment-345749392. Maybe the compact mode should be enabled through an command line option.

PR #147, opened 2017-12-15
Fix reporting under --watch 

### LINKED ISSUE #125: Updating --watch mode report
```scala
object WatchMode {
  def watched() = {
    assert(false) // the assertion stays in the report even after removing that line
  }
}
```

I tried with https://github.com/epfl-lara/stainless/pull/115 as well, but the same behavior happens (the watch mode keeps reporting that the assertion is invalid, even after it has been commented out).

mantognini:
Closes #125.

In a nutshell, now reports keep track of which functions were verified/evaluated/... to generate them. This way, we can remove old VC that no longer exists (and are not replaced by another one).

### DIFF
--- a/core/src/main/scala/stainless/Report.scala
+++ b/core/src/main/scala/stainless/Report.scala
@@ -5,7 +5,10 @@ package stainless
 import inox.utils.Position
 import inox.utils.ASCIIHelpers._
 
-import io.circe.Json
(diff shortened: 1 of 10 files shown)


vkuncak:
Merging this to clarify if it fixes the empty eval table issue #146

PR #153, opened 2017-12-19
Reject bad preconditions

### LINKED ISSUE #114: decreases & require order

mantognini:
Closes #114.

 1. Impose order with `require` followed by `decreases` (documentation is updated accordingly).
 2. Reject functions with multiple `require` or `decreases`, or any pre-condition in an unexpected position.
 3. Add regression tests.
 4. Bonus: reduced copy-paste in integration suite.

Note: the `def traverse` is quite general and should probably be moved to inox in the future if some other modules would benefit from it.


frontends/common/src/it/scala/stainless/ExtractionSuite.scala
```diff
@@ -61,23 +63,32 @@ class ExtractionSuite extends FunSpec with inox.ResourceUtils with InputUtils {
...
+  // the TreeSanitizer.
   def testRejectAll(dir: String): Unit = {
     val (reporter, ctx, files) = testSetUp(dir)
 
     describe(s"Programs extraction in $dir") {
-      val tryPrograms = files map { f => f -> Try(loadFiles(ctx, List(f))._2) }
+      val tryPrograms = files map { f =>
+        f -> Try {
+          val program = loadFiles(ctx, List(f))._2
+          extraction.TreeSanitizer.check(program)
```
samarion: Shouldn't we actually run the full extraction pipeline here?

mantognini: Makes sense. The latest commit addresses this.

PR #160, opened 2017-12-22
WIP: Make lambdas total by default

mantognini:
Supersede #111.

I've updated the frontends & library so that the user can write their partial lambda in a concise fashion, avoiding code duplication and therefore potential bugs in the user programs.

Please have a look at the last commit regarding `ImperativeCodeElimination`. I'm not exactly sure how things should work with `assume` versus `require` here.

Also, I notice now that the tests are failing because `assume` is not supported by `CodeGeneration`. Maybe it can be simply replaced by `require` in `PartialFunctionSugar`?

Note: one could go a bit further and add support for `PartialFunction(f)` where `f` is a regular function and decide to extract its precondition. To do this, however, one would need the symbol for `f` and therefore this should probably happen after the frontend (scalac/dotc). The `UpDown.scala` benchmark would benefit slightly from this in terms of code duplication.


frontends/library/stainless/lang/PartialFunction.scala
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * is converted into
+   *
+   *    ~>(
+   *      { (x: A) => pre(x) },
+   *      { (x: A) => assume(pre(x)); body(x) }
+   *    )
+   */
+  @ignore
+  def apply[A,B](f: A => B): A ~> B = ~>(_ => true, f)
```
samarion: Maybe implement the `pre` with a try?

core/src/main/scala/stainless/extraction/imperative/ImperativeCodeElimination.scala
```diff
@@ -347,6 +347,7 @@ trait ImperativeCodeElimination extends inox.ast.SymbolTransformer {
         case ld @ Lambda(params, body) =>
           val (optPre, lBody) = body match {
             case Require(pred, body) => (Some(pred), body)
```
samarion: Shouldn't we now disallow `require` in lambdas since they are translated to assumes during extraction?

core/src/main/scala/stainless/extraction/imperative/ImperativeCodeElimination.scala
```diff
@@ -347,6 +347,7 @@ trait ImperativeCodeElimination extends inox.ast.SymbolTransformer {
...
+            case Assume(pred, body) => (Some(pred), body) // FIXME This transforms assume into require. Is this okay?
```
samarion: We probably don't want it to become a require. You can change the line
https://github.com/epfl-lara/stainless/blob/a29b39de57f2bd967f08bff5f59826b65ebf8365/core/src/main/scala/stainless/extraction/imperative/ImperativeCodeElimination.scala#L358
to not use `withPrecondition` but inject an assume instead (if `newPre` is defined).

frontends/library/stainless/lang/PartialFunction.scala
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Describe a partial function with precondition [[pre]].
+ *
+ * Do not attempt to create it using the ~>'s companion object's apply method.
+ * Instead, use [[PartialFunction.apply]] defined beflow. (Private constructor
+ * cannot be expressed in Stainless.)
+ */
+@library
+case class ~>[A,B] /*private*/ (pre: A => Boolean, f: A => B) {
```
samarion: I think `f` should be private, especially since you assume the pre in its body.
It's too bad we don't support private constructors. We should maybe add an extra extraction rule in the frontends that rejects manual construction of `~>`.

frontends/library/stainless/lang/PartialFunction.scala
```diff
@@ -0,0 +1,51 @@
...
+  def apply(a: A): B = {
+    require(pre(a))
+    f(a)
+  }
+}
+
+@library
+case class ~>>[A,B](pre: A => Boolean, private val f: A => B, post: B => Boolean) {
```
samarion: I wonder whether we should define this as
```scala
case class ~>>[A,B](private val f: A ~> B, post: B => Boolean) {
  require(forall((x: A) => f.pre(x) ==> post(f(x))))
  val pre = f.pre
  def apply(a: A): B = {
    require(pre(a))
    f(a)
  } ensuring (post)
}
```
To avoid requiring more dedicated support in the frontend.
Also a cool constructor could be like
```scala
implicit class PartialFunctionWrapper[A,B](val f: A ~> B) {
  def withPost(post: B => Boolean): A ~>> B = {
    require(forall((a: A) => f.pre(a) ==> post(f(a))))
    ~>>(f, post)
  }
}
```

core/src/main/scala/stainless/codegen/CodeGeneration.scala
```diff
@@ -441,11 +441,9 @@ trait CodeGeneration { self: CompilationUnit =>
...
             .withTypeParameters(tps)
 
-          val preLambda = if (pre) {
-            lambda.copy(body = BooleanLiteral(true))
-          } else {
-            lambda.copy(body = weakestPrecondition(lambda.body))
-          }
+          // FIXME: should we remove this, since lambdas do not have
+          // preconditions anymore?
+          val preLambda = lambda.copy(body = BooleanLiteral(true))
```
samarion: Yes, lambdas should not have pres anymore. You also need to change the `Lambda.scala` class in the `runtime` package and remove the `pre` argument to `mkLambda`.

Also I guess you noticed that codegen doesn't know about the `Assume` tree. You can handle it the same way as `Assert`.

samarion:
Ok so this is now almost ready.
I still want to change the API of `~>>` a little and add support for the `PartialFunction(f)` pattern @mantognini mentioned but that should (hopefully) not take me much longer.

PR #161, opened 2017-12-22
Add holds on `check`

mantognini:
Following a quick discussion with @jad-hamza & @romac from yesterday.

The idea is that, hopefully, the solver can benefit from this change when `check { ... }` expression are combined (e.g. through conjunctions): when verifying the n-th check expression, the solver shouldn't have to re-verify the n-1 previous ones.

I however don't have concrete proof that this actually makes things faster.


romac:
I don't think that this really does what we thought it would, right @samarion?

samarion:
I'm not entirely sure what this will do, but I have some pending changes in Inox which might make this relevant once I've merged them. I'm changing the `assume-checked` simplifier to keep functions with postconditions around so this could become a stronger form of `assert`.

PR #163, opened 2018-01-05
Type alias support

Phenitei:
Add support for type aliases and type lamdas in code extraction for the scalac and dotty frontends.


samarion:
Merged! Sorry for the long wait...

PR #164, opened 2018-01-11
Update call to deprecated java constructors.

Phenitei:
Starting with JDK9, building values of Byte, Short, Integer, Long,
Boolean and Character through their constructors and not the `valueOf`
method is deprecated.

The `valueOf` methods however exists since at least version 1.5, so this is not a breaking change for anyone (I hope).


samarion:
Merged, thanks!

PR #175, opened 2018-02-28
Use native packager to generate runner scripts.

dragos:
This Sbt plugin is the standard way to make executable projects in Scala
and gives a few more options, like the way to pass JVM arguments or system
properties, in addition to working on all platforms, including Windows.

The `universal:packageBin` task can produce an archive that can be used to
distribute the jars and runner scripts on all platforms.

PS. I did not remove the old scripts that are generated on `compile`, but that should be eventually the case.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -35,8 +35,6 @@ lazy val extraClasspath = taskKey[String]("Classpath extensions passed directly 
 
 lazy val scriptPath = taskKey[String]("Classpath used in the stainless Bash script")
 
-lazy val script = taskKey[Unit]("Generate the stainless Bash script")
-
 lazy val baseSettings: Seq[Setting[_]] = Seq(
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -210,6 +210,7 @@ lazy val `stainless-core` = (project in file("core"))
   //.dependsOn(inox % "compile->compile;test->test")
 
 lazy val `stainless-scalac` = (project in file("frontends/scalac"))
+  .enablePlugins(JavaAppPackaging)
```
samarion: Can this be enabled for the `stainless-dotty` project as well?

samarion:
If this subsumes the scripts, then let's just remove them and update the docs.

dragos:
Just a reminder: the docs need to be regenerated and published, so people find the new way to link to runner scripts.

PR #176, opened 2018-02-28
sbt-stainless plugin

dotta:
This PR adds a sbt-stainless plugin. Please, read the commits message for details.

I expect this PR will fail to execute the scripted test because currently the sbt-stainless plugin requires an external solver to be installed (and I assume that will be missing on the machine running the PR validation). To fix this issue I need a repository to fetch the scalaz3 dependency.

### DIFF
--- a/.larabot.conf
+++ b/.larabot.conf
@@ -1,6 +1,7 @@
 commands = [
     "sbt -batch -Dparallel=10 test"
     "sbt -batch -Dparallel=10 it:test"
+    "sbt -batch scripted"
 ]
 
 trusted = [
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,4 @@
+import sbt.ScriptedPlugin
 
 val osInf = Option(System.getProperty("os.name")).getOrElse("")
 
@@ -24,6 +25,8 @@ lazy val nParallel = {
   }
 }
 
+val SupportedScalaVersions = Seq("2.11.8")
+
 lazy val frontendClass = settingKey[String]("The name of the compiler wrapper used to extract stainless trees")
 
 // FIXME @nv: dotty compiler needs the scala-library and dotty-library (and maybe some other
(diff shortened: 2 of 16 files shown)


samarion:
This looks good to me. If think if you rebase and force-push, the CI should hopefully start running on your PRs.

I'm not sure I understand about the scalaz3 dependency. Will we have one fat jar per os architecture or is there some way to inject the jar into the classpath later on? And if this is possible, why aren't we doing it for Inox?

dotta:
> This looks good to me. If think if you rebase and force-push, the CI should hopefully start running on your PRs.

Done :)

> I'm not sure I understand about the scalaz3 dependency. Will we have one fat jar per os architecture or is there some way to inject the jar into the classpath later on? And if this is possible, why aren't we doing it for Inox?

You are right, we need a stainless-scalac-plugin fat jar per os that includes the z3 dependency. I'll try to work this out now.

dotta:
@samarion To be able to test that scalaz3 is correctly included in the stainless compiler plugin, I need to be able to run stainless on Mac with it. As this is currently not possible (because of https://github.com/epfl-lara/ScalaZ3/issues/56) I suggest we review and merge the PR as is and we create a ticket for including scalaz3 in the stainles-scalac-plugin fat jar (I've actually pushed a commit with the functionality here https://github.com/dotta/stainless/commit/d7748e7ce039449807ffcce434b025367dc7775b, but I can't test it :)). Sounds good?

sbt-plugin/src/sbt-test/sbt-plugin/simple/success/src/main/scala/ConstructorArgsBoxing1.scala
```diff
@@ -0,0 +1,12 @@
+
+import stainless.lang._
+import stainless.collection._
+
+object ConstructorArgsBoxing1 {
```
samarion: You may want to change the successful file here as @dragos robustness PR will break this test case.
You can use `frontends/benchmarks/verification/valid/ADTInvariants1.scala` instead (for example).

samarion: Oups, I forgot about this comment when hitting approve. Please change this file and then I'll merge.

dotta: Done and force pushed.

dotta:
> I'll trust you on the sbt magic and the CI seems happy so I'm good to go.

Yeah, the sbt-stainless plugin might not yet be bulletproof, but I think it's a good start and it definitely works for the simple case I tested. Let's now have it published so that you/Viktor can test it out.

Another aspect we should discuss is if you want the plugin to work with dotty as well (I'm not sure how much work that would be, but hopefully it's not much).

> You should be able to test the fat jar using the CI system also in the mean time. You can change .larabot.conf to only run the scripted test if you need a faster turnaround.

I'm afraid I wouldn't be very productive with this working mode. I think it'd be good to hear from @vkuncak if having scalaz3 working on Mac OS is something he'd want us to do. If so, this can be something we tackle next week.

PR #177, opened 2018-02-28
Improvement/robustness check phase

dragos:
Work in progress, do not merge yet. 

We split the translation in two phases:

- a validation phase that rejects invalid Stainless programs
- a translation step that produces stainless trees

This has several advantages:

- the validation step can report as many errors as possible, without running the risk of cascading errors
- the translation phase can assume correct input and becomes simpler
- failure in the translation phase is a bug (otherwise, a rule should be added to the validation phase)

Still to do:

- [x] add tests for each rule
- [ ] port all validation from translation to the validation phase
- [ ] remove validation code from the translation phase
- [x] make sure existing tests all pass/fail as before


frontends/scalac/src/main/scala/stainless/frontends/scalac/FragmentChecker.scala
```diff
@@ -0,0 +1,109 @@
...
+      tree match {
+        case ClassDef(mods, name, tparams, impl) if sym.isAbstractClass =>
+          val firstParent = sym.info.firstParent
+          if (firstParent != definitions.AnyRefTpe) {
+            // we assume type-checked Scala code, so even though usually type arguments are not the same as
+            // type parameters, we can assume the super type is fully applied (otherwise we could check via
+            // firstParent.typeSymbol.typeParams)
+            val parentTParams = firstParent.typeArgs
+            val tparams = sym.info.typeParams
+            if (tparams.size != parentTParams.size)
```
dragos: I wonder: should this be `tparams.size >= parentTParams.size`? In other words, we allow new type parameters, but we can't "remove" them?

samarion: As long as we don't support variance, we need an exact bijection between parent and child type parameters. They should even be in the same order (we assume this a lot during later transformations).

dragos: So there can't be a case like this?
```
abstract class Foo
abstract class Bar[T] extends Foo
```
If we replace `Foo` with `AnyRef` it's valid, so I thought it should be valid for user-defined classes as well.

samarion: We don't yet support `AnyRef`. There is no top or bottom in our type lattice and (almost) no subtyping.
Adding support for type parameter variance has basically the same issues as supporting `Any` and `Nothing`, so it will all be coming at the same time.

dragos: All right, but for extraction purposes (and validation), `AnyRef` is present as the default superclass of any class, so I should let that kind of inheritance pass. I am fine to let it pass for `AnyRef` and nothing else for the moment.

samarion: Sounds good.

dragos:
@dotta scripted tests failed, I think the classpath may be incomplete:

```
[info] [error] (success/compile:compileIncremental) scala.reflect.internal.MissingRequirementError: class stainless.annotation.extern in compiler mirror not found.
[info] [error] Total time: 3 s, completed Feb 28, 2018 5:43:31 PM
[error] x sbt-plugin / simple
[error]    {line 2}  Command failed: success/compile failed
java.lang.RuntimeException: Failed tests:
	sbt-plugin / simple

	at scala.sys.package$.error(package.scala:27)
	at sbt.test.ScriptedRunner.runAll(ScriptedTests.scala:156)
	at sbt.test.ScriptedRunner.run(ScriptedTests.scala:151)
	at sbt.test.ScriptedRunner.run(ScriptedTests.scala:125)
```

dragos:
@samarion I decided to make the rest of the chances in a separate PR, so please have a look at one. If everything is ok I'll start changing the logic in the extraction phase. That part might be more tricky as it touches more on the existing logic.

frontends/scalac/src/main/scala/stainless/frontends/scalac/FragmentChecker.scala
```diff
@@ -0,0 +1,123 @@
...
+
+  // defined by StainlessExtraction
+  val ctx: inox.Context
+
+
+  class Checker extends Traverser {
+    val ExternAnnotation = rootMirror.getRequiredClass("stainless.annotation.extern")
+    val IgnoreAnnotation = rootMirror.getRequiredClass("stainless.annotation.ignore")
+
+    private val stainlessType: Map[Symbol, String] = Map(
```
samarion: This looks great, could we maybe also add `scala.Option` which is pretty common?
Also, we could fix https://github.com/epfl-lara/stainless/issues/103 using a similar mechanism for common functions/method (like `println`).

samarion:
About the classpath, that was indeed not the best validation benchmark since it doesn't rely on library. We should add some more complete test like `frontends/benchmarks/verification/valid/Monads3.scala` (monad laws on `stainless.collection` lists).

dragos:
@samarion Yes, mostly the checks that are done now during extraction, since those are more risky (touching on actual logic :D)

I'll fix the Dotty failure, of course, and add Option/Either to the list of warnings

dotta:
> scripted tests failed, I think the classpath may be incomplete:

I'm very much puzzled, the PR validator was green on my PR (https://github.com/epfl-lara/stainless/pull/176)...

samarion:
@dotta Yes, but the valid sbt-test we used doesn't depend on the Stainless library so it wouldn't have shown the issue (no need to have the library in the classpath). @dragos is loading the `extern` and `ignore` classes from classpath in his input validation checker which is why we're now seeing the issue.

dragos:
So, besides the Sbt scripted test, there's a failure:

```
[K] Killing 14554 ...
[K] > kill -9 14554 ...
died a natural death
```

Of course, killing can't be ruled as natural death, but I wonder why it failed, since tests seemed to have passed so far.

samarion:
Parallelism goes over the top currently so if two PRs are being validated at the same time they may run into the memory limit and get killed. It's on my TODO list to find a good way to limit max concurrency in Stainless.

I restarted the build so we'll see if the issue persists, though I see @dotta launched another one so it might appear again.

dragos:
@samarion can you please restart the build? Might be a spurious failure

dragos:
Finally, it's green! can we please merge this? :)

PR #186, opened 2018-03-01
Simplify extraction

dragos:
This PR is the next step after #177 is merged. It removes some checks performed during code extraction and moves them to the fragment checker.


dragos:
Looks like a spurious failure again, the code is rebased on the latest master and tests should pass.

PR #193, opened 2018-03-19
Add a partial evaluator for PureScala programs

romac:
**Note:** This is a work-in-progress, as the test suite currently fails in quite a few places.

The partial evaluator can be triggered in two ways:

- By adding a `@partialEval` to a function, whose body is then going to be replaced by its partially evaluated version.
- By setting the `--partial-eval` flag when invoking Stainless, which instructs it to partially evaluate verification conditions before they are submitted to Inox.

A `partial-eval` debug section is also included, which can be enabled with the `--debug=partial-eval` flag.


romac:
I am still getting `AbstractMethodErrors` either in local, or on the build server. @samarion Any idea how to fix this?

romac:
Merged in the `parallel-pipeline` and `romac/parallel-pipeline` branches.

PR #214, opened 2018-03-30
Create one post-condition per path in the function and use assertions to prove post-conditions

jad-hamza:
This PR adds two features:
1) Multiple post-condition VCs are created for one function, one per path of the function. This is done thanks to the `ensuring` function, which perhaps needs to be extended to other expressions of the language.
2) Assertions that are along a path are kept when generating VCs for post-conditions, thanks to the call to `CollectorWithPC`.


samarion:
Merged, thanks!

PR #224, opened 2018-04-06
Infer precondition of partial functions yielded by inexhaustive pattern matches.

romac:
eg.

```scala
def test1 = PartialFunction[Option[String], Int] {
  case Some(str) => 42
}
```

gets precondition:

```scala
x$ match {
  case Some(str) => true
  case _ => false
}
```


romac:
Closing because GitHub does not seem to pick my last forced push.

PR #226, opened 2018-04-09
Turn body of `holds` expressions into postcondition

romac:
As per this commit, the following program

```scala
def foo(x: Int): Boolean = {
  require(x >= 0)
  x == 0 || x > 0
}.holds

def bar(x: Int): Boolean = {
  require(x >= 0)
  x == 0 || x > 0
}.holds because { x != -1 }

def baz(x: Int): Boolean = {
  require(x >= 0)
  x == 0 || x > 0
}.holds(x != -1)
```

gets extracted into

```scala
def foo(x: Int): Boolean = {
  require(x >= 0)
  true
} ensuring { _ => x == 0 || x > 0 }

def bar(x: Int): Boolean = {
  require(x >= 0)
  x != 1
} ensuring { proof => proof && (x == 0 || x > 0) }

def baz(x: Int): Boolean = {
  require(x >= 0)
  x != 1
} ensuring { proof => proof && (x == 0 || x > 0) }
```

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
@@ -117,20 +117,18 @@ trait AntiAliasing extends inox.ast.SymbolTransformer with EffectsChecking { sel
         case _ => ()
       }
 
-      val newFd = fd.copy(returnType = effects.getReturnType(fd))
-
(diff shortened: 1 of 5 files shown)


romac:
Blocked by #277

romac:
Blocked by #227

romac:
This PR is rebased on top of #228, and it is thus suggested to either merge #228 first, or to just close #228 and only merge this one.

PR #228, opened 2018-04-10
Eliminate local effects in postconditions

### LINKED ISSUE #227: Local effects are not properly eliminated in postconditions
```scala
import stainless.lang._

object bug {
  case class Foo(var bar: Int)

  def mutateFoo(foo: Foo): Boolean = {
    foo.bar = foo.bar + 1
    foo.bar > 10
  }

  def test = {
    true
  } ensuring { _ =>
    val foo = Foo(0)
    val after = mutateFoo(foo)
    after == false
  }
}
```

yields 

```scala
[info] [Internal] Type error: {
[info] [Internal]   true
[info] [Internal] } ensuring {
[info] [Internal]   (x$1: Boolean) => {
[info] [Internal]     val foo: Foo = Foo(0)
[info] [Internal]     val after: Boolean = mutateFoo(foo)
[info] [Internal]     after == false
[info] [Internal]   }
[info] [Internal] }, expected Boolean, found <untyped>
```

because the call to `mutateFoo` is not rewritten properly.

romac:
Close #227

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
@@ -117,20 +117,18 @@ trait AntiAliasing extends inox.ast.SymbolTransformer with EffectsChecking { sel
         case _ => ()
       }
 
-      val newFd = fd.copy(returnType = effects.getReturnType(fd))
-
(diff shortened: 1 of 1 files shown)


romac:
The build somehow got stuck, while it went through in #226 which is rebased on this PR. Closing in favour of #226.

PR #236, opened 2018-04-16
Check both constructor and sort invariant on ADT creation

romac:


romac:
@samarion Ready to be merged

frontends/scalac/src/it/scala/stainless/verification/VerificationSuite.scala
```diff
@@ -29,6 +29,9 @@ trait VerificationSuite extends ComponentTestSuite {
     case "verification/valid/IntSet" => WithContext(ctx.withOpts(inox.solvers.optAssumeChecked(false)))
     case "verification/valid/IntSetInv" => WithContext(ctx.withOpts(inox.solvers.optAssumeChecked(false)))
     case "verification/valid/IntSetUnit" => WithContext(ctx.withOpts(inox.solvers.optAssumeChecked(false)))
+    // Require check-models to be turned off
+    case "verification/invalid/ParentInvariant" => WithContext(ctx.withOpts(inox.solvers.optCheckModels(false)))
```
samarion: Is this still necessary now that we can check models with ADT invariants?

romac: No indeed, just removed it.

PR #240, opened 2018-04-16
Replace reference to case object with a call to a constructor function

### LINKED ISSUE #239: References to case object fields or methods are not properly extracted
The following program
```scala
case object Foo {
  def bar: BigInt = 0
}

def test = Foo.bar
```

is incorrectly extracted as

```scala
case object Foo {
  def bar: BigInt = 0
}

def test = this.bar
```

romac:
This enables us to check the ADT invariant for case objects even if no one ever directly references the case object. Previously, this was not the case and could thus potentially lead to runtime errors.

Because the "constructor function" is marked with `@inline`, it will disappear in the resulting program.

This PR also fixes #239.


romac:
This PR needs to be merge after #242.

PR #242, opened 2018-04-19
Check model for invalid ADT invariants in a more careful way

### LINKED ISSUE #237: Valid counter-examples to ADT invariant reported as unknown with `--check-models`
The following testcase correctly reports an invalid ADT invariant for `fail` without `--check-models`, but when the option is set, it is reported as unknown because the model obviously leads to a runtime error, ie. `Model leads to runtime error: ADT invariant violation for Foo(0)`.

```scala
object test {
  case class Foo(x: BigInt) {
    require(x != 0)
  }

  def fail = Foo(0)
}
```

romac:
Previously, the following testcase would correctly report an
invalid ADT invariant for `fail` without `--check-models`,
but when setting the latter option, it would be reported as
unknown because the model would to a runtime error,
ie. `Model leads to runtime error: ADT invariant violation for Foo(0)`.

```scala
object test {
  case class Foo(x: BigInt) {
    require(x != 0)
  }

  def fail = Foo(0)
}
```

Fixes #237

### DIFF
--- a/core/src/main/scala/stainless/verification/DefaultTactic.scala
+++ b/core/src/main/scala/stainless/verification/DefaultTactic.scala
@@ -97,8 +97,9 @@ trait DefaultTactic extends Tactic {
         }
 
       case (a @ ADT(aid, tps, args), path) if a.getConstructor.sort.hasInvariant =>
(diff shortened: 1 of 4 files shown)


romac:
@samarion Ready to be merged

PR #252, opened 2018-04-30
Add support for super calls

romac:
**TODO:**

- [x] Dotty support


core/src/main/scala/stainless/extraction/methods/MethodLifting.scala
```diff
@@ -137,6 +145,7 @@ trait MethodLifting extends inox.ast.SymbolTransformer { self =>
       object transformer extends BaseTransformer {
         override def transform(e: s.Expr): t.Expr = e match {
           case s.This(ct) => arg.toVariable
+          case s.Super(ct) => t.AsInstanceOf(arg.toVariable, default.transform(ct))
```
samarion: Is this cast necessary? Isn't `arg.tpe` a subtype of `default.transform(ct)`?

core/src/main/scala/stainless/extraction/methods/MethodLifting.scala
```diff
@@ -213,6 +223,27 @@ trait MethodLifting extends inox.ast.SymbolTransformer { self =>
...
+        (None, baseImpl)
+      } else {
+        val baseParams = params.map(_.freshen)
+        val baseFun = new t.FunDef(
+          baseIds(fd.id),
+          tparams,
+          baseParams,
+          returnType,
+          t.exprOps.replaceFromSymbols(params.zip(baseParams.map(_.toVariable)).toMap, baseImpl),
+          (flags ++ Seq(t.Synthetic, t.InlineOnce)).filterNot(_ == t.Inline).distinct
```
samarion: I've thought about it a bit more and I don't think we can mark these functions as `Synthetic`. They may contain assertions and postconditions that actually need to be checked.
I suggest we either just keep the super-calls un-inlined which is closer to the actual source code, or we just remove the `Synthetic` flag. For sealed class hierarchies, we could further avoid creating the super functions if they are never called (as an optimization).

core/src/main/scala/stainless/extraction/methods/MethodLifting.scala
```diff
@@ -108,13 +109,20 @@ trait MethodLifting extends inox.ast.SymbolTransformer { self =>
       case ValOverride(cid, vd) => Seq(cid -> Right(vd))
     }
 
+    // We need fresh ids for the baseFuns, per function and per class.
+    val baseIds = memoize[Identifier, Identifier] { _.freshen }
```
samarion: Instead of using `memoize` here, maybe compute the mapping of all necessary identifiers. See the optimization remark below.

frontends/scalac/src/main/scala/stainless/frontends/scalac/FragmentChecker.scala
```diff
@@ -108,6 +110,13 @@ trait FragmentChecker extends SubComponent { _: StainlessExtraction =>
             && !sym.isNonBottomSubClass(definitions.AnnotationClass))
             reportError(tree.pos, "Only abstract classes, case classes and objects are allowed in Stainless.")
 
+          val parents = impl.parents.map(_.tpe).collect {
```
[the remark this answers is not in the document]
romac: That's some weird code indeed, although I can't use `exists` because I need the number of parents. Rewrote it as:

```scala
val parents = impl.parents.map(_.tpe).filterNot(ignoreClasses)
```

frontends/stainless-dotty/src/it/scala/stainless/DottyExtractionSuite.scala
```diff
@@ -4,22 +4,29 @@ package stainless
...
     "extraction/invalid/Variance.scala",
     "extraction/invalid/TypeMember.scala",
     "extraction/invalid/TypeParams.scala",
     "extraction/invalid/Println.scala",
     "extraction/invalid/CtorParams.scala",
     "extraction/invalid/ClassBody.scala",
-    "extraction/invalid/Require.scala")
+    "extraction/invalid/Require.scala",
+    "extraction/invalid/SuperAbstract.scala",
+    "extraction/invalid/MultipleInheritance.scala")
```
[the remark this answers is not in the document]
romac: The implementation is still WIP for Dotty, I haven't yet figured out how to extract a super type (Dotty apparently represents it as a fairly complex `AndType`).

core/src/main/scala/stainless/extraction/methods/MethodLifting.scala
romac: It does not seem to be necessary no, but I figured that would be closer to Scala's semantics, even though I now realize that it's not possible to reference a `super` object so it's kinda of a moot point. Will remove the cast.

core/src/main/scala/stainless/extraction/methods/MethodLifting.scala
romac: Good point. I'm fine with both solutions, although the first one seems cleaner to me.

Yeah I haven't gotten to optimise that yet, I first wanted to just make it work. I can do that in a separate PR if that's okay.

samarion: Let's go with no flags then.

Ok for optimizing in a future PR.

romac:
@samarion Ready for review.

romac:
Superseded by #253.

PR #253, opened 2018-05-02
Laws

romac:
## TODO

- [x] Rebase
- [ ] Set proper positions on `law` VCs
- [ ] Write documentation


core/src/main/scala/stainless/extraction/laws/Trees.scala
```diff
@@ -0,0 +1,32 @@
+/* Copyright 2009-2018 EPFL, Lausanne */
+
+package stainless
+package extraction
+package laws
+
+import inox.utils.{NoPosition, Position}
+import scala.collection.mutable.{Map => MutableMap}
+
+trait Trees extends methods.Trees { self =>
```
samarion: I don't think you need new trees here, you can just add a preceding phase during `methods` extraction.
See `extraction.oo.package` for an example of multiple extraction steps on the same trees.

romac:
Sounds good to me

core/src/main/scala/stainless/extraction/methods/SuperCalls.scala
```diff
@@ -0,0 +1,172 @@
...
+  override protected def getContext(symbols: s.Symbols) = new TransformerContext(symbols)
+
+  protected class TransformerContext(val symbols: s.Symbols) extends oo.TreeTransformer {
+    override val s: self.s.type = self.s
+    override val t: self.t.type = self.t
+
+    import s._
+    import symbols._
+
+    private def findFirstSuperDef(symbol: Symbol, ct: s.ClassType): Option[(s.FunDef, s.ClassDef)] = {
```
samarion: It seems to me you can return the `SymbolIdentifier` instead of `s.FunDef` and avoid having to use `unsafeToSymbolIdentifier` in the `superId` map definition.

core/src/main/scala/stainless/extraction/methods/SuperCalls.scala
```diff
@@ -0,0 +1,172 @@
...
+          case _ => Set()
+
+        } (fd.fullBody)
+
+        superRefs.toSeq
+      }
+    }.toMap
+
+    override def transform(e: s.Expr): t.Expr = e match {
+      case s.MethodInvocation(s.Super(ct), id, tps, args) => super.transform {
```
samarion: Avoid using braces in Scala for parameters, sometimes this has a different meaning than expected.

core/src/main/scala/stainless/extraction/methods/SuperCalls.scala
```diff
@@ -0,0 +1,172 @@
...
+      case None => Seq(context.transform(fd))
+      case Some(dupId) =>
+        val cid = fd.flags collectFirst { case IsMethodOf(cid) => cid }
+
+        assert(cid.isDefined)
+        assert(symbols.classes.contains(cid.get))
+
+        val cd = symbols.classes(cid.get)
+
+        val freshParams = fd.params map (_.freshen)
```
samarion: You should probably also freshen the type parameters.

core/src/main/scala/stainless/verification/DefaultTactic.scala
```diff
@@ -51,9 +57,10 @@ trait DefaultTactic extends Tactic {
   def generatePostconditions(id: Identifier): Seq[VC] = {
     val fd = getFunction(id)
     (fd.postcondition, fd.body) match {
-      case (Some(post), Some(body)) =>
+      case (Some(post @ Lambda(Seq(res), _)), Some(body)) if !res.flags.exists(_.name == Unchecked)=>
```
samarion: The equality `_.name == Unchecked` can never be true (incompatible types).
Also missing a space between `)` and `=>` :)

romac:
Blocked by #316

romac:
Blocked by #318

PR #254, opened 2018-05-04
Add `@inlineOnce` annotation

romac:
This annotation works similarly to `@inline`, with the exception that it handles recursive functions and  will only inline them within themselves once.

### DIFF
--- a/core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
+++ b/core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
@@ -28,26 +28,22 @@ trait FunctionInlining extends inox.ast.SymbolTransformer { self =>
         val hasInlineFlag = tfd.fd.flags contains Inline
         val hasInlineOnceFlag = tfd.fd.flags contains InlineOnce
 
-        if (hasInlineFlag && transitivelyCalls(tfd.fd, tfd.fd)) {
-          throw MissformedStainlessCode(tfd.fd, "Can't inline recursive function, use @inlineOnce instead")
-        }
-
         def willInline = hasInlineFlag || (hasInlineOnceFlag && !inlinedOnce.contains(tfd.id))
+
         if (!willInline) {
           return fi
         }
 
         val body = exprOps.withoutSpecs(tfd.fullBody)
         val uncheckedBody = body match {
+          case None => NoTree(tfd.returnType)
           case Some(body) => annotated(body, Unchecked)
-          case None if hasInlineOnceFlag => NoTree(tfd.returnType)
-          case None => throw MissformedStainlessCode(tfd.fd, "Inlining function with empty body: not supported, use @inlineOnce instead")
         }
 
         val pre = exprOps.preconditionOf(tfd.fullBody)
         def addPreconditionAssertion(e: Expr) = pre match {
           case None => e
-          case Some(pre) => Assert(pre, Some("Inlined precondition of " + tfd.fd.id.name), e).copiedFrom(fi)
+          case Some(pre) => Assert(pre, Some("Inlined precondition of " + tfd.id.name), e).copiedFrom(fi)
         }
 
         val post = exprOps.postconditionOf(tfd.fullBody)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
```diff
@@ -13,65 +13,100 @@ trait FunctionInlining extends inox.ast.SymbolTransformer { self =>
...
       })
+
+    val inlinedOnceFuns = symbols.functions.values.filter(_.flags contains InlineOnce).map(_.id).toSet
+    val pruned = afterInlining.functions.values.toSeq.flatMap { fd =>
+      lazy val callers = afterInlining.transitiveCallers(fd).map(_.id)
+      val isCandidate = inlinedOnceFuns.contains(fd.id) && fd.flags.contains(Synthetic)
+      val isRecursive = afterInlining.isRecursive(fd.id)
+      // We can remove synthetic functions which were marked @inlineOnce if they are either
+      // non-recursive, or if they are recursive but only called recursively by themselves.
+      if (isCandidate && (!isRecursive || callers == Set(fd.id))) None
```
samarion: I wonder whether we could remove synthetic inline-once functions if all their callers were also synthetic and inline-once. WDYT?

core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
```diff
@@ -13,65 +13,100 @@ trait FunctionInlining extends inox.ast.SymbolTransformer { self =>
     import symbols._
     import CallGraphOrderings._
 
-    object transformer extends ast.TreeTransformer {
-      val s: self.s.type = self.s
-      val t: self.t.type = self.t
-    }
+    class Inliner(within: FunDef, inlinedOnce: Set[Identifier] = Set()) extends s.SelfTreeTransformer {
```
samarion: Was it necessary to use a `SelfTreeTransformer` here? This means we'll be performing an extra tree copy so it would be ideal if we could skip it (if possible and not overly complex).
BTW, it seems to me you could simply store `within` as an `Identifier` instead of the full `FunDef` object.

core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
```diff
@@ -13,65 +13,100 @@ trait FunctionInlining extends inox.ast.SymbolTransformer { self =>
...
+        case _ => super.transform(expr)
+      }
+
+      private def inlineFunctionInvocations(fi: FunctionInvocation): Expr = {
+        val (tfd, args) = (fi.tfd, fi.args)
+
+        val hasInlineFlag = tfd.fd.flags contains Inline
+        val hasInlineOnceFlag = tfd.fd.flags contains InlineOnce
+
+        if (hasInlineFlag && transitivelyCalls(tfd.fd, tfd.fd)) {
```
samarion: I wonder if it wouldn't be cleaner to perform the inline soundness checks before starting the transformation.
Something like
```scala
for (fd <- symbols.functions.values) { /* check all requirements */ }
/* perform inlining transformation */
```

core/src/main/scala/stainless/extraction/inlining/Trees.scala
```diff
@@ -7,10 +7,12 @@ package inlining
 trait Trees extends extraction.Trees { self =>
 
   case object Inline extends Flag("inline", Seq())
+  case object InlineOnce extends Flag("inlineOnce", Seq())
```
samarion: Did you get to try the `inline(force = true)` or did you decide against it?

core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
romac: It seems to be that there's no easy way to get around it, because I need to be able to call `new Inliner(/* ... */).transform` on the result of the inlining, and I can't if `Inliner` is not a `SelfTreeTransformer`. Do you see a way around it?

core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
romac: Yes, that should work, I'll write some tests and report back.

core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
romac: Agreed

core/src/main/scala/stainless/extraction/inlining/Trees.scala
romac: Well, I didn't try (yet) because `@inline` is a standard Scala annotation, and I wasn't sure whether we wanted to override it with our own (and add some check if FragmentChecker), or if we should just add another one. The main issue I see with defining our own annotation called `@inline` as well, is that, when compiling the code with scalac, the compiler will then not inline such functions.

What do you think?

core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
samarion: Ah yes, you're right. I hadn't noticed that.

core/src/main/scala/stainless/extraction/inlining/Trees.scala
samarion: Oh right, I actually remember it was a bit of a pain point with Dotty as well... Ok let's keep `@inlineOnce` for now.

PR #255, opened 2018-05-07
Do not inject chooses anymore when generating VCs

romac:


romac:
@samarion Does that work for you, or would you rather I annotate the injected `choose`s as unchecked, so that we don't generate VCs for functions with such bodies?

core/src/main/scala/stainless/verification/VerificationComponent.scala
```diff
@@ -30,28 +30,27 @@ object VerificationComponent extends SimpleComponent {
...
 
-    val res = VerificationChecker.verify(encoder.targetProgram, ctx)(vcs).map(_.mapValues {
+    // While we do not need to inject chooses for generating VCs, we need it for
+    // evaluating the models under `--check-models`.
+    val chooseEncoder = inox.ast.ProgramEncoder(encoder.targetProgram)(ChooseInjector(encoder.targetProgram))
+
+    val res = VerificationChecker.verify(chooseEncoder.targetProgram, ctx)(vcs).map(_.mapValues {
       case VCResult(VCStatus.Invalid(VCStatus.CounterExample(model)), s, t) =>
-        VCResult(VCStatus.Invalid(VCStatus.CounterExample(model.encode(encoder.reverse))), s, t)
+        VCResult(VCStatus.Invalid(VCStatus.CounterExample(model.encode(chooseEncoder.reverse))), s, t)
```
samarion: I think you need to use the full encoder here. Something like:
```scala
val assertionEncoder = inox.ast.ProgramEncoder(p)(assertions)
val vcs = VerificationGenerator.gen(assertionEncoder.targetProgram, ctx)(funs)
```
and then
```scala
val encoder = inox.ast.ProgramEncoder(p)(assertions andThen chooses)
val res = VerificationChecker.verify(encoder.targetProgram, ctx)(vcs).map(_.mapValues {
  case VCResult(VCStatus.Invalid(VCStatus.CounterExample(model)), s, t) =>
    VCResult(VCStatus.Invalid(VCStatus.CounterExample(model.encode(encoder.reverse))), s, t)
  ...
```

Otherwise the cast in `results` isn't actually valid (although it might work at runtime due to erasure).

romac: Good catch, fixed!

Issue #241, opened 2018-04-17
'Unexpected type _ >: Nothing <: Any' error in TypeEncoding


romac:
The following program yields to a `Unexpected type _ >: Nothing <: Any` error in TypeEncoding.

```scala
object BugConst {

  abstract class Semigroup[M] {
    def append(left: M, right: M): M
  }

  case class Const[A](value: A)

  case class ConstSemigroup[X](x: X) extends Semigroup[Const[X]] {
    def append(left: Const[X], right: Const[X]): Const[X] = Const(x)
  }

  def constSemigroup[X](x: X): Semigroup[Const[X]] = ConstSemigroup[X](x)
}
```

The wildcard type seems to be (rightly) inferred during `MethodLifting`, as printing the symbols after that phase yields:

```scala
class Const$0[A$27](value$2: A$27)
class Semigroup$0[M$0]
class ConstSemigroup$0[X$0](x$58: X$0) extends Semigroup$0[Const$0[X$0]]

def constSemigroup$0[X$1](x$59: X$1): Semigroup$0[Const$0[X$1]] = ConstSemigroup$0[X$1](x$59)

def append$2[M$1](thiss$0: Semigroup$0[M$1], left$2: M$1, right$2: M$1): M$1 = if (thiss$0.isInstanceOf[ConstSemigroup$0[_ >: Nothing <: Any]]) {
  append$3[_ >: Nothing <: Any](
    thiss$0.asInstanceOf[ConstSemigroup$0[_ >: Nothing <: Any]],
    left$2.asInstanceOf[Const$0[_ >: Nothing <: Any]],
    right$2.asInstanceOf[Const$0[_ >: Nothing <: Any]]
  ).asInstanceOf[M$1]
} else {
  <empty tree>[M$1]
}

def append$3[X$2](thiss$1: ConstSemigroup$0[X$2], left$3: Const$0[X$2], right$3: Const$0[X$2]): Const$0[X$2] = Const$0[X$2](thiss$1.x$58)
```

romac:
WIP at https://github.com/romac/stainless/tree/type-bounds

samarion:
Fix with new type encoding.

PR #257, opened 2018-05-15
WIP: Inner classes

### LINKED ISSUE #36: Support for anonymous class
I believe this could be part for the supported subset of Scala. Here is a minimised example that doesn't pass extraction:

```scala
object Test {
  val o = new Object {}
  println(o)
}
```

Scalac will desugar that code into a local class definition and an instantiation. It looks like nested class definition are at fault; the following also fails to pass extraction:

```scala
object Test {
  class C {
    def f: Unit = {
      class D
      ()
    }
  }
}
```

romac:
## TODO

- [x] Dotty support
- [x] Handle class selectors (scalac/dotc)
- [x] Handle field assignments (scalac/dotc)
- [ ] Make sure we reject unsupported programs
- [ ] Fix `typeBounds` for `LocalClassType`
- [ ] Fix pretty printer
- [ ] Write documentation

## Postponed

- [ ] Invocations of local functions

Fixes #36


romac:
Currently blocked by #241

PR #263, opened 2018-05-31
Improve reporting of ill-formed programs

### LINKED ISSUE #218: Internal error with case classes without a base case
CC: @jad-hamza

```scala
object Main {
  sealed abstract class D
  case class A(d: D) extends D
  case class B(a: A) extends D
  
  def f(d: D) = ()
}
```

> [Internal] Not well formed definition @invariant(inv)

Changing the program as follow resolves the issue

```scala
object Main {
  sealed abstract class D
  case class C() extends D
  case class A(d: D) extends D
  case class B(a: A) extends D
  
  def f(d: D) = ()
}
```

romac:
Close #218


romac:
Merged in 79aca4aa. To be closed along with #218 when merging branch `romac/parallel-pipeline`.

Issue #85, opened 2017-10-16
Documentation link style


mantognini:
As mentioned in https://github.com/epfl-lara/stainless/pull/76#discussion_r141372449

    :doc:`title <target>`

are not properly supported by GitHub. An alternative style would be welcome to make the doc easily browsable directly from this repo.

samarion:
Seems like a good lead: https://github.com/spatialaudio/nbsphinx/blob/6edd6993851e8dbd0bf6fee91aab55297e04be9a/src/nbsphinx.py#L1004-L1088

vkuncak:
Or, we can just wait for github to support sphinx...

PR #264, opened 2018-05-31
Fix documentation links

andrevidela:
Some links in the documentation in `core/src/sphinx/*` were broken and did not follow the `.rst` spec. This pull request aims to fix those links in a consistent manner.


romac:
Thanks! Unfortunately, while this allows browsing the documentation on GitHub, it also breaks these very same links in the generated Sphinx doc.

See #85 for a potential lead if you are interested in getting this to work.

You can find an online but slightly out-of-date (eg. installation section) version of the documentation on http://stainless.epfl.ch/doc/

andrevidela:
Alright I'll look into it

PR #272, opened 2018-06-18
[WIP] Extern types

romac:
> **Warning**: This is just a proof-of-concept for now, as only `ClassType`s can be annotated, and their annotations might very well be lost during the pipeline.

While we can already use `@extern` methods to hide
calls to methods defined by existing Scala libraries
which either make use of Scala features we do not support,
or are provided via the classpath, we currently have no way
of wrapping the types exported by such libraries, because
the registry will complain that such types cannot be found
amongst the symbols.

This palliate to this problem, this PR adds support for annotating
types with the `@extern` annotation, which basically tells the
registry to effectively ignore such types from its dependency graph.

Attached is a simple testcase featuring a wrapper for
`scala.collection.concurrent.TrieMap`.


romac:
An alternative would be to extract any type annotated with ‘@extern’ as ‘Any’, or perhaps just some synthetic case class.

Another open question is that of equality between classes holding values of such types.

romac:
Subsumed by #274.

PR #274, opened 2018-06-18
[WIP] Ignored fields

romac:
While we can already use `@extern` methods to hide
calls to methods defined by existing Scala libraries
which either make use of Scala features we do not support,
or are provided via the classpath, we currently have no way
of wrapping the types exported by such libraries, because
the registry will complain that such types cannot be found
amongst the symbols.

This palliate to this problem, this PR adds support for annotating
fields with the `@ignore` annotation, which effectively replace
the type of the field by `Int`. It is safe and sound to do so,
because such fields are not meant to be accessed via non-`@extern`
methods, and we will thus never actually care about their real type.

Attached is a simple testcase featuring a wrapper for
`scala.collection.concurrent.TrieMap`.


romac:
@samarion Ready for review.

frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
```diff
@@ -255,11 +255,20 @@ class CodeExtraction(inoxCtx: inox.Context, cache: SymbolsContext)(implicit val
...
       val flags = annotationsOf(vdSym, ignoreOwner = true)
+      val isIgnored = flags contains xt.Ignore
+
+      // Flags marked @ignore are extracted as having type BigInt, in order
+      // for us to not have to extract their type while keeping a value
+      // around for equality/effect analysis.
+      val tpe = if (isIgnored) xt.IntegerType()
+                else stainlessType(vd.tpt.tpe)(fieldCtx, vd.pos)
+
+      // @romac: TODO - Once we have @pure, mark as var if not @pure
```
samarion: Let's mark `@ignore`d fields as vars for now and change them to vals once we have `@pure`.

frontends/scalac/src/main/scala/stainless/frontends/scalac/CodeExtraction.scala
```diff
@@ -302,10 +302,18 @@ trait CodeExtraction extends ASTExtractors {
...
       val flags = annotationsOf(sym, ignoreOwner = true)
+      val isIgnored = flags contains xt.Ignore
+
+      // Flags marked @ignore are extracted as having type BigInt, in order
+      // for us to not have to extract their type while keeping a value
+      // around for equality/effect analysis.
+      val tpe = if (isIgnored) xt.IntegerType().setPos(vd.pos)
+                else stainlessType(vd.tpt.tpe)(tpCtx, vd.pos)
+
+      // @romac: TODO - Once we have @pure, mark as var if not @pure
```
samarion: Same remark as for dotty frontend.

frontends/scalac/src/it/scala/stainless/verification/VerificationSuite.scala
```diff
@@ -31,6 +31,10 @@ trait VerificationSuite extends ComponentTestSuite {
     case "verification/valid/IntSetInv" => WithContext(ctx.withOpts(inox.solvers.optAssumeChecked(false)))
     case "verification/valid/IntSetUnit" => WithContext(ctx.withOpts(inox.solvers.optAssumeChecked(false)))
 
+    // Otherwise fail because validation of lucky models reports errors which should be ignored
+    // TODO: Re-enable when bug fixed in Inox
+    case "verification/valid/IgnoredField" => WithContext(ctx.withOpts(inox.solvers.unrolling.optFeelingLucky(false)))
```
samarion: What bug is this? Could you maybe open an issue with a testcase?

romac: See https://github.com/epfl-lara/inox/issues/74

romac:
@samarion Rebased and ready for review/merge.

PR #275, opened 2018-06-19
Fix bug where models for counter-examples to @extern functions were deemed invalid

romac:


core/src/main/scala/stainless/verification/ChooseInjector.scala
```diff
@@ -36,9 +36,20 @@ trait ChooseInjector extends inox.ast.SymbolTransformer {
...
-      })
+        val newBody = if ((fd.flags contains Extern) || (fd.flags contains Opaque)) {
+          val Lambda(Seq(vd), post) = fd.postOrTrue(symbols)
+
+          fd.precondition(symbols) match {
+            case Some(pre) => Require(pre, Choose(vd, post))
+            case None => Choose(vd, post)
+          }
+        } else {
+          val newBody = injectChooses(body.getOrElse(NoTree(fd.returnType)))
```
samarion: (Comment about line 34 but Github won't let me add a comment outside the diff)

We should also add a `Let` case for the `injectChooses` function just in case one appeared during some transformation.

frontends/common/src/test/scala/stainless/ast/OpaqueSuite.scala
```diff
@@ -37,9 +38,10 @@ class OpaqueSuite extends FunSuite with InputUtils {
     import program.trees._
     import program.trees.dsl._
 
-    val factory = solvers.SolverFactory(program, ctx)
+    val encoded = inox.ast.ProgramEncoder(program)(verification.ChooseInjector(program)).targetProgram
```
samarion: Isn't this `chooses.targetProgram`?

PR #280, opened 2018-07-17
Add support for 'copy' method and default parameters

### LINKED ISSUE #21: Support for generic copy() method
I tried to compile Leon/Stainless code using the generic `copy()` method for generic case classes (see [here](http://stackoverflow.com/questions/7249396/how-to-clone-a-case-class-instance-and-change-just-one-field-in-scala)), but it seems that it is not yet in the Stainless library. Is it possible to provide support for this construct?

romac:
Close #21

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
@@ -9,10 +9,14 @@ import inox._
 trait AntiAliasing
   extends CachingPhase
      with SimpleSorts
-     with SimpleFunctions
      with EffectsAnalyzer
      with EffectsChecker { self =>
   import s._
+
+  override protected type FunctionResult = Option[FunDef]
+
+  override protected def registerFunctions(symbols: t.Symbols, functions: Seq[Option[t.FunDef]]): t.Symbols =
+    symbols.withFunctions(functions.flatten)
 
   protected case class SymbolsAnalysis(symbols: Symbols, effects: EffectsAnalysis) {
     import symbols._
@@ -65,11 +69,15 @@ trait AntiAliasing
   override protected type TransformerContext = SymbolsAnalysis
   override protected def getContext(symbols: Symbols) = SymbolsAnalysis(symbols, EffectsAnalysis(symbols))
 
-  override protected def extractFunction(analysis: SymbolsAnalysis, fd: FunDef): FunDef = {
+  override protected def extractFunction(analysis: SymbolsAnalysis, fd: FunDef): Option[FunDef] = {
     import analysis._
     import symbols._
 
-    checkFunction(fd)(symbols, effects)
+    checkFunction(fd)(symbols, effects) match {
+      case CheckResult.Ok => ()
+      case CheckResult.Skip => return None
+      case CheckResult.Error(err) => throw err
+    }
 
     type Environment = (Set[ValDef], Map[ValDef, Expr], Map[ValDef, LocalFunDef])
     implicit class EnvWrapper(env: Environment) {
@@ -382,7 +390,7 @@ trait AntiAliasing
       rec(effect.receiver, effect.target.path)
     }
 
-    transformer.transform(updateFunction(Outer(fd), Environment.empty).toFun)
+    Some(transformer.transform(updateFunction(Outer(fd), Environment.empty).toFun))
   }
 
   override protected def extractSort(analysis: SymbolsAnalysis, sort: ADTSort): ADTSort =
--- a/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
@@ -4,12 +4,32 @@ package stainless
 package extraction
 package imperative
 
+sealed abstract class CheckResult
+object CheckResult {
+  case object Ok extends CheckResult
+  case object Skip extends CheckResult
+  case class Error(err: ImperativeEliminationException) extends CheckResult
+}
+
 trait EffectsChecker { self: EffectsAnalyzer =>
   import s._
 
-  def checkFunction(fd: FunDef)(symbols: Symbols, effects: EffectsAnalysis): Unit = {
+  protected def checkFunction(fd: FunDef)(symbols: Symbols, effects: EffectsAnalysis): CheckResult = {
     import symbols._
     import effects._
+
+    def isMutableSynthetic(id: Identifier): Boolean = {
+      val fd = symbols.functions(id)
+      fd.flags.contains(Synthetic) &&
+      fd.params.exists(vd => isMutableType(vd.tpe)) &&
+      !exprOps.withoutSpecs(fd.fullBody).forall(isExpressionFresh)
+    }
+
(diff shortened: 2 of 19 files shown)


romac:
@samarion Ready for review.

romac:
@samarion Ready for review.

frontends/dotty/src/main/scala/stainless/frontends/dotc/ASTExtractors.scala
```diff
@@ -442,8 +456,10 @@ trait ASTExtractors {
             !dd.symbol.is(Label)
           ) || (
             (dd.symbol is Synthetic) &&
-            (dd.symbol is Implicit) &&
-            !(getAnnotations(tpt.symbol) contains "ignore")
+            (
+              canExtractSynthetic(dd.symbol) ||
+              !(getAnnotations(tpt.symbol) contains "ignore")
```
samarion: Why did this become an `or`?

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
```diff
@@ -4,13 +4,36 @@ package stainless
...
 
-  def checkFunction(fd: FunDef)(symbols: Symbols, effects: EffectsAnalysis): Unit = {
+  protected def checkFunction(fd: FunDef)(symbols: Symbols, effects: EffectsAnalysis): CheckResult = {
     import symbols._
     import effects._
 
+    def isMutableSynthetic(id: Identifier): Boolean = {
+      val fd = symbols.functions(id)
+      (fd.flags contains Synthetic) && (fd.params match {
+        case receiver :: _ =>
```
samarion: Looking only at the receiver seems a bit brittle. Why not generalize to all params?

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -382,7 +390,8 @@ trait AntiAliasing
       rec(effect.receiver, effect.target.path)
     }
 
-    transformer.transform(updateFunction(Outer(fd), Environment.empty).toFun)
+    val res = transformer.transform(updateFunction(Outer(fd), Environment.empty).toFun)
+    Some(res)
```
samarion: Remove superfluous let binding?

frontends/scalac/src/main/scala/stainless/frontends/scalac/CodeExtraction.scala
```diff
@@ -319,18 +319,12 @@ trait CodeExtraction extends ASTExtractors {
       case EmptyTree =>
         // ignore
 
-      case t if (
-        (annotationsOf(t.symbol) contains xt.Ignore) ||
-        (t.symbol.isSynthetic && !t.symbol.isImplicit)
-      ) =>
+      case t if annotationsOf(t.symbol) contains xt.Ignore =>
```
samarion: Aren't you missing a condition here? What about `t.symbol.isSynthetic && !canExtractSynthetic(t.symbol)`?

frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
```diff
@@ -377,7 +373,8 @@ class CodeExtraction(inoxCtx: inox.Context, cache: SymbolsContext)(implicit val
     var flags = annotationsOf(sym) ++
       (if ((sym is Implicit) && (sym is Synthetic)) Set(xt.Inline, xt.Synthetic) else Set()) ++
       (if (sym is Inline) Set(xt.Inline) else Set()) ++
-      (if (!(sym is Method)) Set(xt.IsField(sym is Lazy)) else Set())
+      (if (!(sym is Method)) Set(xt.IsField(sym is Lazy)) else Set()) ++
+      (if (isCopyMethod(sym) || isDefaultGetter(sym)) Set(xt.Synthetic, xt.Inline) else Set())
```
samarion: I'm not sure the `copy` method should be inlined. It seems like that would be a bit surprising for the user, no? But the default getter should indeed be inlined.

frontends/scalac/src/main/scala/stainless/frontends/scalac/CodeExtraction.scala
```diff
@@ -442,7 +436,8 @@ trait CodeExtraction extends ASTExtractors {
 
     var flags = annotationsOf(sym) ++
       (if (sym.isImplicit && sym.isSynthetic) Set(xt.Inline, xt.Synthetic) else Set()) ++
-      (if (sym.isAccessor) Set(xt.IsField(sym.isLazy)) else Set())
+      (if (sym.isAccessor) Set(xt.IsField(sym.isLazy)) else Set()) ++
+      (if (isDefaultGetter(sym) || isCopyMethod(sym)) Set(xt.Synthetic, xt.Inline) else Set())
```
samarion: Same remark as above about inlining the copy method.

romac:
Looks like we should add a ADT's invariant as a precondition to the `copy` method, as otherwise such code goes through without errors:

```scala
  case class Foo(x: BigInt) {
    require(x != 0)
  }

  def updateFoo(foo: Foo) = {
    foo.copy(x = 0)
  }
```

romac:
@samarion Rebased and ready for review/merge.

core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
```diff
@@ -64,7 +66,7 @@ trait FunctionInlining extends CachingPhase with IdentitySorts { self =>
             Let(vd, e, Assume(post, vd.toVariable).copiedFrom(fi)).copiedFrom(fi)
         }
 
-        val newBody = addPreconditionAssertion(addPostconditionAssumption(uncheckedBody))
+        val newBody = addPreconditionAssertion(addPostconditionAssumption(body))
```
samarion: We can't assume the post on `@synthetic` methods as it won't be checked anywhere. It should be inlined into an assertion here.

frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
```diff
@@ -377,7 +373,9 @@ class CodeExtraction(inoxCtx: inox.Context, cache: SymbolsContext)(implicit val
     var flags = annotationsOf(sym) ++
       (if ((sym is Implicit) && (sym is Synthetic)) Set(xt.Inline, xt.Synthetic) else Set()) ++
       (if (sym is Inline) Set(xt.Inline) else Set()) ++
-      (if (!(sym is Method)) Set(xt.IsField(sym is Lazy)) else Set())
+      (if (!(sym is Method)) Set(xt.IsField(sym is Lazy)) else Set()) ++
+      (if (isDefaultGetter(sym)) Set(xt.Synthetic, xt.Inline, xt.Annotation("specialMethod", Seq("defaultGetter"))) else Set()) ++
+      (if (isCopyMethod(sym)) Set(xt.Synthetic, xt.Inline, xt.Annotation("specialMethod", Seq("copy"))) else Set())
```
samarion: Let's not add the `specialMethod` flags as long as they're not necessary.

PR #286, opened 2018-07-25
Add ghost annotation and its checks.

dragos:
~~There is no rewrite yet, just checking valid use of ghost code.~~


romac:
@samarion Ready for review

romac:
Rebased in #302

PR #292, opened 2018-08-09
Update Stainless to latest Inox

romac:

PR #295, opened 2018-08-13
Ensure inner functions conserve their flags when deconstructed

### LINKED ISSUE #294: Partial evaluator is not triggered on inner functions
eg.

```scala
import stainless.lang._
import stainless.annotation._

object test {

  def prop(x: BigInt): BigInt = {

    @partialEval
    def go(y: BigInt): BigInt = if (y <= 0) 0 else go(y - 1)

    go(42)
  } ensuring { _ == 0 }

}
```

romac:
Close #294


romac:
To rework when #292 is merged.

romac:
@samarion Ready for review/merge.

samarion:
I can't reproduce the bug this was designed to fix. After taking a closer look, I'm not sure I understand how the flags could disappear since they shouldn't be removed from the `name` value definition.

romac:
Indeed, can't reproduce either anymore. Might be due to the newly updated deconstructors within Inox.

PR #302, opened 2018-08-16
Add ghost annotation and its checks

### LINKED ISSUE #269: Support for ghost state
- [x] Add a new `@ghost` annotation
- [x] Mark appropriate methods from Stainless' library as ghost
- [x] Within Stainless, check that trees marked `@ghost` are never referenced outside of ghost trees.
- [x] Implement as a new compiler phase, within the sbt plugin, that removes trees marked `@ghost`.

romac:
Rebased version of #286.

Close #269


romac:
@samarion Ready for review/merge.

romac:
Rebased.

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
```diff
@@ -46,14 +46,19 @@ trait EffectsChecker { self: EffectsAnalyzer =>
...
                   throw ImperativeEliminationException(e, "Illegal aliasing: " + e)
               }
+
+              if (vd.flags.contains(Ghost) && effects(e).nonEmpty) {
+                throw ImperativeEliminationException(e, "Right-hand side of ghost variable must be pure")
+              }
+
               super.traverse(l)
 
             case l @ LetVar(vd, e, b) if isMutableType(vd.tpe) =>
```
samarion: I believe we need the same check as in the `Let` case here. Maybe you can handle them both in a single case?

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
```diff
@@ -75,12 +80,34 @@ trait EffectsChecker { self: EffectsAnalyzer =>
...
+            case fi @ FunctionInvocation(id, tps, args) if fi.tfd.params.exists(_.flags contains Ghost) =>
+              fi.tfd.params.zip(args)
+                .filter { case (vd, _) => vd.flags contains Ghost }
+                .foreach { case (vd, arg) =>
+                  if (effects(arg).nonEmpty)
+                    throw ImperativeEliminationException(arg,
+                      s"Argument to ghost parameter `${vd.id}` of method `${fi.id}` must be pure")
+                }
+
+              super.traverse(fi)
```
samarion: We probably need a similar case for `ApplyLetRec`, or are they not allowed to have ghost parameters?
Note that you'll probably need to track the `LocalFunDef` for these, but you can pre-compute the mapping from identifeirs to `FunAbstraction` in the symbol table (see effects analysis) and then handle both the `FunctionInvocation` and `ApplyLetRec` in a single case.

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
```diff
@@ -75,12 +80,34 @@ trait EffectsChecker { self: EffectsAnalyzer =>
...
               }
+
+              cons.fields.zip(args)
+                .filter { case (vd, _) => vd.flags contains Ghost }
+                .foreach { case (vd, arg) =>
+                  if (effects(arg).nonEmpty)
+                    throw ImperativeEliminationException(arg,
+                      s"Argument to ghost field `${vd.id}` of class `${id}` must be pure")
+                }
+
```
samarion: I think we still need cases for `Assignment` and `FieldAssignment` as well.

romac:
Good catch, I indeed missed a few cases sorry. Will take care of it when I am back.

jad-hamza:
I have a change here: https://github.com/jad-hamza/stainless/tree/ghost-checks, not sure if it's ok

PR #308, opened 2018-08-24
Add debug output of symbols before and after each transformation

jad-hamza:
This adds debug output (for `--debug=trees`) of the symbols before and after each transformation. `debugTransformation = false` is set to `false` for compound transformations (in `andThen`) to avoid duplication of output.

The output is still a bit hard to read sometimes (maybe because I didn't set `debugTransformation` to `false` somewhere), but it is generally ok. We could add names to each transformations to have a nicer output.

This one is OK:
> Symbols before extraction stainless.extraction.imperative.ImperativeCleanup$$anon$1@1ccfcec5

This one less so:
> Symbols before extraction stainless.extraction.ExtractionPipeline$$anon$3@1212d8a5

### DIFF
--- a/core/src/main/scala/stainless/Component.scala
+++ b/core/src/main/scala/stainless/Component.scala
@@ -40,6 +40,13 @@ object optDebugObjects extends inox.OptionDef[Seq[String]] {
   val usageRhs = "o1,o2,..."
 }
 
+object optDebugPhases extends inox.OptionDef[Seq[String]] {
+  val name = "debug-phases"
+  val default = Seq[String]()
+  val parser = inox.OptionParsers.seqParser(inox.OptionParsers.stringParser)
+  val usageRhs = "p1,p2,..."
+}
+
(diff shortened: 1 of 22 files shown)


jad-hamza:
Now depends on https://github.com/epfl-lara/inox/pull/75 for the `filterObjects` function.

jad-hamza:
No more dependency on an Inox change, but I introduced some ugly casts to access `classes` from `oo.trees.Symbols`.

jad-hamza:
Rebased on top of https://github.com/epfl-lara/stainless/pull/318, with a commit to put both debug outputs at the same place. Tree outputs and position checkers can be filtered with ` --debug-phases`, and only get printed when `debugTransformation` is true (i.e. not for compound phases built from `andThen`).

Additionally, `--debug-objects` can be used to filter the trees to print.

jad-hamza:
Superseded by https://github.com/epfl-lara/stainless/pull/333

PR #313, opened 2018-08-29
Support for trait variables

### LINKED ISSUE #285: Trait variables
Add support for variables within trait definitions.
These should be injectable into the base class constructors and handled in imperative desugaring.

Also see #283

romac:
Close #285


romac:
Works with scalac, now on to Dotty!

PR #317, opened 2018-08-31
Adapt code to new BVType Inox API

jad-hamza:
Depends on https://github.com/epfl-lara/inox/pull/76

### DIFF
--- a/core/src/main/scala/stainless/codegen/CodeGeneration.scala
+++ b/core/src/main/scala/stainless/codegen/CodeGeneration.scala
@@ -196,7 +196,7 @@ trait CodeGeneration { self: CompilationUnit =>
     case Int16Type() => "S"
     case Int32Type() => "I"
     case Int64Type() => "J"
-    case BVType(true, _) => "L" + BitVectorClass + ";"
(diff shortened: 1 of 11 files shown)


jad-hamza:
At the moment the CodeGenEvaluatorSuite is not going through:
```
[info] stainless.evaluators.CodeGenEvaluatorSuite *** ABORTED *** (970 milliseconds)
[info]   java.lang.VerifyError: (class: Stainless$CodeGen$Expr$42, method: eval signature: (Lstainless/codegen/runtime/Monitor;)S) Expecting to find integer on stack
```
The problem came when I did this last change: https://github.com/jad-hamza/stainless/blob/49fde7f7f52f44b8b992e54223e6fa6e4cb93276/core/src/main/scala/stainless/codegen/CodeGeneration.scala#L1634-L1637 changing `case BVType(true, _) =>` to `case BVType(_, _) =>`.

Any tips on how to debug this?

I have to distinguish between signed and unsigned division (and other operators) here somewhere but I didn't do it yet.

samarion:
Merged

PR #324, opened 2018-09-11
Remove Leon-only Isabelle annotations.

romac:
Background: https://github.com/epfl-lara/leon/pull/142


vkuncak:
I would, in fact, prefer that we add back the support for isabelle, in one form or another.

PR #332, opened 2018-09-19
Make cache keys parameterizable by the user

jad-hamza:


jad-hamza:
Superseded by https://github.com/epfl-lara/stainless/pull/333

PR #339, opened 2018-09-20
Assertions for operations on unsigned bitvectors

jad-hamza:
* `a + b`: check that `a + b >= a`
* `a - b`: check that `b <= a`
* `a * b`: check that `a = 0` or `(a * b) / a = b`
* `a / b`: check that `b != 0` (that should be enough for unsigned division?)

### DIFF
--- a/core/src/main/scala/stainless/verification/AssertionInjector.scala
+++ b/core/src/main/scala/stainless/verification/AssertionInjector.scala
@@ -171,15 +171,12 @@ trait AssertionInjector extends ast.TreeTransformer {
     case BVTyped(signed, size, BVShift(rhs0)) if strictArithmetic =>
       val rhs = transform(rhs0)
       val newE = super.transform(e)
-      t.Assert(
-        // Ensure the operation doesn't shift more bits than there are.
-        t.And(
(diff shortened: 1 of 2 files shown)


core/src/main/scala/stainless/verification/AssertionInjector.scala
```diff
@@ -152,7 +174,7 @@ trait AssertionInjector extends ast.TreeTransformer {
       t.Assert(
         // Ensure the operation doesn't shift more bits than there are.
         t.And(
-          t.GreaterEquals(rhs, zero(size, rhs.getPos)).copiedFrom(rhs),
+          t.GreaterEquals(rhs, zero(signed, size, rhs.getPos)).copiedFrom(rhs),
```
samarion: If the bitvector is unsigned, isn't this a useless check?

jad-hamza: Yes that's true, I've removed it in the last commit. This assumes that if `e` is unsigned, then so is its `rhs`, is that safe?

samarion: Yes, I believe so. Otherwise, the program wouldn't type check (according to Inox' type system).

PR #342, opened 2018-09-21
Add pipeline wrappers for debugging trees and positions

### LINKED ISSUE #316: Missing positions on VCs
A lot of VCs are currently missing positions, so I suggest adding a `--debug=position` flag that will enable a checking phase after every `ExtractionPipeline` that will walk the trees while keeping track of the last know position, and report missing positions on every (non-synthetic) tree.

What do you think, @samarion?

jad-hamza:
Fix #316


samarion:
I think a cleaner approach to this would be to provide a trait `DebugPipeline` that wraps an `ExtractionPipeline` with debugging support. You could define it something like:
```scala
trait DebugPipeline extends ExtractionPipeline {
  val name: String
  val underlying: ExtractionPipeline
  override val s: underlying.s.type = underlying.s
  override val t: underlying.t.type = underlying.t
  override val context = underlying.context

  override def extract(symbols: s.Symbols): t.Symbols = {
    val newSymbols = context.timers.extraction.get(name).run(underlying.extract(symbols))
    // call debugging code for the new symbols
    newSymbols
  }

  override def invalidate(id: Identifier) = underlying.invalidate(id)
}

object DebugPipeline {
  def apply(nme: String, pipeline: ExtractionPipeline): ExtractionPipeline {
    val s: pipeline.s.type
    val t: pipeline.t.type
  } = new {
    override val underlying: pipeline.type = pipeline
  } with DebugPipeline {
    override val name: String = nme
  }
}
```

Then, in order to provide debugging for an extraction phase we're interested in, we can simply wrap the phase construction with the `DebugPipeline` as follows:
```scala
DebugPipeline("ADT specialization", AdtSpecialization(trees, trees))
```

We can also wrap multi-phases, for example:
```scala
DebugPipeline(
  "Class elimination",
  AdtSpecialization(trees, trees) andThen
  RefinementLifting(trees, trees) andThen
  TypeEncoding(trees, trees)      andThen
  lowering
)
```

jad-hamza:
Thanks, I've added the `DebugPipeline` and squashed/rebased

core/src/main/scala/stainless/ast/SymbolOps.scala
```diff
@@ -260,23 +262,27 @@ trait SymbolOps extends inox.ast.SymbolOps { self: TypeOps =>
...
+    * Without it, we would lose position information in the resulting tree.
     */
-  def withShared(path: Path, es: Seq[Expr], recons: (Expr, Seq[Expr]) => Expr): Expr = {
+  def withShared(path: Path, es: Seq[Expr], recons: (Expr, Seq[Expr]) => Expr, pos: Position): Expr = {
     import Path._
 
     val (outers, rest) = path.elements span { !_.isInstanceOf[Condition] }
     val bindings = rest collect { case CloseBound(vd, e) => vd -> e }
-    val cond = fold[Expr](BooleanLiteral(true), Let, And(_, _))(rest)
+    val cond = fold[Expr](BooleanLiteral(true).setPos(pos), let, and(_, _))(rest)
```
samarion: You shouldn't use the `let` constructor here as it will rely on `simplifier` to determine purity of the rhs and the simplifier is not guaranteed to work for all types of trees.

The other uses of `let` below have the same issue.

core/src/main/scala/stainless/extraction/DebugPipeline.scala
```diff
@@ -0,0 +1,103 @@
...
+  override val s: underlying.s.type = underlying.s
+  override val t: underlying.t.type = underlying.t
+  override val context = underlying.context
+
+  override def invalidate(id: Identifier) = underlying.invalidate(id)
+
+  // make a String representation for a table of Symbols `s`, only keeping
+  // functions and classes whose names appear in `objs`
+  def symbolsToString(tt: ast.Trees)(s: tt.Symbols, objs: Set[String]): String = {
+    val printerOpts = tt.PrinterOptions.fromContext(context)
```
samarion: I'd declare a single global `printerOpts` in the debug pipeline since the context is fixed.
Also you should precompute the `phases` and `objs` values during the pipeline initialization.

core/src/main/scala/stainless/extraction/DebugPipeline.scala
```diff
@@ -0,0 +1,103 @@
...
+    val printerOpts2 = oo.trees.PrinterOptions.fromContext(context)
+    def objectsToString(m: Iterable[(Identifier,tt.Definition)]): String =
+      m.collect {
+        case (id,d) if objs.isEmpty || objs.contains(id.name) =>
```
samarion: You can keep `objs` as an option of set to differentiate between no option and an empty set of objects (although I agree the empty set of objects doesn't sound that useful).

core/src/main/scala/stainless/extraction/DebugPipeline.scala
```diff
@@ -0,0 +1,103 @@
...
+          d.asString(tt.PrinterOptions.fromContext(context))
```
samarion: Don't create yet another instance of `PrinterOptions`.

core/src/main/scala/stainless/extraction/DebugPipeline.scala
```diff
@@ -0,0 +1,103 @@
...
+      }.mkString("\n\n")
+
+    val functions = objectsToString(s.functions)
+    val sorts = objectsToString(s.sorts)
+    val classes =
+      if (tt == oo.trees)
```
samarion: The condition should be `tt.isInstanceOf[oo.Trees]`, otherwise you won't print classes in `oo.Trees` subtypes such as `methods.Trees`. To avoid a few casts you can change the signature of `objectsToString` to
```scala
def objectsToString(trees: ast.Trees)(m: Iterable[(Identifier, trees.Definition)]): String
```
and use the following
```scala
val classes = tt match {
  case trees: oo.Trees => objectsToString(trees)(trees.classes)
  case _ => ""
}
```

core/src/main/scala/stainless/extraction/DebugPipeline.scala
```diff
@@ -0,0 +1,102 @@
...
+    wrapWith("Sorts", sorts) ++
+    wrapWith("Classes", classes)
+  }
+
+  // `extract` is a wrapper around `super.extract` which outputs trees for
+  // debugging and which outputs position checks
+  override def extract(symbols: s.Symbols): t.Symbols = {
+    implicit val debugSection = inox.ast.DebugSectionTrees
+    val phases = context.options.findOption(optDebugPhases)
+    val objs = context.options.findOption(optDebugObjects).getOrElse(Seq()).toSet
```
samarion: Precompute these at `DebugPhase` creation time.

core/src/main/scala/stainless/extraction/DebugPipeline.scala
```diff
@@ -0,0 +1,102 @@
...
+    // We print debug output for this phase only if the user didn't specify
+    // any phase with --debug-phases, or gave the name of (or a string
+    // contained in) this phase
+    val debug: Boolean =
+      (phases.isEmpty || (phases.isDefined && phases.get.exists(name.contains _)))
+
+    context.reporter.synchronized {
+      val symbolsToPrint = symbolsToString(s)(symbols, objs)
```
samarion: Don't generate the string if it won't be printed. This is slowing the regressions down significantly.

core/src/main/scala/stainless/extraction/DebugPipeline.scala
```diff
@@ -0,0 +1,102 @@
...
+  // debugging and which outputs position checks
+  override def extract(symbols: s.Symbols): t.Symbols = {
+    implicit val debugSection = inox.ast.DebugSectionTrees
+    val phases = context.options.findOption(optDebugPhases)
+    val objs = context.options.findOption(optDebugObjects).getOrElse(Seq()).toSet
+    // We print debug output for this phase only if the user didn't specify
+    // any phase with --debug-phases, or gave the name of (or a string
+    // contained in) this phase
+    val debug: Boolean =
+      (phases.isEmpty || (phases.isDefined && phases.get.exists(name.contains _)))
```
samarion: I think this turns on phase debugging even if you haven't passed the option, no?

core/src/main/scala/stainless/extraction/utils/PositionChecker.scala
```diff
@@ -0,0 +1,53 @@
...
+import inox.utils.{Position, NoPosition}
+
+object DebugSectionPositions extends inox.DebugSection("positions")
+
+/** Inspect trees, detecting missing positions. */
+object PositionChecker {
+
+  private[this] val seen: MutableHashSet[ast.Trees#Expr] = MutableHashSet.empty
+
+  def apply(phaseName: String)(tr: ast.Trees)(context: inox.Context)(before: Boolean): tr.TreeTraverser { val trees: tr.type } = new tr.TreeTraverser {
```
samarion: The `PositionChecker` API looks a bit strange to me. I would just take the symbols as input directly and run the checker on all of them. I feel like the internal state that the traverser is maintaining (i.e. `lastKnownPosition`) means it's not really re-usable, so I don't think it should be returned.

core/src/main/scala/stainless/extraction/utils/PositionChecker.scala
```diff
@@ -0,0 +1,53 @@
...
+      traverse(fd.id)
+      fd.tparams.foreach(traverse)
+      fd.params.foreach(traverse)
+      traverse(fd.returnType)
+      traverse(fd.fullBody)
+      fd.flags.foreach(traverse)
+    }
+
+    override def traverse(e: Expr): Unit = {
+      if (seen contains e) return ()
```
samarion: You might miss some missing positions since they won't participate int the `hashCode` or `equals` checks here.

core/src/main/scala/stainless/extraction/DebugPipeline.scala
jad-hamza: `symbolsToString` is sometimes called for `s` trees, and sometimes for `t` trees, can I use the same `printerOpts` for both?

samarion: No, you'll need different ones. But you should already have a `s.PrinterOptions` from `ExtractionPipeline` so you just need a (differently named) one for `t.PrinterOptions`.
You can then add an implicit `tt.PrinterOptions` argument to `symbolsToString`.

However, I guess this isn't really that important, building a new one is probably almost as expensive as passing the printer options in to `symbolsToString`.

jad-hamza:
I did some changes and rebased. For `phases` and `objs`, should I leave them abstract and initialize them in the companion object?

samarion:
Hmm, I think you can directly initialize them from the `context` in the `DebugPipeline` trait definition.

jad-hamza:
I'm not sure what to do about the `PositionChecker` suggestions, maybe you or @romac could fix them

romac:
> I'm not sure what to do about the `PositionChecker` suggestions, maybe you or @romac could fix them

I'm on it.

jad-hamza:
Thanks! There seems to be an issue with that part as well (looks like the cast fails), I can have a look a bit later. Maybe it would be better to have these `toString` functions that filters the objects directly part of the symbols, so that we can override them when we add classes in `oo.trees.Symbols`. 

```scala
      if (tt.isInstanceOf[oo.Trees])
        objectsToString(oo.trees)(syms.asInstanceOf[oo.trees.Symbols].classes, ooPrinterOpts)
```

romac:
@jad-hamza Can you please pull this branch and rebase it on latest master? I cannot force-push the rebased version myself. It should only take a minute as they were no conflicts (on my machine™️)

jad-hamza:
Done, thanks for the changes!

PR #343, opened 2018-09-24
Modifications to dependency graph

jad-hamza:


core/src/main/scala/stainless/extraction/methods/DependencyGraph.scala
```diff
@@ -79,8 +79,17 @@ trait DependencyGraph extends ast.DependencyGraph with CallGraph {
...
     for (fd <- symbols.functions.values) {
-      overrides(fd) foreach { id => g += SimpleEdge(fd.id, id) }
+      for (oid <- overrides(fd)) {
+        symbols.functions(oid).flags.collectFirst {
+          case IsMethodOf(cid) =>
+            for (n <- g.transitivePred(fd.id) & g.transitivePred(cid)) {
+              g += SimpleEdge(n, oid)
+            }
+        }
+      }
```
samarion: Do you think this needs to be in some sort of fixpoint since we're adding edges based on the current graph?

jad-hamza: Yes probably, should be fixed now

Issue #346, opened 2018-09-25
Allow ghost effects on ghost fields of inductive ADTs


jad-hamza:
The first example is accepted by the ghost checker, but not the second one:

```scala
import stainless.annotation._
import stainless.lang._
import stainless.collection._

object Ghost {
  case class Ghost(@ghost var p: BigInt) {
    def f(x: BigInt) = {
      ghost {
        p = p + 1
      }
    }
  }
}
```

```scala
import stainless.annotation._
import stainless.lang._
import stainless.collection._

object Ghost {
  case class Ghost(@ghost var l: List[BigInt]) {
    def f(x: BigInt) = {
      ghost {
        l = x :: l // Argument to ghost parameter `value` of `ghost` must only have effects on ghost fields
      }
    }
  }
}
```

jad-hamza:
Another test that fails for the same reason (@samarion said effects analysis is not precise enough at the moment)

```scala
import stainless.annotation._
import stainless.collection._

object Ghost {
  case class Ghost(@ghost var l: List[BigInt]) {
    @ghost
    def add(x: BigInt) = {
      l = x :: l
    }
  }
}
```

romac:
Yeah it looks like we lose information about the effect target (ie. the ghost field `l`) because of this [call](https://github.com/epfl-lara/stainless/blob/6e0a6f118ee00b5093616885924a0b79e452a953/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala#L359) to [`truncate`](https://github.com/epfl-lara/stainless/blob/6e0a6f118ee00b5093616885924a0b79e452a953/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala#L320-L342). Commenting the call allows both examples through.

romac:
In both the second and third example, `Ghost` is treated as an inductive ADT as it contains a `List`, and the effect target is thus truncated from `this.l` to just `this`.

samarion:
Fixed in https://github.com/epfl-lara/stainless/commit/2a991ab9a9a1ebc3b5ed1511bd45c870ca237813

PR #348, opened 2018-09-26
Relax ghost checks a little bit in ghost functions

### LINKED ISSUE #345: Ghost check too restrictive
This is currently rejected by the Ghost checker. Should it be ok, since the whole function `f` is ghost?

```scala
import stainless.annotation._

object Ghost {
  case class Ghost(
    @ghost var x: BigInt
  ) {
    @ghost
    def f() = {
      val y = x // Right-hand side of non-ghost variable cannot be ghost
    }
  }
}
```

romac:
Close #345
Related to #346


samarion:
Sorry for not telling you earlier but I fixed the ghost issues in https://github.com/epfl-lara/stainless/commit/a5ea688b416d470e0a976f7133ffdb2c144b6424
However, skipping `Synthetic` functions sounds good to me, we should probably keep that condition.

PR #349, opened 2018-09-27
Treat contract of methods in `stainless.lang.StaticChecks` as ghost

romac:


romac:
I think I should I do the same for `decreases`, right? @jad-hamza @samarion

romac:
Do we want to keep the `GhostAnnotationChecker` in `FragmentChecker` or should I remove it, as those checks are now performed as part of the imperative phase?

romac:
Well, never mind, `GhostAnnotationChecker` actually does more stuff than we do.

romac:
I think we should annotate `stainless.lang.forall` and `stainless.lang.choose` as `@ghost`, as to disallow their use in non-ghost context, where they will throw an exception if left in the compiled code. What do you think @samarion @jad-hamza?

romac:
Build failed because of 

```
[info] -  51: verification/valid/CovariantList solvr=smt-cvc4 lucky=false check=true codegen *** FAILED *** (6 seconds, 230 milliseconds)
[info]   inox.package$FatalError: Unexpected error from cvc4 solver: Solver encountered exception: java.io.IOException: Stream closed
```

Re-starting the build.

Issue #354, opened 2018-10-01
Fix method overrides


samarion:
The new dependency computation that drops children of non-sealed classes has the unfortunate consequence that method overrides are no longer correctly checked. This issue is also problematic for the [laws](https://github.com/epfl-lara/stainless/pull/253) PR where an ad-hoc fix was implemented (but only for laws).

While we're at it, a related issue is that we don't correctly check overrides in `ensureWellFormed`. We need to _at least_ check that
1. the typing relation for overriden symbols,
2. that methods with empty bodies are indeed overriden by non-abstract descendants (note that the frontends won't check this as the empty body `???` is considered as a valid term for them).

It is not strictly necessary to check that methods marked with `override` actually override something as the compiler frontends should perform these checks themselves (although possibly later in the pipeline than Stainless).

samarion:
Fixed by https://github.com/epfl-lara/stainless/pull/355, as well as https://github.com/epfl-lara/stainless/commit/f8b488ba3d61aea058bddd041c8482f179df012e and https://github.com/epfl-lara/stainless/commit/8c9cfe41e262d9d1b34a0e0b44a7816904efdec5

PR #355, opened 2018-10-01
Add dependency from override to overridden method

jad-hamza:
Add dependency from override to overridden method, and Derived flags (to overrides) in synthesized methods (`MethodLifting`)


jad-hamza:
Fixes one part of #354

PR #371, opened 2018-10-09
Use TransformerContext in CacheKey computations

jad-hamza:
Following discussions with @samarion, this pull request changes a bit the caches API

* The `ExtractionCache`'s used by `CachingPhase` now use the `TransformerContext` instead of `s.Symbols` for `CacheKey` computations
* Replace `DependencyKey` by a more generic `SetKey`
* Fix some cache issues (thanks @romac for the report of the cache issue in `Laws`)
* Fix minor bug in `Trees` (thanks @samarion)

Edit: As we discussed, `MethodLifting` needs to be adapted to the new cache API and key computation must be done in the `TransformerContext`, but I leave this for another time

### DIFF
--- a/core/src/main/scala/stainless/extraction/ExtractionCaches.scala
+++ b/core/src/main/scala/stainless/extraction/ExtractionCaches.scala
@@ -50,9 +50,6 @@ trait ExtractionCaches { self: ExtractionContext =>
     def apply(fd: s.FunDef): CacheKey = new FunctionKey(fd)
   }
 
-  final def funKeys(fds: Set[s.FunDef]): CacheKey =
-    SetKey(fds.map(FunctionKey(_)))
-
   private final class SortKey private(private val sort: s.ADTSort) extends CacheKey {
     override def dependencies = Set(sort.id)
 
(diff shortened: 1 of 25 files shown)


romac:
This is superseded by #370 right? If so we can close it I think

core/src/main/scala/stainless/extraction/ExtractionCaches.scala
```diff
@@ -77,22 +75,30 @@ trait ExtractionCaches { self: ExtractionPipeline =>
...
+  protected def getSimpleKey(id: Identifier)(implicit symbols: s.Symbols): CacheKey =
+    symbols.lookupFunction(id).map(FunctionKey(_))
+      .orElse(symbols.lookupSort(id).map(SortKey(_)))
       .getOrElse(throw new RuntimeException(
         "Couldn't find symbol " + id.asString + " in symbols\n\n" + symbols.asString))
 
+  /** Returns a [[CacheKey]] given some identifier and the symbols from which
+    * it was taken (uses `symbols.dependencies` to compute the set of dependencies). */
+  protected def getDependencyKey(id: Identifier)(implicit symbols: s.Symbols): CacheKey =
+    SetKey((symbols.dependencies(id) + id).map(getSimpleKey)) + ValueKey(FreshIdentifier(""))
```
jad-hamza: Build now goes through when I completely disable `getDependencyKey` by adding a fresh `ValueKey`. There's something wrong but not sure what yet.

jad-hamza: Fixed, nevermind! The problem with that code is that two different functions can have the same dependency key.

PR #373, opened 2018-10-15
Fix soundness issue in MethodLifting

romac:
In a method's dispatcher, we now constrain forwarded arguments to
have the expected type, rather than just casting them to said type.

Not doing leads in some cases to invalid models where some argument's
model is of a totally unrelated type.

It is sound to use `assume` for this, as this is a constraint that
is guaranteed by both Scala's and our type system.

### DIFF
--- a/core/src/main/scala/stainless/extraction/methods/MethodLifting.scala
+++ b/core/src/main/scala/stainless/extraction/methods/MethodLifting.scala
@@ -12,7 +12,7 @@ trait MethodLifting extends oo.ExtractionContext with oo.ExtractionCaches { self
 
   private[this] def isAccessor(fd: FunDef): Boolean =
     fd.flags exists { case IsAccessor(_) => true case _ => false }
-    
+
   override protected final type TransformerContext = Symbols
   override protected final def getContext(symbols: s.Symbols) = symbols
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/stainless/extraction/methods/MethodLifting.scala
```diff
@@ -228,12 +228,15 @@ trait MethodLifting extends oo.ExtractionContext with oo.ExtractionCaches { self
         def wrap(e: t.Expr, tpe: s.Type, expected: s.Type): t.Expr =
           if (symbols.isSubtypeOf(tpe, expected)) e else (e match {
             case v: t.Variable =>
-              unchecked(t.AsInstanceOf(e, transformer.transform(expected.getType(symbols))).copiedFrom(e))
+              val expectedType = transformer.transform(expected.getType(symbols))
+              t.Assume(t.IsInstanceOf(e, expectedType),
```
samarion: Don't forget the `copiedFrom` for the type check.

core/src/main/scala/stainless/extraction/methods/MethodLifting.scala
```diff
@@ -228,12 +228,15 @@ trait MethodLifting extends oo.ExtractionContext with oo.ExtractionCaches { self
...
+                unchecked(t.AsInstanceOf(e, expectedType).copiedFrom(e))).copiedFrom(e)
             case _ =>
               val vd = t.ValDef.fresh("x", transformer.transform(tpe), true).copiedFrom(e)
-              t.Let(vd, e, unchecked(
-                t.AsInstanceOf(vd.toVariable, transformer.transform(expected.getType(symbols))).copiedFrom(e)
-              ))
+              val expectedType = transformer.transform(expected.getType(symbols))
+              t.Let(vd, e,
+                t.Assume(t.IsInstanceOf(vd.toVariable, expectedType),
```
samarion: Same.

PR #377, opened 2018-10-17
Freshen local variables in refinement lifting

jad-hamza:
I had a problem before doing this change, but unfortunately I cannot reproduce the issue I got. If we don't freshen, I think this might create conflicts when building a `Path`.


samarion:
Merged

PR #383, opened 2018-10-24
Simplify verification conditions before caching and displaying

jad-hamza:
Does `simplifyExpr` subsume `simplifyLets`?


samarion:
No, it doesn't quite subsume it. The simplifications in `simplifyLets` are not really useful for solving, only for displaying the VC. Also, `simplifyExpr` doesn't subsume `simplifyAssertions` either (but it's again only useful for display). I think we should call both of these simplifiers on the VC, either before or after `simplifyExpr`, it doesn't really matter.

I think it would be more efficient to run the simplification in `checkVCs` or `verify`, otherwise you call `simplifyExpr` twice when using the verification cache (and it can be fairly slow).

PR #388, opened 2018-10-25
Keep ghost/library flags when overriding methods

jad-hamza:
This enables inheritance of the ghost flag (and library) when overriding accessors and when creating fields for accessors in the `Sealing` phase.

PS: We don't need the `Final` flag here right?

### DIFF
--- a/core/src/main/scala/stainless/extraction/methods/Sealing.scala
+++ b/core/src/main/scala/stainless/extraction/methods/Sealing.scala
@@ -130,17 +130,24 @@ trait Sealing extends oo.CachingPhase
           }.toMap
       }.get)
 
-      // These are the flags that we keep when overriding an accessor and creating a field
-      def overrideAccessorKeepFlags(flag: Flag) = flag match {
+      // These are the flags that we *discard* when overriding an accessor or a methodand creating a field
+      def overrideDiscardFlag(flag: Flag) = flag match {
+        case IsAbstract => true
+        case IsAccessor(_) => true
+        case IsMethodOf(_) => true
+        case _ => false
+      }
+
+      // These are the flags that we *keep* when creating a field for an accessor
+      def fieldKeepFlag(flag: Flag) = flag match {
         case Ghost => true
-        case Annotation("library", Seq()) => true
         case _ => false
       }
 
       val fields = fieldAccessors.map { fd =>
         val id = FreshIdentifier(fieldName(fd))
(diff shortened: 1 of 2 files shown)


core/src/main/scala/stainless/extraction/methods/Sealing.scala
```diff
@@ -162,24 +170,27 @@ trait Sealing extends oo.CachingPhase
         val instantiator = getInstantiator(fd)
 
         val ct = ClassType(dummyClass.id, cd.typeArgs)
+        val newFlags = fd.flags.filter(overrideAccessorKeepFlags) ++
```
samarion: This doesn't seem right. Shouldn't we keep all flags except for `IsMethodOf`, `IsAccessor`, and `IsAbstract`? I'm afraid that listing the flags we _do_ keep is a bit brittle.

core/src/main/scala/stainless/extraction/methods/Sealing.scala
```diff
@@ -162,24 +170,27 @@ trait Sealing extends oo.CachingPhase
...
-            flags = Seq(Synthetic, IsMethodOf(ct.id), Derived(id), Final, IsAccessor(Some(vd.id)))
+            flags = newFlags
           )
         }))
       }
 
       // These are the flags that we keep when overriding a method
       def overrideKeepFlags(flag: Flag) = flag match {
         case IsPure => true
+        case Ghost => true
```
samarion: Again, I'd prefer keeping all flags except some specific ones we know shouldn't be kept.

core/src/main/scala/stainless/extraction/methods/Sealing.scala
```diff
@@ -130,13 +130,21 @@ trait Sealing extends oo.CachingPhase
...
+      def overrideAccessorKeepFlags(flag: Flag) = flag match {
+        case Ghost => true
+        case Annotation("library", Seq()) => true
+        case _ => false
+      }
+
       val fields = fieldAccessors.map { fd =>
         val id = FreshIdentifier(fieldName(fd))
         val instantiator = getInstantiator(fd)
+        val flags = fd.flags.filter(overrideAccessorKeepFlags)
```
samarion: Here I agree that listing out the flags we do keep makes sense, as `ValDef` and `FunDef` flags are usually pretty different. I don't think we need to keep the `@library` annotation flag for fields though.

PR #394, opened 2018-10-29
Remove extern on accessors and drop ignored accessors

romac:


romac:
We should probably address #395 before merging this.

romac:
Closing in favour of an upcoming PR, after discussion with @jad-hamza and @samarion.

PR #396, opened 2018-10-29
Add packaging script for stainless-scalac-standalone

gsps:
This PR adds a packaging script that can be run sporadically for standalone "releases" which we could distribute separately from any Scala- and Java-related package managers. The script can be invoked from the stainless root via `scripts/package-standalone.sh`. It will then

1. assemble the `stainless-scalac-standalone` sbt subproject into a fat jar
2. download Z3 binaries (for a version specified in the script)
3. unpack the binaries and pick out the files needed to run stainless with the `z3native` or the `smt-z3` solver
4. and repackage all of the aforementioned files along with a launcher script as an archive.

Note that this process is done for both Linux and OSX binaries. Currently I probably package a few additional dynamic libraries that are not strictly needed.

@samarion @romac Could I ask you to test these steps as well? Simply try to run the script, you'll then get an archive for your platform containing a self-contained stainless installation. E.g.
```
cd stainless/
scripts/package-standalone.sh
unzip stainless-scalac-standalone-0.1.0-$PLATFORM.zip -d standalone-test
cd standalone-test/
./stainless --solvers=nativez3 ../frontends/benchmarks/verification/valid/Trees1.scala
```
where `PLATFORM` is to be replaced by `linux` or `osx`. The last command should run stainless and end up using `z3native`.

Thanks!


gsps:
I got rid of the dynamic libaries that come with the z3 release, since Nicolas pointed out to me that scalaz3 comes with those anyways. I also reverted to assembling the fat jar without scalaz3, since we might as well copy the platform-specific version thereof and keep the stainless jar platform-independent. Would be great if you could give it a try now @samarion @romac

bin/package-standalone.sh
```diff
@@ -0,0 +1,137 @@
...
+}
+function fail {
+  info "${RED}    Failed. See log ($LOG) for details."
+  exit 1
+}
+
+# -----
+
+# TMP_DIR=`mktemp -d 2>/dev/null || mktemp -d -t 'stainless-package-standalone'`
+TMP_DIR="./package-tmp"; mkdir $TMP_DIR 2>/dev/null || true
```
samarion: Why is the temporary directory option commented out? Since we delete the temporary directory at the end anyway, might as well use a clean path.

romac:
Closing in favour of clean PR at #487

PR #398, opened 2018-11-02
Minor changes in EffectsChecker error messages and EffectsAnalysis

jad-hamza:
- Use `asString` when displaying effects (one function per line)
- Display the effect at fault when checking ghost code


romac:
Merged into #399.

PR #402, opened 2018-11-21
Add syntax highlighting to output

romac:
Currently uses a modified, standalone version of the syntax highlighter from the Dotty version we depend against.

- [ ] Add Dotty copyright and check licensing


vkuncak:
This appears to be based on escape codes for ANSI terminals?
It might be good if the results also work in case we want to adapt the output to work in a future HTML enabled console or a REPL that runs over web (have just two different formatters).

PR #403, opened 2018-11-22
 Use @extern instead of @ignore for fields whose type we cannot extract

romac:


romac:
Actually already superseded by #397. Closing.

PR #405, opened 2018-11-23
Add option to consolidate all symbols into a single symbol table

romac:
Just a quick and dirty hack for now. Not tested with `--watch` mode.


romac:
Superseded by #439

Issue #406, opened 2018-11-27
Adding @inline makes a verification condition fail


jad-hamza:
```scala
import stainless.annotation._

object Pack {
  def unpack[A](proof: () => Boolean): Unit = {
    require(!proof() || false)
    ()
  }

  @inline // adding @inline makes the precondition VC of the call to unpack fail (in f)
  def pack[A](proof: => Boolean): () => Boolean = {
    require(proof)
    () => proof
  }

  def axiom(): Boolean = {
    ??? : Boolean
  } ensuring(_ => false)

  def f(x: BigInt, y: BigInt) = {
    unpack(pack(axiom()))
  }
}
```

Output with `@inline`:

```
[  Info  ]  - Now solving 'precond. (call unpack[{ x: Object | @unchecked false }] ...)' VC for f @20:5...
[ Debug  ] ¬{
[ Debug  ]   assert(axiom, "Inlined precondition of pack")
[ Debug  ]   () => true
[ Debug  ] }()
[ Debug  ] Solving with: nativez3
[  Info  ]  - Result for 'precond. (call unpack[{ x: Object | @unchecked false }] ...)' VC for f @20:5:
[Warning ]  => INVALID
[Warning ] Found counter-example:
[Warning ]   (Empty model)
[  Info  ]  - Checking cache: 'body assertion: Inlined precondition of pack' VC for f @20:12...
[  Info  ] Cache hit: 'body assertion: Inlined precondition of pack' VC for f @20:12...
[  Info  ]   ┌───────────────────┐
[  Info  ] ╔═╡ stainless summary ╞═════════════════════════════════════════════════════════════════════════════════════════════════════════════════════╗
[  Info  ] ║ └───────────────────┘                                                                                                                     ║
[  Info  ] ║ f  precond. (call unpack[{ x: Object | @unchecked false }] ...)  invalid           nativez3  NoEquationPropagationTest.scala:20:5   3,862 ║
[  Info  ] ║ f  body assertion: Inlined precondition of pack                  valid from cache            NoEquationPropagationTest.scala:20:12  0,002 ║
[...]

Issue #404, opened 2018-11-23
Post-condition of @extern function is ignored


jad-hamza:
```scala
import stainless.annotation._

object DynamicAssert {
  @library // @extern // adding extern here makes the assert(x == 0) fail
  def dynamicAssert(b: Boolean): Unit = {
    ??? : Unit
  } ensuring(_ => b)

  def ff(x: BigInt) = {
    dynamicAssert(x == 0)
    assert(x == 0)
  }
}
```

Verification succeeds without `@extern` but fails with it. Perhaps an Inox issue?

There is a small difference in the tip files that are generated, but Inox reports "unsat" on both when run from the command-line. Z3, however, reports `sat` on the smt2 file that is generated for the case with `@extern` (with the model `x = 1`).

Edit: @samarion You're right Inox reports "sat" with `--assume-checked ` for the case with `@extern`

With `@extern`:
```smt
(declare-datatypes () ((Unit!1 (Unit!2))))
(define-fun dynamicAssert!0 ((b!6 Bool)) Unit!1 (choose x$1!10 Unit!1 b!6))
(declare-const x!86 Int)
(assert (not (let ((tmp!0 (dynamicAssert!0 (= x!86 0)))) (= x!86 0))))
(check-sat)
```

Without `@extern`:
```smt
(declare-datatypes () ((Unit!1 (Unit!2))))
(define-fun dynamicAssert!0 ((b!6 Bool)) Unit!1 (let ((x$1!10 (choose res!58 Unit!1 b!6))) (assume b!6 x$1!10)))
(declare-const x!86 Int)
(assert (not (let ((tmp!0 (dynamicAssert!0 (= x!86 0)))) (= x!86 0))))
(check-sat)
```

romac:
Fixed by #408

PR #408, opened 2018-11-28
Remove all assertions, and use assumeChecked for simplifications

jad-hamza:
Fixes https://github.com/epfl-lara/stainless/issues/406

### DIFF
--- a/core/src/main/scala/stainless/transformers/SimplifierWithPC.scala
+++ b/core/src/main/scala/stainless/transformers/SimplifierWithPC.scala
@@ -32,7 +32,7 @@ trait SimplifierWithPC extends Transformer with inox.transformers.SimplifierWith
       case (BooleanLiteral(true), true) => simplify(body, path)
       case (rp, _) =>
         val (rb, _) = simplify(body, path)
-        (Ensuring(rb, Lambda(Seq(res), rp).copiedFrom(l)).copiedFrom(e), opts.assumeChecked)
+        (Ensuring(rb, Lambda(Seq(res), rp).copiedFrom(l)).copiedFrom(e), false)
     }
 
     case MatchExpr(scrut, cases) =>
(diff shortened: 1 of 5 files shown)


samarion:
The `SimplifierWithPC` in Stainless considers postconditions to be pure, which it shouldn't. The `Ensuring` case should return `false` instead of `opts.assumeChecked` here:
https://github.com/epfl-lara/stainless/blob/93dbd338e1814994aa273da62f33a7f212c7610e/core/src/main/scala/stainless/transformers/SimplifierWithPC.scala#L35

jad-hamza:
Thanks! That fixed `IntSet.scala`. Are the other `opts.assumeChecked` in `SimplifierWithPC` fine?

samarion:
Yes I think so.

jad-hamza:
(Also fixes https://github.com/epfl-lara/stainless/issues/404)

core/src/main/scala/stainless/transformers/SimplifierWithPC.scala
```diff
@@ -32,7 +32,7 @@ trait SimplifierWithPC extends Transformer with inox.transformers.SimplifierWith
       case (BooleanLiteral(true), true) => simplify(body, path)
       case (rp, _) =>
         val (rb, _) = simplify(body, path)
-        (Ensuring(rb, Lambda(Seq(res), rp).copiedFrom(l)).copiedFrom(e), opts.assumeChecked)
+        (Ensuring(rb, Lambda(Seq(res), rp).copiedFrom(l)).copiedFrom(e), false)
```
romac: Shouldn't this rather be the following? ie. Can't we deduce the actual purity of the `Ensuring` by checking whether both its body and predicate are pure?

```scala
      case (rp, ps) =>
	        val (rb, bs) = simplify(body, path)
	        (Ensuring(rb, Lambda(Seq(res), rp).copiedFrom(l)).copiedFrom(e), opts.assumeChecked)
	        (Ensuring(rb, Lambda(Seq(res), rp).copiedFrom(l)).copiedFrom(e), ps && bs)
```

jad-hamza: Perhaps we could distinguish between purity for state mutation and purity for the effect given by an ensuring. Here I think we wanted to always consider an ensuring to be impure (even when there is no state mutation) to avoid having it simplified away by the simplifier, and thus having a post-condition of a function call being thrown away (and ignored by the solver).

PR #410, opened 2018-11-29
Allow fields/methods named after tuple selectors

romac:
@manoskouk :)


romac:
Was already working with dotty, this is just a port of the corresponding code.

PR #271, opened 2018-06-15
Inner classes

romac:
## TODO

- [x] Handle class selectors (scalac/dotc)
- [x] Handle field assignments (scalac/dotc)
- [x] Fix pretty printer
- [x] Fix bugs in `InnerClasses`
- [x] Allow defining classes within local functions (done but fails in `AntiAliasing`)
- [x] Fix `typeBounds` for `LocalClassType`
- [x] Fix caching bug (in `FieldAccessors` and `MethodLifting`)
- [x] Add caching to `InnerClasses`
- [x] Properly reject unsupported programs
- [x] Always add reference to outer class
- [x] Dotty support (Weird registry errors)
- [x] Refactor
- [x] Write documentation

Fixes #36

PR #414, opened 2018-12-12
Help solver deduce that reified type args do not influence evaluation

romac:
If function has gained new parameters for reifed types during
TypeEncoding, we create a dummy version of it without any
type arguments as a witness that type arguments do not
influence the result of the computation of its invocation.

We do this by duplicating the encoded function into
another function without the reified type parameters
and an empty body.

The post of the encoded function is then modified to assert
that its result is equals to the result of this new function,
when applied to (non-type args) parameters.

This allows the solver to deduce that the reified type parameters
themselves do not influence the result of a call
to the original function, something that is needed eg.,
in the presence of GADTs where equivalent calls to the same method
via both the top-level dispatcher, and a concrete implementation
can carry different reified type parameters.

Thanks @samarion for helping me figuring this stuff out and coming up with this fix.

### DIFF
--- a/core/src/main/scala/stainless/extraction/oo/TypeEncoding.scala
+++ b/core/src/main/scala/stainless/extraction/oo/TypeEncoding.scala
@@ -1184,16 +1184,15 @@ trait TypeEncoding
       val elimParams = encoded.params
         .drop(reifiedTypeArgsCount)
         .map(vd => vd.copy(tpe = dropRefinements(vd.tpe)))
-        .map(_.freshen)
 
-      val eliminated = new t.FunDef(
+      val eliminated = t.exprOps.freshenSignature(new t.FunDef(
         original.id.freshen,
-        encoded.tparams.map(_.freshen),
+        encoded.tparams,
         elimParams,
         dropRefinements(encoded.returnType),
         t.NoTree(dropRefinements(encoded.returnType)),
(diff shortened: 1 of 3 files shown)


romac:
There is some kind of bug that crashes the termination checker on the `LawTypeArgsElim` benchmark:

```
[info] [Internal] Error: Type error: {
[info] [Internal]   x match {
[info] [Internal]     case StructureExt(__x) =>
[info] [Internal]       1 + integerAbs(__x)
[info] [Internal]     case OptionStructure(ev) =>
[info] [Internal]       1 + fullSize$Object(ev)
[info] [Internal]     case Option(value) =>
[info] [Internal]       1 + fullSize$Option[Object](value)
[info] [Internal]     case Open(value) =>
[info] [Internal]       1 + integerAbs(value)
[info] [Internal]   }
[info] [Internal] } ensuring {
[info] [Internal]   (res: BigInt) => res >= 0
[info] [Internal] }, expected BigInt,
[info] [Internal] found <untyped>
```

which stems from a missing synthetic size function:

```
[info] [Internal]       fullSize$Option[Object](value) is of type <untyped>
[info] [Internal]         value is of type Option[Object] but couldn't find function fullSize$Option
```

romac:
Merged into #271

PR #399, opened 2018-11-07
Conditional effect targets for accessors, mutable Any, and phase re-ordering

romac:
- [x] Move `imperative` phase before `oo`
- [x] Allow aliasing through accessors thanks to conditional effect targets
- [x] Treat `Any` as a mutable type
- [x] Fix #400 
- [x] Fix #409
- [x] Fix #425
- [x] Fix #395

PR #418, opened 2018-12-17
Add snapshot keyword for imperative code

jad-hamza:
Depends on https://github.com/epfl-lara/stainless/pull/399, thanks @romac for the tips.

The `Snapshot` tree can be used to save the state of a mutable object, so that we can refer to it at a later point of the code.


jad-hamza:
> Wow that’s much simpler than I expected. (Once you figure out what to put on the right hand side of the arrow obviously :p) Nice :)

Yes I was surprised as well! It seems you can also modify the object that was snapshot and it works out of the box (?). We should probably restrict the expressions that can be snapshot though. Can we allow for instance all pure expressions?

romac:
Rebased at #491 491

PR #423, opened 2018-12-19
Remove inline annotation on stainless.lang.ghost

### LINKED ISSUE #419: Extraction of inlined functions in the dotty frontend
```scala
import stainless.lang._
import stainless.annotation._

case class GhostInlineExtraction(@ghost var x: BigInt) {
  def f() = {
    ghost { x }
  }
}
```

The call to `ghost` is not extracted properly in the dotty frontend (due to some early inlining). I guess that's a known issue seeing that ghost benchmarks are ignored in the `DottyExtractionSuite`

```scala
[ Debug  ] Symbols before PartialFunctions
[ Debug  ] 
[ Debug  ] @method(GhostInlineExtraction)
[ Debug  ] def f: Unit = {
[ Debug  ]   val value: BigInt = this.x
[ Debug  ]   @unchecked ()
[ Debug  ] }
```

romac:
Fixes #419

### DIFF
--- a/frontends/library/stainless/lang/package.scala
+++ b/frontends/library/stainless/lang/package.scala
@@ -8,7 +8,7 @@ import scala.language.implicitConversions
 package object lang {
   import stainless.proof._
 
-  @inline @library
+  @library
   def ghost[A](@ghost value: A): Unit = ()
 
   @ignore


frontends/library/stainless/lang/package.scala
```diff
@@ -8,7 +8,7 @@ import scala.language.implicitConversions
 package object lang {
   import stainless.proof._
 
-  @inline @library
+  @library
   def ghost[A](@ghost value: A): Unit = ()
```
vkuncak: Do we use this? 
What difference does it make?
Are we sure we want `value: A` and not `value: =>A`  ?

romac: > What difference does it make if it is inlined or not?

If it is inlined, then the ghost sbt plugin phase will not see its invocation and thus not remove the corresponding trees.

> Also, are we sure we want value: `A` and not `value: =>A` ?

Right, when using the sbt plugin it does not change anything, as the parameter is replaced with null, but in users do not use it, it would indeed be better to make it by-name. Will change that!

PR #426, opened 2019-01-07
Add an optional parameter to @induct which allow specifying which parameter to perform the induction over

### LINKED ISSUE #420: Specify which arguments should the induction be performed over with @induct
This would let us write

```scala
  implicit def listMonoid[A]: Monoid[List[A]] = new Monoid[List[A]] {
    override def empty: List[A] = Nil[A]()
    override def append(xs: List[A], ys: List[A]): List[A] = xs ++ ys

    @induct(xs)
    override def law_associativity(xs: List[A], ys: List[A], zs: List[A]): Boolean = {
      true
    }
  }
```

instead of

```scala
  @induct
  def lemma_List_append_associativity[A](xs: List[A], ys: List[A], zs: List[A]): Boolean = {
    ((xs ++ ys) ++ zs) == (xs ++ (ys ++ zs))
  }.holds

  implicit def listMonoid[A]: Monoid[List[A]] = new Monoid[List[A]] {
    override def empty: List[A] = Nil[A]()
    override def append(xs: List[A], ys: List[A]): List[A] = xs ++ ys

    override def law_associativity(xs: List[A], ys: List[A], zs: List[A]): Boolean = {
      lemma_List_append_associativity(xs, ys, zs)
    }
  }
```

If we just specify `@induct` in the first example, the induction will be performed over `this` instead of `xs`, ie. the first parameter of the function after method lifting. This might also be a problem when we have reified type arguments but I haven't checked yet.

romac:
Close #420

### DIFF
--- a/core/src/main/scala/stainless/ast/Deconstructors.scala
+++ b/core/src/main/scala/stainless/ast/Deconstructors.scala
@@ -182,6 +182,7 @@ trait TreeDeconstructor extends inox.ast.TreeDeconstructor {
     case s.Final => (Seq(), Seq(), Seq(), (_, _, _) => t.Final)
     case s.Unchecked => (Seq(), Seq(), Seq(), (_, _, _) => t.Unchecked)
     case s.Synthetic => (Seq(), Seq(), Seq(), (_, _, _) => t.Synthetic)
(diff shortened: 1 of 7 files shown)


romac:
Superseded by #452

PR #428, opened 2019-01-10
Disallow holes where the type has been inferred to Nothing

### LINKED ISSUE #416: `???` is extracted as `???[Nothing]`
```scala
object TrueBoolean {
  def b: Boolean = ???
  def f = assert(b)
}
```

This is reported as valid because `???` is extracted as `???[Nothing]` instead of `???[Boolean]`. It was a bit confusing at first. Maybe we should completely prevent the use of `???` when the type is not specified?

romac:
Close #416

### DIFF
--- a/frontends/benchmarks/extraction/invalid/TypedHoleInferredNothing.scala
+++ b/frontends/benchmarks/extraction/invalid/TypedHoleInferredNothing.scala
@@ -1,0 +1,4 @@
+object TypedHoleInferredNothing {
+  def b: Boolean = ???
+  def f = assert(b)
+}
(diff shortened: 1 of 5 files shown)


vkuncak:
We may wish to prevent this, but I am now worried that this does not fix the issue at its root.
Is `???` an abbreviation for `choose` or something else?
I thought I saw that, when we use choose, a check is generated to find a witness for the type.
Why don't we generate an analogous check for `???` to ensure that the type is not empty?

romac:
@vkuncak Agreed, that seems like a better solution.

PR #434, opened 2019-01-17
Report positions when going through the compiler reporter from the sbt plugin

romac:
Result:

screenshot 2019-01-17 15 06 59
screenshot 2019-01-17 16 00 30

Some more work is needed on the Inox side to avoid having duplicated output in the SBT console.

Fixes #434


romac:
Blocked by epfl-lara/inox#91

PR #436, opened 2019-01-22
Upgrade Dotty to 0.12.0-RC1

romac:
## Supported

- [x] Extension Methods
- [x] Refinement types in functions parameters
- [x] Refinement types in constructor parameters
- [x] Erased Terms as ghost terms (partially working, still need to propagate annotations)
- [x] Opaque Types (they get dealiased during extraction, better support in #470)
- [x] Dependent Function Types (#470)

## Untested

- ADTs
- Enumerations
- Intersection Types
- Union Types
- Type Lambdas
- Match Types
- Typeclass Derivation
- Multiversal Equality
- Trait Parameters
- Inlining by Rewriting
- Meta Programming
- By-Name Implicits
- Auto Parameter Tupling
- Named Type Arguments

## Untested (coming in 0.13)

- Witnesses
- Top-level definitions

## Unsupported

- Kind Polymorphism
- TASTy Reflect

### DIFF
--- a/core/src/main/scala/stainless/extraction/oo/RefinementLifting.scala
+++ b/core/src/main/scala/stainless/extraction/oo/RefinementLifting.scala
@@ -63,15 +63,16 @@ trait RefinementLifting
     } (tpe)
 
     def dropRefinements(tpe: s.Type): s.Type = liftRefinements(tpe) match {
-      case s.RefinementType(vd, _) => vd.tpe
+      case s.RefinementType(vd, _) => dropRefinements(vd.tpe)
       case _ => tpe
     }
 
     def parameterConds(vds: Seq[s.ValDef]): (Seq[s.ValDef], s.Expr) = {
       val (newParams, conds) = vds.map(vd => liftRefinements(vd.tpe) match {
         case s.RefinementType(vd2, pred) =>
-          val nvd = vd.copy(tpe = vd2.tpe).copiedFrom(vd)
-          (nvd, s.exprOps.replaceFromSymbols(Map(vd2 -> nvd.toVariable), pred))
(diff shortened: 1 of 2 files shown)


romac:
Sneak peek at the WIP refinement types integration:

### Input:
screenshot 2019-01-25 17 46 03

### Output:
screenshot 2019-01-25 17 46 15

/cc @gsps @vkuncak @samarion @jad-hamza

vkuncak:
This is very, very cool, @romac !

jad-hamza:
Nice!

romac:
Somehow, the DottyExtraction suite fails when type-checking with errors such as `expected A, found A`, whereas feeding `stainless-dotty/run` all the files in eg. `verification/valid` succeeds.

vkuncak:
Is it not again the issue that stainless does not run all phases of the compiler by default?

romac:
I don’t think so. Both the test suite and the regular way of invoking stainless should run the same compiler phases. Might be a big due to building one single symbol table for all benchmarks, or something like that. Will continue to investigate further.

samarion:
I pushed https://github.com/epfl-lara/stainless/commit/1c89a595de7abbed43aa315226ac67a7085ddb21 to https://github.com/epfl-lara/stainless/commit/cd5518ed73edeca93f87f0cdea0fb795d6263f6b which re-enables the dotty frontend.

However, support for pi- and refinement-types, as well as type members is still not complete so we can leave this PR open until those features have been implemented.

romac:
Thanks for fixing the plugin :) I will get on the rest!

romac:
Π-types (dependent functions types) are now supported in #470, and I opened an issue (#477) to track the remaining work (ie. extract types by traversing type trees recursively).

PR #437, opened 2019-01-22
Bump Scala to v2.12.8

### LINKED ISSUE #326: Make `stainless-core` compatible with 2.12

romac:
Close #326

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -272,30 +272,30 @@ lazy val `stainless-scalac-standalone` = (project in file("frontends") / "stainl
 //    buildInfoPackage := "stainless"
 //  )
 
-//lazy val `sbt-stainless` = (project in file("sbt-plugin"))
-//  .enablePlugins(BuildInfoPlugin)
-//  .settings(baseSettings)
-//  .settings(publishSbtSettings)
-//  .settings(
-//    description := "Plugin integrating Stainless in sbt.",
-//    sbtPlugin := true,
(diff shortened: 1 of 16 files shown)


vkuncak:
I a little lost, was this not done already?

romac:
We first went from 2.11.1 to 2.11.9, and this PR goes from 2.11.9 to 2.12.8. So yeah it’s not the first PR that updates the Scala compiler version.

samarion:
Why are the dotty projects commented in `build.sbt`? Was this really ready to be merged?

romac:
My bad, I must have forgotten to uncomment them as I then moved on to work on the ‘latest-dotty’ branch. I can re-enable them and open a new PR.

romac:
Oh I remember why I left them commented out: there is no published version of Dotty for 2.12, the only available non-bootstrapped version are for 2.11 (version 0.1 of Dotty). So we need to publish our own non-bootstrapped version to our Bintray before we can re-enable Dotty on 2.12. I can either prepare a branch with only the changes pertaining to upgrading Dotty (without refinement types, erased terms, etc.) or else we need wait until #436 is fully ready.

romac:
In any case, I should have left the `work in progress` label on. My bad, sorry about that.

vkuncak:
I actually take the blame for this one. There's a reason we do pull requests and reviews and I was unable to spot something that was apparently rather basic. I will try to be even more conservative when doing merges, if I do them at all.

PR #440, opened 2019-01-24
Custom front-end based on the Inox parser and elaborator

OStevan:
Flexible front end prototype based on Inox type inference algorithm.


vkuncak:
If we do this, we will likely take a very different path than this particular implementation

PR #442, opened 2019-02-14
Update README with links to HTML documentation

romac:


vkuncak:
Pointing to github.io seems like an improvement

vkuncak:
LGTM

PR #451, opened 2019-03-02
Continue processing the remaining symbols after an extraction error

romac:


romac:
Will integrate this in #422.

PR #452, opened 2019-03-03
 Add TypeChecker component to generate verification conditions 

jad-hamza:
This commit adds an option `--type-checker` to generate verification conditions using type-checking rules. There are three debug options:

- `type-checker` to see the state of the type-checker in the command-line during type-checking
- `type-checker-vcs` to see the VCs
- `derivation` to get HTML files containing the derivation trees, one per function and one per ADT (beware if you have HTML files in your folder as this option will overwrite them if they have the same name)

I added a copy of the `verification/valid` benchmarks in `typechecker/valid` with some modifications to have them go through, and with things that do not work yet commented out. The `TypeCheckerVerificationSuite` runs these examples, and additionally there is a `TypeCheckerInternalSuite` with manually written stream examples.

Edit: TODO list moved to https://github.com/epfl-lara/stainless/issues/573


jad-hamza:
@romac Any idea why this example is failing? (also on master)

 ```scala
import scala.collection.concurrent.TrieMap

object IgnoredField {
  case class TrieMapWrapper[K, V](theMap: TrieMap[K, V])
}
```

```
[ Error  ] theMap$4 depends on missing dependencies: TrieMap$1.
[ Error  ] TrieMapWrapper$1 depends on missing dependencies: TrieMap$1.
[ Fatal  ] Cannot recover from missing dependencies
[ Error  ] Cannot recover from missing dependencies
```

romac:
I think `theMap` should be marked `@extern` for recovery to kick in.

jad-hamza:
Thanks!

romac:
This seems to rebase cleanly on top of latest master \o/

jad-hamza:
Closed in favor of https://github.com/epfl-lara/stainless/pull/479

PR #466, opened 2019-03-15
Support for configuration file

### LINKED ISSUE #465: Support for configuration file
It would be nice to be able to configure which flags to pass to Stainless via a configuration file, as it would allow

a) to not have to always repeat the same flags over and over across incantations
b) let users configure how to run Stainless via the sbt plugin without having to mess with the build.sbt (something we don't even support yet afaik)

romac:
The file must be named `stainless.conf` or `.stainless.conf` and be a valid HOCON file.

It will be searched for recursively starting from the current
directory and walking up the directory hierarchy.

So, if one runs stainless from `/a/b/c`, and there is a config
file in any of `c`, `b` or `a`, the first of those is going to be
loaded.

For example, the config file could contain the following:

```
vc-cache = false
debug = [verification, trees]
timeout = 5
check-models = true
print-ids = true
```

This will translate to the following command line options:

```
--vc-cache=false --debug=verification,trees --timeout=5 --print-ids
```

Closes #465

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 # Stainless
 /bin/stainless-*
 .stainless-cache/
+.stainless.conf
+stainless.conf
 
 # Vim
 *.swp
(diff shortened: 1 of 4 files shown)


romac:
Blocked by https://github.com/epfl-lara/inox/pull/95.

PR #470, opened 2019-04-02
Type aliases, type members, and dependent function types

romac:
- [x] Type aliases (ie. top-level type members)
- [x] Type members
- [x]  Dependent function types (Dotty only)

This PR builds on top of #422.


vkuncak:
frontends/benchmarks/extraction/valid/Typedef.scala seems identical to Typedefs1.scala in the same directory

vkuncak:
What is the main reason for failing extraction tests on this branch?

PR #476, opened 2019-04-17
Keep annotation and option to avoid filtering out library symbols

jad-hamza:
Implement @romac's idea for keeping symbols 
* At the moment writing `keep` forces the use of `BatchedCallBack`
* In the `BatchedCallBack`, we filter all library symbols except those marked by `@keep(g)` for some `g` given in the command-line option `--keep=g1,g2,...`

Blocked by https://github.com/epfl-lara/stainless/pull/422


romac:
Closing for rebased PR at #488.

PR #479, opened 2019-04-19
Indexed recursive types

jad-hamza:
Blocked by https://github.com/epfl-lara/stainless/pull/452

This adds some minimal frontend support for indexed recursive types (see file typechecker/valid/Streams.scala). At the moment the test is ignored, because the test suites use `scalac` but this code uses the dotty frontend (the example works when using the command-line).

The phase `SizedADTExtraction` transforms the annotations into `RecursiveType` (for types) and `SizedADT` (for terms).

Next step (following @gsps' suggestion) is to change the dotty parser to parse `Stream[T](n)` as an annotated type `Stream[T] @indexedAt(n)`, and add an annotation to avoid writing `indexedAt` in terms.


vkuncak:
Can someone explain what the path merging here is?
"At the moment the test is ignored" does not sound like the state we want to be in.
I guess we can have dotty specific test and scalac tests?

jad-hamza:
> I guess we can have dotty specific test and scalac tests?

We can follow the same pattern as the tests for Z3 and CVC4, i.e. have all tests be run by default and then exclude some as in https://github.com/epfl-lara/stainless/blob/9c867aed0a6bad468f5ab5f95b6c0c57eada6c7e/frontends/scalac/src/it/scala/stainless/verification/VerificationSuite.scala#L97-L100

But I'm not sure what's the right way to set this up. At the moment, all verification suites are in the `frontends/scalac` folder.

Would you like me to have a look at the parsing on Dotty?

PR #498, opened 2019-05-19
Documentation: updated installation; reordered pages

vkuncak:
The installation page is changed: remove Eclipse instructions, gave sbt instructions that directly invoke sbt-launch.jar, the order of subsections changed and java as dependency factored out.

Reorganized the order of documentation pages.


romac:
What's the rationale behind directly invoking sbt-launch.jar rather than installing sbt using a package manager or via https://github.com/paulp/sbt-extras/?

vkuncak:
The instructions are for the first run to build from sources. None of the instructions requires privilege to run a package manager. Running sbt by default depends on the state of your cache, which the option provided avoids. (default sbt run has actually failed with cryptic messages for me when I ran it on laraquad3). Invoking a jar file is pretty much the same thing that we instruct when we invoke the release jar file and it only depends on java.

romac:
Makes sense! Thanks for adding the explanation to the doc!

vkuncak:
I pushed this one to master directly. Will do better next time.

PR #504, opened 2019-05-21
Allow specifying which arguments should the induction be performed over with @induct

### LINKED ISSUE #420: Specify which arguments should the induction be performed over with @induct

romac:
This PR from @jad-hamza lets us write

```scala
  implicit def listMonoid[A]: Monoid[List[A]] = new Monoid[List[A]] {
    override def empty: List[A] = Nil[A]()
    override def append(xs: List[A], ys: List[A]): List[A] = xs ++ ys

    override def law_associativity(@induct xs: List[A], ys: List[A], zs: List[A]): Boolean = {
      true
    }
  }
```

instead of

```scala
  @induct
  def lemma_List_append_associativity[A](xs: List[A], ys: List[A], zs: List[A]): Boolean = {
    ((xs ++ ys) ++ zs) == (xs ++ (ys ++ zs))
  }.holds

  implicit def listMonoid[A]: Monoid[List[A]] = new Monoid[List[A]] {
    override def empty: List[A] = Nil[A]()
    override def append(xs: List[A], ys: List[A]): List[A] = xs ++ ys

    override def law_associativity(xs: List[A], ys: List[A], zs: List[A]): Boolean = {
      lemma_List_append_associativity(xs, ys, zs)
    }
  }
```

If we just specify `@induct` in the first example, the induction will be performed over `this` instead of `xs`, ie. the first parameter of the function after method lifting. Of course this also lets us induct over any other (single) parameter of the function if wanted.

Closes #420


romac:
@jad-hamza Are all commits here or do you think I missed some?

jad-hamza:
There are some `decrease`s that are inserted with types not supported by the current master (lists and other adts). We can either remove those or add the `SizeInjection` phase which is in the `extraction/termination` folder.

romac:
Yeah I just saw that, last commit should add the required phases, thanks!

jad-hamza:
Thanks for the new PR by the way!
At the moment I think the tests are failing due to some unnecessary inductions that need to be removed.

Issue #413, opened 2018-12-12
Variable of mutable type in trait


jad-hamza:
cc: @RomainJufer 

```scala
case class Counter(var x: BigInt)
trait T { var c: Counter }
```

Should this be supported? I've tried on @romac branch (https://github.com/epfl-lara/stainless/pull/399) but there is an aliasing issue:

```scala
[ Debug  ] Symbols before AntiAliasing
[ Debug  ] 
[ Debug  ] -------------Functions-------------
[ Debug  ] @accessor
[ Debug  ] @derived(c_=)
[ Debug  ] @accessor
[ Debug  ] def c_=(thiss: T, x$1: Counter): Unit = @unchecked thiss.asInstanceOf[TExt].c = x$1
[ Debug  ] 
[ Debug  ] @accessor
[ Debug  ] @derived(c)
[ Debug  ] @accessor
[ Debug  ] def c(thiss: T): Counter = @unchecked thiss.asInstanceOf[TExt].c
[ Debug  ] 
[ Debug  ] 
[ Fatal  ] File.scala:2:7: Illegal aliasing: x$1
           trait T { var c: Counter }
```

PS: We have no use-case for this, it's ok if this is disabled

jad-hamza:
@romac Which PR fixes this issue?

PR #526, opened 2019-06-03
No EffectsChecker in accessors

jad-hamza:
Fixes https://github.com/epfl-lara/stainless/issues/413 by removing aliasing checks from accessors. Does this workaround make sense?

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
@@ -266,7 +266,9 @@ trait EffectsChecker { self: EffectsAnalyzer =>
     }
 
     try {
-      check(Outer(fd))
+      // We only check the bodies of functions which are not accessors
+      if (!isAccessor(Outer(fd)))
+        check(Outer(fd))
       CheckResult.Ok
     } catch {
       case e: ImperativeEliminationException => CheckResult.Error(e)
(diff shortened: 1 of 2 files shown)


romac:
I think that makes sense since we only extract accessors synthesised by the compiler.

Can you please add the test case from #413 to the imperative suite? Then we can merge it I guess.

jad-hamza:
I've added `TraitVar3.scala', were `TraitVar1.scala` and `TraitVar2.scala' going through because they're sealed?

romac:
Now I am not quite sure whether that's the right solution. Should the analysis depend on whether a class is sealed or not in your opinion?

Perhaps it is better that we discuss this tomorrow in detail, and see from there?

jad-hamza:
Actually the difference was not about `sealed`, `TraitVar1` and `TraitVar2` have `var`'s on non-mutable type, while the new example in a `var` on a mutable type.

romac:
Oh right, then it makes sense!

samarion:
Does the aliasing tracker correctly track stuff through accessors now?

It seems like this creates a bug:
```scala
import stainless.lang._

case class A(var x: BigInt)
trait B { var a: A }

object Test {
  def test(b: B): Boolean = {
    val a = b.a
    a.x += 1
    b.a.x += 1
    a == b.a
  }.holds
}
```

Issue #532, opened 2019-06-06
Tuples with members with post-conditions are simplified away (2)


jad-hamza:
This seems to be a different bug than #531 (not fixed by https://github.com/epfl-lara/inox/pull/99)

```scala
import stainless.annotation._

object LastDynAssert {
  @extern
  def dynAssert(cond: Boolean): Unit = {
    (??? : Unit)
  } ensuring(cond)

  def f() = {
    // dynAssert(false) // first case: goes through
    (dynAssert(false), 0) // second case: doesn't go through
    assert(false)
  }
}
```

```scala
[ Debug  ] Symbols before ImperativeCodeElimination
[ Debug  ] 
[ Debug  ] -------------Functions-------------
[ Debug  ] def f: Unit = {
[ Debug  ]   (dynAssert(false), 0)
[ Debug  ]   assert(false)
[ Debug  ]   ()
[ Debug  ] }
[ Debug  ] 
[ Debug  ] 
[ Debug  ] Symbols after ImperativeCodeElimination
[ Debug  ] 
[ Debug  ] -------------Functions-------------
[ Debug  ] def f: Unit = {
[ Debug  ]   assert(false)
[ Debug  ]   ()
[ Debug  ] }
```

jad-hamza:
I think this example requires https://github.com/epfl-lara/inox/pull/99.

PR #533, opened 2019-06-06
Prevent over-simplifications in ImperativeCodeElimination

jad-hamza:
(In conjunction with https://github.com/epfl-lara/inox/pull/99) Fixes https://github.com/epfl-lara/stainless/issues/532

I'm not sure what this code does, but it was throwing away the tuple https://github.com/epfl-lara/stainless/issues/532, and only keeping `FunctionInvocation`'s and `ApplyLetRec`'s. I replaced that part so that all expressions are kept, does that seem correct?


romac:
Yeah I guess the simplification was too eager, as it should have checked whether any sub-expressions in Unit position had a impact on the VC path. This change sounds good to me, as doing such a check sounds a bit overkill to me.

PR #541, opened 2019-06-14
Disallow aliasing in let's

jad-hamza:
Fixes https://github.com/epfl-lara/stainless/pull/526#issuecomment-500335792


samarion:
This looks too strict to me (although it's a valid fix).
The ideal fix would be to correctly compute the aliased fields that are returned by accessors. Didn't you start working on that at some point @romac?

Issue #546, opened 2019-06-19
Recursive opaque method fails verification


jad-hamza:
```scala
import stainless.annotation._
import stainless.collection._

object IndexOfInjective {
  @opaque
  def indexOfInjective[T](l: List[T], x1: T, x2: T): Unit = {
    require(l.contains(x1) && l.contains(x2) && x1 != x2)

    if (!l.isEmpty && l.head != x1 && l.head != x2)
      indexOfInjective(l.tail, x1, x2)

  } ensuring(_ => l.indexOf(x1) != l.indexOf(x2))
}
```

This goes through without the `@opaque annotation`. When checking the postcondition of the function, we have the call `indexOfInjective(l.tail, x1, x2)` in the context. This call is removed by the simplifier, which thinks that the function `indexOfInjective` is pure. Its body is:

```scala
require(l.contains(x1) && l.contains(x2) && x1 != x2)
choose((u: Unit) => l.indexOf(x1) != l.indexOf(x2)))
```

PR #547, opened 2019-06-19
Make purity of requires and assertions depend on their bodies

jad-hamza:
I needed to fix https://github.com/epfl-lara/stainless/issues/546, but I'm not sure if that's the right approach, we now completely ignore the `assumeChecked` here (except for one case).

### DIFF
--- a/core/src/main/scala/stainless/transformers/SimplifierWithPC.scala
+++ b/core/src/main/scala/stainless/transformers/SimplifierWithPC.scala
@@ -14,18 +14,21 @@ trait SimplifierWithPC extends Transformer with inox.transformers.SimplifierWith
     case Assert(pred, oerr, body) => simplify(pred, path) match {
       case (BooleanLiteral(true), true) => simplify(body, path)
       case (BooleanLiteral(false), true) =>
-        val (rb, _) = simplify(body, path)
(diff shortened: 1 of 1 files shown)


samarion:
Hmm, if we have `assumeChecked == false`, then we should always return `false` in the purity check, so it shouldn't be ignored here. Probably something like `opts.assumeChecked && p`.

Also, you probably want to reflect this in the `InoxEncoder` since it will drop the require/assertion even if it contains an impure expression.

jad-hamza:
> Hmm, if we have `assumeChecked == false`, then we should always return `false` in the purity check, so it shouldn't be ignored here. Probably something like `opts.assumeChecked && p`.

Thanks, I fixed it. Can you describe a bit how Stainless/Inox interprets `assumeChecked`?
 
> Also, you probably want to reflect this in the `InoxEncoder` since it will drop the require/assertion even if it contains an impure expression.

In `InoxEncoder`, we drop the things being asserted and required, but not the bodies, so that's fine no?

samarion:
`assumeChecked` means that the expression will be verified by the system and can therefore be assumed to be pure. In practice, we have to be a bit careful about where we use this `assumeChecked` flag since things probably shouldn't be assumed checked before getting to Inox (or at least after type checking).

You're right about the `InoxEncoder`, it is actually a wanted feature that we ignore the purity of the asserted expressions. Just ignore that comment.

Issue #549, opened 2019-06-20
Ensuring of extern function gets silently dropped during extraction


jad-hamza:
```scala
import stainless.annotation._

object EnsuringExtraction {
  @extern def f(x: BigInt, y: BigInt): Unit = {
    require(x >= 0)
    (??? : Unit).ensuring(_ => y >= 0)
  }
}
```

This is extracted as:

```scala
@extern
def f(x: BigInt, y: BigInt): Unit = {
  require(x >= 0)
   <empty tree>[Unit]
}
```

Extraction works fine when `ensuring` is outside or when we remove the `@extern` annotation.

jad-hamza:
The problem here is that `ensuring` is part of the body of the extern function `f`, and is therefore dropped as any other expression. I'm not sure what's the best way to report this. We could emit a warning when we drop a `require/ensuring/assert` (which are a bit Stainless keywords) from an `@extern` function. What do you think?

romac:
Closed via #549

PR #562, opened 2019-06-26
Emit warning when dropping the non-empty body of a user @extern function

jad-hamza:
Run on the example from https://github.com/epfl-lara/stainless/issues/549, this will print:

```
[Warning ] Dropped non-empty body of @extern function g.
```

Would that even help the user figuring out what's happening? Also, I'm afraid this warning will get annoying if you actually want to have non-empty `@extern` functions.

### DIFF
--- a/frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
+++ b/frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
@@ -589,7 +589,24 @@ class CodeExtraction(inoxCtx: inox.Context, cache: SymbolsContext)(implicit val 
       }
     }
 
+    object KeywordChecker extends xt.SelfTreeTraverser {
+      override def traverse(e: xt.Expr) = {
+        e match {
+          case _: xt.Require =>
+            reporter.warning(e.getPos, s"This require is ignored for verification because it is not at the top-level of this @extern function.")
+          case _: xt.Ensuring =>
+            reporter.warning(e.getPos, s"This ensuring is ignored for verification because it is not at the top-level of this @extern function.")
+          case _: xt.Assert =>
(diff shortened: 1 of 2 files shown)


jad-hamza:
The previous build failed because I was trying to initialize some printer options from the context in `CodeExtraction`, while `ctx` is initialized after `StainlessExtraction` in:

https://github.com/epfl-lara/stainless/blob/7d205b5c24a52a3f27554d1ee9c079a5a2cb596d/frontends/scalac/src/main/scala/stainless/frontends/scalac/StainlessPlugin.scala#L42-L65

What's the syntax to initialize it before?

romac:
In most cases, extern functions are supposed to have a body, as they are often used for interop with non-PureScala code, so I think having the warning could get a bit annoying. We could hide it behind a debug section or something but perhaps there is a better way to report such issues (ie. when a we lose an ensuring clause because it is not at the top-level). We could perhaps also fix extraction to actually properly extract specs from such code.

romac:
Looks good!

frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
```diff
@@ -589,6 +589,21 @@ class CodeExtraction(inoxCtx: inox.Context, cache: SymbolsContext)(implicit val
       }
     }
 
+    object KeywordChecker extends xt.SelfTreeTraverser {
+      override def traverse(e: xt.Expr) = {
+        e match {
+          case _: xt.Require => reporter.warning(s"Require's inside the body of @extern functions are ignored for verification ($id).")
```
romac: - Report tree position
- Change message to "This require is ignored for verification because it not at the top-level of this @extern function"

PR #566, opened 2019-06-27
Decrease memory limit for sbt

jad-hamza:


romac:
I don’t this is actually needed now that we removed the memory limit on larabot.

PR #570, opened 2019-06-27
Add FAQ extracted from the C4DT newsletter

romac:


romac:
@vkuncak Does this look good to you?

vkuncak:
Thanks! I think we can drop Stainless-OmniLedger integration part. I had nothing to do with it and I think it's better to talk about that on OmniLedger related sites or in /smart/ version of stainless.

PR #595, opened 2019-07-11
Removes @library flag on typeclasses which have non-library subclasses

### LINKED ISSUE #521: Remove `@library` annotation on library methods which are extended from user-land code
This should hopefully solve the issue with type classes marked `@library` which don't get verified when instantiated in user-land.

Gorzen:
Removes the library flag to laws flagged as library that have non-library subclasses, in order to verify non-library case classes that extend library abstract classes.

Close #521


romac:
Great, good job! Will merge soon, please do not update this branch again :)

PR #599, opened 2019-07-16
Quick check that tools are installed before packaging

cgrigis:
Just a quick check that the required tools are installed on the machine before running the packaging, to avoid being surprised with errors (as I was since I did not have zip installed :-) ).


romac:
Thanks!

PR #605, opened 2019-07-17
Enforce sound equality tests

### LINKED ISSUE #372: Mismatch in semantics between Stainless and Scala can lead to runtime crashes
While discussing the upcoming support for inner classes (#271) with @jad-hamza and @vkuncak, we realised that those introduce a semantics mismatch with Scala in some cases.

Consider the following program:

```scala
  abstract class Test {
    def bar: BigInt
  }

  def foo(outer: BigInt): Test = {
    case class Foo(y: BigInt) extends Test {
      def bar = outer
    }
    Foo(12)
  }

  def oops = {
    // Assertion verifies but crashes at runtime when compiled with scalac
    assert(foo(1) != foo(2))
  }
```

The mismatch happens because, within Stainless, the inner class `Foo` gets lifted to the top-level, and the free variable `outer` gets added as a field, essentially turning the code above into:

```scala
  case class Foo(y: BigInt, outer: BigInt)

  def foo(outer: BigInt): Test = {
    Foo(12, outer)
  }
```

This means that Stainless will consider `foo(1) != foo(2)` because `Foo(42, 1) != Foo(42, 2)`, while scalac's `equals` methods only checks the original fields of the case class.

It should be noted that the Scala semantics break referential equality, ie. `foo(1) == foo(2) ==> foo(1).bar == foo(2).bar` does not hold in Scala, while it does in PureScala.

Nonetheless, as shown above this mismatch can lead to runtime crashes, and we should therefore do something about this, for example disallow equality tests on classes with have inner subclasses

@samarion @jad-hamza @vkuncak Thoughts?

PS: This is not the only mismatch in equality semantics between Stainless and Scala, lambdas notably contribute to this issue as well:

```scala
val a = (x: BigInt) => x
val b = (x: BigInt) => x
assert(a == b)
```

verifies with Stainless but crashes with scalac.

romac:
Close #372


core/src/main/scala/stainless/extraction/xlang/TreeSanitizer.scala
```diff
@@ -142,6 +197,132 @@ trait TreeSanitizer {
...
+      val ancestors = (ct.tcd +: ct.tcd.ancestors)
+      ancestors.exists(tcd => tcd.cd.isAbstract && !tcd.cd.isSealed)
+    }
+
+    override def traverse(e: Expr): Unit = e match {
+      case Annotated(body, flags) if flags contains Ghost =>
+        withinGhostContext(traverse(body))
+
+      case Decreases(_, body) =>
+        withinGhostContext(traverse(body))
```
samarion: This doesn't look right.
Only the measure should be in ghost context, no? Also you're not sanitizing the measure here.

core/src/main/scala/stainless/extraction/xlang/TreeSanitizer.scala
```diff
@@ -142,6 +197,132 @@ trait TreeSanitizer {
...
+
+  /** Disallow equality between lambdas and non-sealed abstract classes in non-ghost code */
+  private[this] class SoundEquality(syms: Symbols, ctx: inox.Context)
+    extends Sanitizer(syms, ctx) with SelfTreeTraverser {
+
+    final val PEDANTIC = true
+
+    private[this] var errors: ListBuffer[MalformedStainlessCode] = ListBuffer.empty
+
+    private[this] var ghostContext: Boolean = false
```
samarion: Don't we have a traverser with environment? That would be a cleaner solution no?
Also because expressions can appear in types (and vice-versa), it's generally safer to call `super.traverse` whenever possible instead of manually controlling the traversal.

Since we actually rely on knowing whether we are in ghost context in other parts of Stainless, it would maybe make sense to extend the traverser with environment into a ghost context traverser that implements this logic in a single place.

core/src/main/scala/stainless/extraction/xlang/TreeSanitizer.scala
```diff
@@ -142,6 +197,132 @@ trait TreeSanitizer {
       case _ => super.traverse(e)
     }
   }
+
+  /** Disallow equality between lambdas and non-sealed abstract classes in non-ghost code */
+  private[this] class SoundEquality(syms: Symbols, ctx: inox.Context)
+    extends Sanitizer(syms, ctx) with SelfTreeTraverser {
+
+    final val PEDANTIC = true
```
samarion: Haha awesome flag :)

core/src/main/scala/stainless/extraction/xlang/TreeSanitizer.scala
```diff
@@ -142,6 +197,132 @@ trait TreeSanitizer {
...
+            errors += MalformedStainlessCode(e, "Cannot compare local classes for equality in non-ghost code")
+          case (lct1, lct2: LocalClassType) =>
+            errors += MalformedStainlessCode(e, "Cannot compare local classes for equality in non-ghost code")
+
+          case (ft1: FunctionType, ft2) =>
+            errors += MalformedStainlessCode(e, "Cannot compare lambdas for equality in non-ghost code")
+          case (ft1, ft2: FunctionType) =>
+            errors += MalformedStainlessCode(e, "Cannot compare lambdas for equality in non-ghost code")
+
+          case _ => ()
```
samarion: You need to recursively traverse the type here. What if I'm comparing tuples of functions?

romac: Good catch, thanks!

core/src/main/scala/stainless/extraction/xlang/TreeSanitizer.scala
romac: Yeah, good point. I actually thought about writing a GhostTraverser based on those traversers but eventually decided against it, as this is currently the only place where we do such a traversal. GhostChecker in imperative does a different kind of traversal it seems, although it might be possible to express both as a GhostTraverser. Anyway, point taken, I'll write a GhostTraverser :)

core/src/main/scala/stainless/extraction/xlang/TreeSanitizer.scala
romac: > Only the measure should be in ghost context, no?

Right!

> Also you're not sanitizing the measure here.

Do I really need to sanitize the measure? Isn't it assumed to be ghost?

https://github.com/epfl-lara/stainless/blob/8475105be41b8f848357c38a9886877a1099a937/frontends/library/stainless/lang/package.scala#L64-L68

romac:
@samarion 

> In order to be sound wrt Scala equality semantics, you also have to implement a type class approach for type parameters. As a first approximation, you can just be pedantic and reject equality between type parameters. 
Thank you very much for the review! I did this a bit too quickly…

And yeah I kinda forgot about type parameters. Indeed, this is what I will be doing at first, and will convert benchmarks to explicitly using an `Eq` type class.

> An ideal approach would both allow annotating the equality type class and infer it based on method bodies.

Can you expand on what you meant by inferring the type class? Kinda like doing our own implicit search for `Eq[A]` when we have `(x: B) == (y: C)` with `lub(B, C) == A`, and rewriting the equality to use the appropriate type class instance? Wouldn't that potentially create a mismatch between the standard Scala equality and the typeclass-based one?

core/src/main/scala/stainless/extraction/xlang/TreeSanitizer.scala
romac: > Do I really need to sanitize the measure? Isn't it assumed to be ghost?

Nevermind, you are right! :)

PR #608, opened 2019-07-18
Enable strict arithmetic by default

### LINKED ISSUE #598: Enable --strict-arithmetic by default?

romac:
Close #598

### DIFF
--- a/core/src/main/scala/stainless/ast/Deconstructors.scala
+++ b/core/src/main/scala/stainless/ast/Deconstructors.scala
@@ -216,6 +216,7 @@ trait TreeDeconstructor extends inox.ast.TreeDeconstructor {
     case s.IsField(isLazy) => (Seq(), Seq(), Seq(), (_, _, _) => t.IsField(isLazy))
     case s.IsUnapply(isEmpty, get) => (Seq(isEmpty, get), Seq(), Seq(), (ids, _, _) => t.IsUnapply(ids(0), ids(1)))
     case s.PartialEval => (Seq(), Seq(), Seq(), (_, _, _) => t.PartialEval)
+    case s.Wrapping => (Seq(), Seq(), Seq(), (_, _, _) => t.Wrapping)
     case _ => super.deconstruct(f)
   }
 }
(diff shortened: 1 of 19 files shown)


vkuncak:
Is there anything blocking this? Or maybe we introduce these errors as warnings and run them as default, whereas strict-arithmetic turns the warnings into errors?

romac:
> Is there anything blocking this?

A lot of benchmarks are failing and I haven’t gotten around to fix them all yet.

> Or maybe we introduce these errors as warnings and run them as default, whereas strict-arithmetic turns the warnings into errors?

Yeah that is likely a sensible middle-ground, regardless of the status of the benchmarks. I’ll get on it and will fix the benchmarks all the same while I am at it, so that we don’t pollute the CI log with warnings anyway.

romac:
For reference, here are the benchmarks that currently fail with strict arithmetic turned on: https://gist.github.com/romac/7eac6e7c474e0785e4fc305fcc3d52f0

vkuncak:
I am reading these benchmarks and they seem fine with me. It's just that Int is a modular arithmetic. Given that we check e.g. negative indices and that, if things should be increasing, we should specify them, I am having second thoughts on whether we should fix the benchmarks. How about define a type alias 
```scala
type Int32 = Int
```
but then if you have something of this type then you do not report warnings on it? Ideally you would propagate a bit so you do emit warning on code such as
```scala
def f(x : Int32, y: Int) = x + y + 1
```

romac:
As discussed today, we are just going to disable --strict-arithmetic for regular benchmarks, since we already have a StrictArithmetic test suite which exercises this feature.

romac:
We still need a way for users to instruct Stainless that wrapping around is the expected behaviour for some expression, and thus opt-out of strict-arithmetic VC generation for this expression.

I suggest we go the Rust way and:

- Introduce such a `Wrapping` value class, with `+`, `-`, etc.
- Also provide an implicit value class with `wrapping_add`, `wrapping_sub`, etc.

and disable strict-arithmetic VC gen for calls to these methods.

What do you think?

romac:
I think there is a better way of opting out of overflow checks, introduce an annotation `@wrapping` and a method

```scala
package stainless.math

@ignore
def wrapping[A](a: A): A = a
```

which is extracted as:

```scala
wrapping { body }

=>

@wrapping { body }
```

Then, we just disable overflow checks within any expression annotated with `@wrapping`.

romac:
Note to self: Push documentation after merging.

vkuncak:
@romac : This indeed sounds like a good plan: make strict-arithmetic default, introduce wrapping pseudo-function and @wrapping annotation for named methods. For benchmarks either change the options when invoking tests, or add the @wrapping once available.

PR #620, opened 2019-07-25
Do not lift refinement into pre-/post-conditions when --type-checker is enabled

romac:


romac:
That, and they also won’t trigger any VC at all actually, neither as pre- or postconditions.

There is a WIP PR to actually go the other way when the type checker is enabled, ie. lift contracts into refinements.

jad-hamza:
The second PR is to remove the special handling of pre and post in the type-checker?

romac:
It would enable that, yes.

PR #621, opened 2019-07-26
Follow symbolic links when searching for base directory

jad-hamza:
@gsps Is `cd` followed by `pwd` more robust than directly using `dirname`?

I added a call to `readlink` to follow symbolic links, in case someone makes a symbolic link pointing to the generated `stainless` script.

### DIFF
--- a/bin/launcher.tmpl.sh
+++ b/bin/launcher.tmpl.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+
+### The following section defines `realpath` to resolve symbolic links.
+###
+### Copyright (c) 2014 Michael Kropat
+### Licensed under the terms of the MIT License (https://opensource.org/licenses/MIT)
+### Original code at https://github.com/mkropat/sh-realpath
+
+realpath() {
+    canonicalize_path "$(resolve_symlinks "$1")"
+}
+
+resolve_symlinks() {
+    _resolve_symlinks "$1"
+}
+
+_resolve_symlinks() {
+    _assert_no_path_cycles "$@" || return
+
+    local dir_context path
+    path=$(readlink -- "$1")
+    if [ $? -eq 0 ]; then
+        dir_context=$(dirname -- "$1")
+        _resolve_symlinks "$(_prepend_dir_context_if_necessary "$dir_context" "$path")" "$@"
+    else
+        printf '%s\n' "$1"
+    fi
+}
+
+_prepend_dir_context_if_necessary() {
+    if [ "$1" = . ]; then
+        printf '%s\n' "$2"
+    else
+        _prepend_path_if_relative "$1" "$2"
+    fi
+}
+
+_prepend_path_if_relative() {
+    case "$2" in
+        /* ) printf '%s\n' "$2" ;;
+         * ) printf '%s\n' "$1/$2" ;;
+    esac
+}
+
+_assert_no_path_cycles() {
+    local target path
+
+    target=$1
+    shift
+
+    for path in "$@"; do
+        if [ "$path" = "$target" ]; then
+            return 1
+        fi
+    done
+}
+
+canonicalize_path() {
+    if [ -d "$1" ]; then
+        _canonicalize_dir_path "$1"
+    else
+        _canonicalize_file_path "$1"
+    fi
+}
+
+_canonicalize_dir_path() {
+    (cd "$1" 2>/dev/null && pwd -P)
+}
+
+_canonicalize_file_path() {
+    local dir file
+    dir=$(dirname -- "$1")
+    file=$(basename -- "$1")
+    (cd "$dir" 2>/dev/null && printf '%s/%s\n' "$(pwd -P)" "$file")
+}
+
+### end of realpath code
+
+BASE_DIR="$( dirname "$( realpath "${BASH_SOURCE[0]}" )" )"
+Z3_DIR="$BASE_DIR/z3"
+STAINLESS_JAR="$BASE_DIR/lib/{STAINLESS_JAR_BASENAME}"
+SCALAZ3_JAR="$BASE_DIR/lib/{SCALAZ3_JAR_BASENAME}"
+JARS="$STAINLESS_JAR:$SCALAZ3_JAR"
+
+for JAR in "$STAINLESS_JAR" "$SCALAZ3_JAR"; do
+  if ! [[ -r "$JAR" ]]; then
+    echo "Read access for the jar file $JAR is required."
+    exit 1
+  fi
+done
+
+exec env PATH="$Z3_DIR:$PATH" java -cp "$JARS $JAVA_OPTS" stainless.Main "$@"
--- a/bin/package-standalone.sh
+++ b/bin/package-standalone.sh
@@ -93,24 +93,11 @@ function fetch_z3 {
 function generate_launcher {
   local TARGET="$1"
   local SCALAZ3_JAR_BASENAME="$2"
-  cat << END > "$TARGET"
-#!/usr/bin/env bash
 
-BASE_DIR="\$( cd "\$( dirname "\${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
-Z3_DIR="\$BASE_DIR/z3"
-STAINLESS_JAR="\$BASE_DIR/lib/$STAINLESS_JAR_BASENAME"
-SCALAZ3_JAR="\$BASE_DIR/lib/$SCALAZ3_JAR_BASENAME"
-JARS="\$STAINLESS_JAR:\$SCALAZ3_JAR"
-
-for JAR in \$STAINLESS_JAR \$SCALAZ3_JAR; do
-  if ! [[ -r \$JAR ]]; then
-    echo "Read access for the jar file \$JAR is required."
-    exit 1
-  fi
-done
-
-exec env PATH="\$Z3_DIR:\$PATH" java -cp \$JARS \$JAVA_OPTS stainless.Main "\$@"
-END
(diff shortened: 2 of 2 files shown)


romac:
Unfortunately, `-f` argument to `readlink` is only present in the GNU version of the tool. I think we should revert to the previous method instead use `pwd -P` which seems to be present with both the GNU and BSD toolset (it's in the POSIX standard), and will resolve symlinks.

jad-hamza:
Oh I see thanks. I wasn't sure it was reasonable to assume `readlink -f` was present everywhere. Would `pwd -P` help if only the script file is sym-linked and not the whole folder?

romac:
> Would `pwd -P` help if only the script file is sym-linked and not the whole folder?

Right, it does not.

We could use this instead from  https://github.com/mkropat/sh-realpath/

```sh
# Utility from https://github.com/mkropat/sh-realpath/

realpath() {
    canonicalize_path "$(resolve_symlinks "$1")"
}

resolve_symlinks() {
    _resolve_symlinks "$1"
}

_resolve_symlinks() {
    _assert_no_path_cycles "$@" || return

    local dir_context path
    path=$(readlink -- "$1")
    if [ $? -eq 0 ]; then
        dir_context=$(dirname -- "$1")
        _resolve_symlinks "$(_prepend_dir_context_if_necessary "$dir_context" "$path")" "$@"
    else
        printf '%s\n' "$1"
    fi
}

_prepend_dir_context_if_necessary() {
    if [ "$1" = . ]; then
        printf '%s\n' "$2"
    else
        _prepend_path_if_relative "$1" "$2"
    fi
}

_prepend_path_if_relative() {
    case "$2" in
        /* ) printf '%s\n' "$2" ;;
         * ) printf '%s\n' "$1/$2" ;;
    esac
}

_assert_no_path_cycles() {
    local target path

    target=$1
    shift

    for path in "$@"; do
        if [ "$path" = "$target" ]; then
            return 1
        fi
    done
}

canonicalize_path() {
    if [ -d "$1" ]; then
        _canonicalize_dir_path "$1"
    else
        _canonicalize_file_path "$1"
    fi
}

_canonicalize_dir_path() {
    (cd "$1" 2>/dev/null && pwd -P)
}

_canonicalize_file_path() {
    local dir file
    dir=$(dirname -- "$1")
    file=$(basename -- "$1")
    (cd "$dir" 2>/dev/null && printf '%s/%s\n' "$(pwd -P)" "$file")
}

BASE_DIR="$( dirname $(realpath ${BASH_SOURCE[0]}) )"
```

romac:
I tested it with both symlinks in the path to the script, and a symlink to the script itself + combination of the two.

jad-hamza:
Nice! Thanks. We should make the generator standalone so that we don't have to escape everything in `generate_launcher`. Maybe take the things that are not static as arguments of the script?

romac:
I'd rather not ship too many things in the ZIP file, and have the launcher standalone as to not confuse people. But yes there is perhaps a way to define the launcher in a separate file, and then #include it or something in the packaging script, or just copy it into the folder to be zipped and perform search and replace over it (eg. with `sed`).

jad-hamza:
Sure, I'll try!

romac:
Something like having this in `bin/launcher.tmpl.sh`:

```sh
#!/usr/bin/env bash

realpath() {
    canonicalize_path "$(resolve_symlinks "$1")"
}

resolve_symlinks() {
    _resolve_symlinks "$1"
}

_resolve_symlinks() {
    _assert_no_path_cycles "$@" || return

    local dir_context path
    path=$(readlink -- "$1")
    if [ $? -eq 0 ]; then
        dir_context=$(dirname -- "$1")
        _resolve_symlinks "$(_prepend_dir_context_if_necessary "$dir_context" "$path")" "$@"
    else
        printf '%s\n' "$1"
    fi
}

_prepend_dir_context_if_necessary() {
    if [ "$1" = . ]; then
        printf '%s\n' "$2"
    else
        _prepend_path_if_relative "$1" "$2"
    fi
}

_prepend_path_if_relative() {
    case "$2" in
        /* ) printf '%s\n' "$2" ;;
         * ) printf '%s\n' "$1/$2" ;;
    esac
}

_assert_no_path_cycles() {
    local target path

    target=$1
    shift

    for path in "$@"; do
        if [ "$path" = "$target" ]; then
            return 1
        fi
    done
}

canonicalize_path() {
    if [ -d "$1" ]; then
        _canonicalize_dir_path "$1"
    else
        _canonicalize_file_path "$1"
    fi
}

_canonicalize_dir_path() {
    (cd "$1" 2>/dev/null && pwd -P)
}

_canonicalize_file_path() {
    local dir file
    dir=$(dirname -- "$1")
    file=$(basename -- "$1")
    (cd "$dir" 2>/dev/null && printf '%s/%s\n' "$(pwd -P)" "$file")
}

BASE_DIR="$( dirname "$( realpath "${BASH_SOURCE[0]}" )" )"
Z3_DIR="{BASE_DIR}/z3"
STAINLESS_JAR="{BASE_DIR}/lib/{STAINLESS_JAR_BASENAME}"
SCALAZ3_JAR="{BASE_DIR}/lib/{SCALAZ3_JAR_BASENAME}"
JARS="{STAINLESS_JAR}:{SCALAZ3_JAR}"

for JAR in "{STAINLESS_JAR}" "{SCALAZ3_JAR}"; do
  if ! [[ -r "{JAR}" ]]; then
    echo "Read access for the jar file {JAR} is required."
    exit 1
  fi
done

exec env PATH="{Z3_DIR}:$PATH" java -cp {JARS} {JAVA_OPTS} stainless.Main "$@"
``

and then in the packaging script, we could do:

```sh
cat launcher.tmpl.sh | \
  sed "s#{BASE_DIR}#$(pwd)#g" | \
  sed etc. | \
  sed etc. \
  > "$TARGET"`
```

romac:
Edited the command at the end.

jad-hamza:
Thanks I'll add that. Is it ok license wise to include that code?

romac:
Yes, as it is licensed under the MIT, which is compatible with Apache 2.0, provided we include the proper attribution.

romac:
By proper attribution, I mean something like

```
Copyright (c) 2014 Michael Kropat
Licensed under the terms of the MIT License (https://opensource.org/licenses/MIT)
Original code at https://github.com/mkropat/sh-realpath
```

jad-hamza:
It should be ok now, thanks for the help. I tested the generated package, but we should remember testing it again for next release.

I put `bin/launcher.tmpl.sh` in the script so that we  launch the script from the root, is that ok?

> bin/package-standalone.sh

romac:
Yes that’s fine, I don’t expect to launch it from anywhere else, though if you have time you could alternatively use the old $(cd $(dirname BASH_SOURCE) && pwd)) trick to get the path to the the `bin` directory from the packaging script, just to make it a bit more robust.

jad-hamza:
Why not just `dirname`?

I added it, but it looks like the script already assumes it is at the root to interact with `sbt`.

romac:
Right, dirname is enough here. Oh you are right actually, nevermind then!

bin/package-standalone.sh
```diff
@@ -93,24 +93,11 @@ function fetch_z3 {
...
-for JAR in \$STAINLESS_JAR \$SCALAZ3_JAR; do
-  if ! [[ -r \$JAR ]]; then
-    echo "Read access for the jar file \$JAR is required."
-    exit 1
-  fi
-done
 
-exec env PATH="\$Z3_DIR:\$PATH" java -cp \$JARS \$JAVA_OPTS stainless.Main "\$@"
-END
+  cat "$(dirname BASH_SOURCE)/launcher.tmpl.sh" | \
```
romac: Shouldn't this line read:

```sh
cat "$(dirname "${BASH_SOURCE[0]}")/launcher.tmpl.sh" | \
```

Note: `dirname non_existing_file` seems to always return `.`, which is perhaps why the script still worked when tested.

bin/launcher.tmpl.sh
```diff
@@ -0,0 +1,92 @@
...
+JARS="$STAINLESS_JAR:$SCALAZ3_JAR"
+
+for JAR in "$STAINLESS_JAR" "$SCALAZ3_JAR"; do
+  if ! [[ -r "$JAR" ]]; then
+    echo "Read access for the jar file $JAR is required."
+    exit 1
+  fi
+done
+
+exec env PATH="$Z3_DIR:$PATH" java -cp "$JARS $JAVA_OPTS" stainless.Main "$@"
```
romac: This should read:

```sh
exec env PATH="$Z3_DIR:$PATH" java -cp "$JARS" "$JAVA_OPTS" stainless.Main "$@"
```

otherwise `$JAVA_OPTS` will be passed as an argument to `-cp`, I think.

jad-hamza: If I put them separately in quotes, I get the error:

> Erreur : impossible de trouver ou charger la classe principale

I completely removed the quotes for now

bin/package-standalone.sh
jad-hamza: Thanks, I put back `bin` since we must be in the root.

bin/launcher.tmpl.sh
romac: I guess that's because if JAVA_OPTS is not defined then `java` treats the empty quotes surrounding the argument as the name of the class. Leaving it without quotes is appropriate here since JAVA_OPTS can contain spaces and those delimit the different arguments to the JVM.

bin/package-standalone.sh
romac: 👍

romac:
LGTM!

PR #623, opened 2019-07-26
Add script to run tests on stainless-actors and bolts

jad-hamza:
This is a script we discussed with @romac that can be used to run more extensive tests on external benchmarks (bolts, stainless-actors).

### DIFF
--- a/.larabot.conf
+++ b/.larabot.conf
@@ -1,5 +1,4 @@
 commands = [
-    "bin/external-tests.sh"
     "sbt -batch -Dparallel=5 test"
     "sbt -batch -Dparallel=5 it:test"
     "sbt -batch scripted"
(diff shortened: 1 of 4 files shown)


romac:
Nice! Could you try adding them to .larabot.conf commands so that we can see how much time they add to the test suite?

jad-hamza:
The tests failed but were shown as `succeeded`. I have to play around with the exit codes. I think it will be around 20 minutes total (leader-election is 13 minutes)

jad-hamza:
I ran the tests on this branch https://github.com/jad-hamza/stainless-actors/tree/configs-fix: http://laraquad4.epfl.ch:9000/epfl-lara/stainless/build/4044 (with and without `--type-checker`). The `external-tests.sh` script took 22 minutes, and the overall build 46 minutes.

PR #640, opened 2019-08-19
Fix algebra test suite

romac:


romac:
@Gorzen Looks like it is indeed the termination check in LibrarySuite that fails.

PR #645, opened 2019-08-21
Algebraic hierarchy of type classes and associated instances for the Stainless library

romac:
Rebased version of #636


romac:
This slows the test suite down too much to be merged at the moment.

We need some kind of way to extract only the part of the library which are relevant to the program, rather than pulling all of it. Perhaps with something similar to `@keep` annotations and the `--keep` command-line flag?

PR #651, opened 2019-08-28
Remove warnings for asserts in extern functions

jad-hamza:
Asserts at the top-level would be ignored as well right?


romac:
LGTM

Issue #644, opened 2019-08-20
@induct breaks termination


Gorzen:
If we check the termination for the following program we get a successful output.
```scala
import stainless.lang._
import stainless.annotation._
import stainless.proof._

object BugTermination {
  abstract class SemiGroup[T] {
    def combine(x: T, y: T): T

    @law
    def law_associativity(x: T, y: T, z: T): Boolean = {
      combine(combine(x, y), z) == combine(x, combine(y, z))
    }
  }
  
  case class BugTermination[T](ev: SemiGroup[T]) extends SemiGroup[Option[T]] {
    def combine(x: Option[T], y: Option[T]): Option[T] = {
      (x, y) match {
        case (Some(a), Some(b)) => Some(ev.combine(a, b))
        case (Some(a), None()) => Some(a)
        case (None(), Some(b)) => Some(b)
        case (None(), None()) => None[T]()
      }
    }
  
    override def law_associativity(x: Option[T], y: Option[T], z: Option[T]): Boolean = {
      (combine(combine(x, y), z) == combine(x, combine(y, z))) because {
        (x, y, z) match {
          case (Some(a), Some(b), Some(c)) => check(ev.law_associativity(a, b, c))
          case _ => true
        }
      }
    }
  }
}
```
*The output...*
```
[info] [  Info  ] Result for isSemiGroupExt
[info] [  Info  ]  => CLEARED (Non-recursive)
[info] [  Info  ] Result for combine
[info] [  Info  ]  => CLEARED (Non-recursive)
[info] [  Info  ] Result for isOption
[info] [  Info  ]  => CLEARED (Non-recursive)
[info] [  Info  ] Result for combine
[info] [  Info  ]  => CLEARED (Non-recursive)
[info] [  Info  ] Result for combine
[info] [  Info  ]  => CLEARED (Non-recursive)
[info] [  Info  ] Result for combine
[info] [  Info  ]  => CLEARED (Non-recursive)
[info] [  Info  ] Result for isBugTermination
[info] [  Info  ]  => CLEARED (Unchecked)
[info] [  Info  ] Result for isSemiGroup
[info] [  Info  ]  => CLEARED (Unchecked)
[info] [  Info  ] Result for combine
[info] [  Info  ]  => CLEARED (Relation Processor comparing argument lists lexicographically)
[info] [  Info  ] Result for combine
[info] [  Info  ]  => CLEARED (Relation Processor comparing argument lists lexicographically)
[...]

PR #657, opened 2019-08-29
Induct flag only adds decreases check if type checker is enabled

Gorzen:


jad-hamza:
> Can you please explain the rationale for this change? It seems to me that this commit does not actually change the behaviour of the test suite, is that right?
> 
> EDIT: By change I mean the second commit: [936f5f2](https://github.com/epfl-lara/stainless/commit/936f5f251232e66a5afb770b9a74756b39ffbedf)

The problem is that in the test suite, there are two contexts (one empty, and one with the correct options), and it looks like the empty context was used for the pipeline (and the type-checker option was off during `InductElimination`).

romac:
@jad-hamza I see, thanks!

Gorzen:
Fix #644

romac:
@Gorzen Can you please check why the build failed?

Gorzen:
> @Gorzen Can you please check why the build failed?

It's because of ParialCompiler and BodyEnsuring, check-models fails on them

romac:
Blocked by#676

Gorzen:
This may break caching in ComponentTestSuite, but as of right now options aren't passed properly to the extractor. (TypeCheckerEnabled) Which causes the tests to break as now we only add `decreases` if the type checker is enabled.

romac:
Nice, thanks!

PR #659, opened 2019-08-31
Verification using tactics in Coq

jad-hamza:
This adds work done during @czipobence's project on translation to Coq. The transformation replaces the `VerificationChecker` when the `--coq` option is turned on. Some small tests are given in `frontends/benchmarks/coq` and have been added to the nightly tests.


romac:
@czipobence Could please sign the CLA by following this link: https://cla-assistant.io/epfl-lara/stainless?pullRequest=659? Thanks!

.larabot.conf
```diff
@@ -5,10 +5,12 @@ commands = [
 
 nightly {
   commands = [
+    "sbt universal:stage"
     "sbt -batch -Dparallel=5 test"
     "sbt -batch -Dparallel=5 it:test"
     "bash bin/external-tests.sh"
     "sbt -batch scripted"
+    "frontends/scalac/target/universal/stage/bin/stainless-scalac --coq frontends/benchmarks/coq/*.scala"
```
romac, line 13: @jad-hamza How about `set -batch stainless-scalac/run -coq ../benchmarks/coq/*.scala` here instead? Is the issue that glob expansion is not triggered here?

jad-hamza: I think this was working but the library needs to be built using `cd slc-lib; ./configure; make; cd ..`

PR #660, opened 2019-09-02
Fix some missing and wrong positions

jad-hamza:
Blocked by https://github.com/epfl-lara/inox/pull/107


jad-hamza:
I temporarily introduced some functions, so this is no longer blocked by https://github.com/epfl-lara/inox/pull/107

PR #668, opened 2019-09-03
Inline functions also inside their own bodies

jad-hamza:
This allows (recursive) calls inside a recursive functions to be inlined as well.


jad-hamza:
Oh I see there are some tests set up, maybe it was like that for a good reason?

I was thinking the recursive call to `f` here should be inlined:

```scala
@inlineOnce
def f(): BigInt  = {
  val x = f()
  0
}
```

romac:
Yeah, it was indeed a choice we made, but I unfortunately don't really recall why (should've documented that somewhere…)

Perhaps @samarion has a better memory than me?

PR #671, opened 2019-09-04
Fix type instantiation in type-checker and improve VC generation

jad-hamza:


jad-hamza:
The new commit changes the way that VCs are generated, by using `let`s instead of equalities. Inox simplifications work well on `let`'s but not on equalities. The generated VCs are now simpler. (@romac: maybe this is what made `TweetSet` fail?)

jad-hamza:
Subsumed by https://github.com/epfl-lara/stainless/pull/674

Issue #673, opened 2019-09-04
TypeCheckerSuite isn't run on larabot on master, it:test


Gorzen:
The bot doesn't seem to run the TypeCheckerSuite

http://laraquad4.epfl.ch:9000/epfl-lara/stainless/build/4265/24778

Only VerificationSuite is run, probably do to some extraction error. The bot starts running it but then stops.

jad-hamza:
Thanks, I'm trying to refactor the benchmarks to take into account the `qed` type change in https://github.com/epfl-lara/stainless/pull/674
I also added an assertion to make sure there are no errors during extraction

romac:
I just merged #674, does this mean we can close this issue?

jad-hamza:
Yes, thanks!

PR #674, opened 2019-09-04
Qed unit refactor

jad-hamza:


jad-hamza:
Looks like all is well! This fixes some bugs in type-checker and https://github.com/epfl-lara/stainless/issues/673

PR #685, opened 2019-09-10
Properly encode generic bounds on type parameters

### LINKED ISSUE #666: Type error when using type bounds in case classes
```scala
object Pair {
  case class Pair[A](x: A, y: A)

  def makePair[A, B >: A](x: A, y: B): Pair[B] = {
    Pair[B](x, y)
  }
}
```

```
[ Fatal  ] Run has failed with error: inox.ast.TypeOps$TypeErrorException: Type error: Pair[B](x, y), expected Pair[B], 
[ Fatal  ] found <untyped>
[ Fatal  ] 
[ Fatal  ] Typing explanation:
[ Fatal  ] Pair[B](x, y) is of type <untyped>
[ Fatal  ]   x is of type A
[ Fatal  ]   y is of type B because Pair was instantiated with A:=B with fields (B,B)
```

romac:
Close #666


romac:
Note to self: Squash commit when merging to use PR title instead of commit title.

PR #686, opened 2019-09-10
Document type aliases and type members

### LINKED ISSUE #682: Document type members support

romac:
Close #682


romac:
Looks like I broke larabot, will investigate…

PR #732, opened 2020-02-07
Less measure inference

jad-hamza:


romac:
I had tried that, but it seems to me that we actually need to pass functions which already have a measure through the whole phase as well, as we sometimes need the `DecreasesProcessor` to rewrite measures of inter-dependent functions to give them the same shape.

jad-hamza:
What if we remove the measures for these inter-dependent functions?

romac:
Well, then we wouldn't be able to provide measures for functions for which we cannot infer them.

But perhaps we could force the programmer to write measures of the same shapes for mutually recursive groups of functions?

jad-hamza:
> But perhaps we could force the programmer to write measures of the same shapes for mutually recursive groups of functions?

Yes this makes sense. I think either the programmer either gives the same measure type for all mutually recursive functions, or no measure at all for these functions.

jad-hamza:
> Well, then we wouldn't be able to provide measures for functions for which we cannot infer them.

What do you mean?

romac:
> > But perhaps we could force the programmer to write measures of the same shapes for mutually recursive groups of functions?
> 
> Yes this makes sense. I think either the programmer either gives the same measure type for all mutually recursive functions, or no measure at all for these functions.

Alright, then I'll see if I can get to it next week, as I am in the middle of cutting a new release right now.

romac:
> > Well, then we wouldn't be able to provide measures for functions for which we cannot infer them.
> 
> What do you mean?

Perhaps I misunderstood what you meant. I was saying that if we automatically remove measures on functions which already have some, then we cannot override the measure inference mechanism.

jad-hamza:
Ah, I meant we remove them at the user-level (for the ones that can be inferred)

romac:
> Ah, I meant we remove them at the user-level (for the ones that can be inferred)

Sure, that makes more sense :)

jad-hamza:
This is going through now, I added a few improvements for error handling (I didn't test them a lot). Now the phase `MeasureInference` complains if a function has a measure but a mutually recursive function doesn't. The typechecker doesn't crash when a measure is missing, but adds a dummy (false) VC when a measure is missing.

core/src/main/scala/stainless/termination/MeasureInference.scala
```diff
@@ -54,7 +59,26 @@ trait MeasureInference
...
     }
 
-    def needsMeasure(fd: FunDef): Boolean = symbols.isRecursive(fd.id)
+    def needsMeasure(fd: FunDef): Boolean = {
+      if (symbols.isRecursive(fd.id) && fd.measure(symbols).isEmpty) {
+        symbols.dependencies(fd.id).find(id =>
+          symbols.lookupFunction(id).exists(fd2 =>
+            symbols.dependencies(fd2.id).contains(fd.id) &&
+            !fd2.measure(symbols).isEmpty
+          )
```
romac, line 68: This block could perhaps be given a name to improve readability?

jad-hamza: Thanks!

Issue #733, opened 2020-02-07
Wrong line number for assertion


jad-hamza:
```scala
case class C() {
  val f = true

  def g(): Unit = {
    assert(f)
  }
}
```

The assertion is reported at line 2 (the line of `f`) instead of 5

romac:
Closed in 322ca010b65236d0649b7593ccf24a8241032aa8.

PR #734, opened 2020-02-07
Bad line field accessor

jad-hamza:


romac:
FYI the build got killed/stuck, so I just restarted it.

romac:
@jad-hamza Does this fix #733?

jad-hamza:
> @jad-hamza Does this fix #733?

Yes, but I'm not sure that's the correct way to do it?

romac:
Looks good to me. I merged it manually.

Issue #573, opened 2019-06-28
Track progress of type-checker features


jad-hamza:
Fix existing features:
* [ ] Inox doesn’t restrict the measure for which we can unfold non-decreasing recursive calls (see Nicolas’ example using lambda equality)
* [ ] Figure out how we want to handle indexed recursive types in `Inox`. At the moment they are just erased into their non-indexed versions (probably unsound?)
* [ ] ADT invariants are not checked

New core features:
* [ ] Definition of `if then else` on types (`ite`) is not complete
* [ ] forall/exists propositions
* [ ] Remove the constraint that all types are strictly positive, and add the corresponding checks during type-checking

New convenience features:
* [ ] Type inference for `match` (now goes through `if then else` transformation). (Not Important)
* [ ] Enable erasure of ghost parameters (such as the indices `n` in Stream functions)
* [ ] Add type generalization (forall n. Stream n -> Stream n is a subtype of Stream -> Stream).

User interface features:
* [ ] A frontend for recursive types and dependent types (partially working on Dotty with annotations, see [Streams.scala]()
* [ ] A phase to infer the measures automatically
* [ ] A way to get feedback from the solver to update the HTML derivation trees in real-time as VCs get solved

romac:
Moved to #687

PR #737, opened 2020-02-11
Check match exhaustiveness in type-checker

romac:


jad-hamza:
Isn't this covered by the translation `matchToIfThenElse` (when given `assumeExhaustive = false`)?

romac:
Oh right, I didn't notice that.

In any case, I think that having a dedicated VC is a bit better for the user experience.

What do you think? Are there any downside to that?

jad-hamza:
Yes I agree, the error reporting I wrote using exceptions everywhere is quite clumsy. I'm thinking maybe the current typing context should always hold an error message to know what to say when it fails.

If we add the separated exhaustiveness check, should we give `assumeExhaustive = true`?

romac:
Oh, actually a VC was already emitted as a body assertion, but not always (ie. when the exhaustiveness check is trivial it seems).

But yeah I would keep as a dedicated VC, and pass `assumeExhaustive = true` to `matchToIfThenElse`. And perhaps, even type check the `MatchExpr` directly without translating it to an `if-then-else`?

romac:
> I'm thinking maybe the current typing context should always hold an error message to know what to say when it fails.

Yes, that sounds like a good idea!

jad-hamza:
> Oh, actually a VC was already emitted as a body assertion, but not always (ie. when the exhaustiveness check is trivial it seems).
> 
> But yeah I would keep as a dedicated VC, and pass `assumeExhaustive = true` to `matchToIfThenElse`. And perhaps, even type check the `MatchExpr` directly without translating it to an `if-then-else`?

Yes that's on the TODO list :) https://github.com/epfl-lara/stainless/issues/573

PR #738, opened 2020-02-11
Add rule for checking assertions (instead of inferring)

jad-hamza:


jad-hamza:
This PR was failing because of an issue with `@unchecked` expressions. Now I've made it so that checking: `(@unchecked e): { x: Nat | p(x) }` doesn't generate verification conditions while inferring a type for `e`, but does generate verification conditions when checking the refinement.

In the algorithm, if we need to check `(@unchecked e): B`, we first infer a type `A` for `e` without generating verification conditions, and then check that `{ x: A | x = e }` is a subtype of `B` (potentially generating verification conditions).

PR #741, opened 2020-02-18
Recommend using Z3 4.8.6 instead of 4.7.1

jad-hamza:


romac:
What's the rationale behind inlining the invariant? Is it only to work around the Z3 bug?

jad-hamza:
Initially it was a workaround to the 4.8.7 bug yes, but since we're using 4.8.6 (nevermind the branch name :)) it shouldn't be needed. I thought I'd still leave this annotation so that the user can decide to inline or not (I guess this can speed up the verification in some cases).

vkuncak:
Where is inlining actually taking place (which file and line)? I was curious but did not find it in the diff.

jad-hamza:
It's in TypeChecker.scala when we see an ADT constructor (801 and 841)

jad-hamza:
Can we merge this?

vkuncak:
Yes. Can you do it, @jad-hamza ?

PR #745, opened 2020-05-08
TraceInduct implementation

drganam:


core/src/main/scala/stainless/extraction/trace/Trace.scala
```diff
@@ -0,0 +1,156 @@
...
+  private[this] object identity extends transformers.TreeTransformer {
+    override val s: self.s.type = self.s
+    override val t: self.t.type = self.t
+  }
+
+  override protected def extractFunction(symbols: Symbols, fd: FunDef): t.FunDef = {
+    import symbols._
+    var funInv: Option[FunctionInvocation] = None
+
+    if(fd.flags.filter(elem => elem.name == "traceInduct").size == 1) {
```
romac, line 31: How about `fd.flags.contains(TraceInduct)`?

core/src/main/scala/stainless/extraction/trace/Trace.scala
```diff
@@ -0,0 +1,156 @@
...
+      fd.flags.filter(elem => elem.name == "traceInduct").head match {
+        case Annotation("traceInduct", fun) => {
```
romac, line 33: It would be a bit cleaner to define the `TraceInduct` flag as a case class, rather than pattern match on the untyped `Annotation` params. See below.

core/src/main/scala/stainless/extraction/trace/Trees.scala
```diff
@@ -0,0 +1,41 @@
+/* Copyright 2009-2018 EPFL, Lausanne */
+
+package stainless
+package extraction
+package trace
+
+trait Trees extends termination.Trees { self =>
+
+  //case object Induct extends Flag("induct", Seq())
+  case object TraceInduct extends Flag("traceInduct", Seq())
```
romac, line 10: You could define this as `case class TraceInduct(fun: Identifier)`, and extract the `fun` argument in `extractFlag` below.

romac:
I added a few inline comments, but it looks good otherwise :)

core/src/main/scala/stainless/extraction/trace/Trace.scala
```diff
@@ -0,0 +1,156 @@
...
+ 
+
+  private def replaceFromIDs(substs: Map[Variable, Expr], expr: Expr) : Expr = {
+    exprOps.postMap({
+      case a: Variable => substs.get(a)
+      case _ => None
+    })(expr)
+  }
+
+  private def createAnd(exprs: Seq[Expr]): Expr = {
```
romac, line 129: This already exists as `andJoin` .

See https://github.com/epfl-lara/inox/blob/f0f34a033e033c24951a1aa993efa6ee52840db0/src/main/scala/inox/ast/Constructors.scala#L66

core/src/main/scala/stainless/extraction/trace/Trace.scala
```diff
@@ -0,0 +1,156 @@
...
+    val postcondition = function.postcondition
+ 
+    val bodyPre = exprOps.withPrecondition(body, precondition)
+    //val bodyPost = exprOps.withPostcondition(bodyPre,postcondition)
+
+    function.copy(function.id, function.tparams, function.params, function.returnType, bodyPre, function.flags)
+  }
+ 
+
+  private def replaceFromIDs(substs: Map[Variable, Expr], expr: Expr) : Expr = {
```
romac, line 122: This is `exprOps.replaceFromSymbols`, see https://github.com/epfl-lara/inox/blob/172933e5138925ee229cb03fc7d5297092800e2a/src/main/scala/inox/ast/ExprOps.scala#L41.

romac:
For larabot to test your PR, you need to open and merge another PR which adds your GitHub username to the `trusted` array in https://github.com/epfl-lara/stainless/blob/master/.larabot.conf#L15. Then you will likely have to push a new commit to trigger the build.

PR #746, opened 2020-05-12
Add noDuplicate to ListOps and Set.toList contract

OStevan:


romac:
@vkuncak Should we perhaps even remove these functions from the library, since they require universally quantified postconditions, and leave it to users to define those if they really need it?

In any case, I am going to merge this as is for now, since those methods were already problematic before this PR.

vkuncak:
Yes, let's merge this for now and then later discuss strategy regarding such automatic 'forall's

PR #756, opened 2020-05-29
Error reporting

jad-hamza:
* Add an exception handler that hides the stack trace and prints in the file `stainless-debug.txt`
* Use this exception handler in the Stainless Plugin
* Add an option to print the stack trace when there is an error
* Minor error messages changes

### DIFF
--- a/core/src/main/scala/stainless/MainHelpers.scala
+++ b/core/src/main/scala/stainless/MainHelpers.scala
@@ -187,13 +187,6 @@ trait MainHelpers extends inox.MainHelpers { self =>
           compiler = newCompiler()
       }
 
-      def regularRunCycle() = try {
-        baseRunCycle()
-      } catch {
-        case e: inox.FatalError => throw e
-        case e: Throwable => reporter.internalError(e)
-      }
-
       val watchMode = isWatchModeOn(ctx)
       if (watchMode) {
         val files: Set[File] = compiler.sources.toSet map {
@@ -204,7 +197,7 @@ trait MainHelpers extends inox.MainHelpers { self =>
         watchRunCycle() // first run
         watcher.run()   // subsequent runs on changes
       } else {
-        regularRunCycle()
+        baseRunCycle()
       }
 
       // Export final results to JSON if asked to.
--- a/core/src/main/scala/stainless/MainHelpers.scala
+++ b/core/src/main/scala/stainless/MainHelpers.scala
@@ -48,6 +48,7 @@ trait MainHelpers extends inox.MainHelpers { self =>
     optWatch -> Description(General, "Re-run stainless upon file changes"),
     optCompact -> Description(General, "Print only invalid elements of summaries"),
     optNoColors -> Description(General, "Disable colored output"),
+    frontend.optPrintStackTrace -> Description(General, "Print stack trace when an exception occurs"),
     frontend.optPersistentCache -> Description(General, "Enable caching of program extraction & analysis"),
     frontend.optBatchedProgram -> Description(General, "Process the whole program together, skip dependency analysis"),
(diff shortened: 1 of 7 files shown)


core/src/main/scala/stainless/MainHelpers.scala
```diff
@@ -141,81 +142,91 @@ trait MainHelpers extends inox.MainHelpers { self =>
...
-
-    if (ctx.options.findOptionOrDefault(optVersion)) {
-      displayVersion(ctx.reporter)
-      System.exit(0)
+  def main(args: Array[String]): Unit = {
+    implicit val ctx: inox.Context = try {
+      setup(args)
+    } catch {
+      case _: Throwable =>
+        println("There was an exception while parsing command-line options")
```
romac, line 150: This should probably print to stderr instead of stdout.

romac: Or perhaps just use `sys.error` instead of `println` + `System.exit`.

frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
```diff
@@ -279,7 +279,7 @@ class CodeExtraction(inoxCtx: inox.Context, cache: SymbolsContext)(implicit val
         outOfSubsetError(t, "Mutable fields in static containers such as objects are not supported")
 
       case other =>
-        reporter.warning(other.pos, "Could not extract tree in static container: " + other)
+        reporter.warning(other.pos, s"Stainless does not support `$other` in static containers")
```
romac, line 282: If `$other` is a large tree then the output might not be very readable since the error message will be split in two by a tree potentially spanning multiple lines. Might thus be better to keep the offending tree at the end of the error message.

frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
```diff
@@ -510,7 +510,7 @@ class CodeExtraction(inoxCtx: inox.Context, cache: SymbolsContext)(implicit val
         // ignore
 
       case other =>
-        reporter.warning(other.pos, "Could not extract tree in class: " + other)
+        reporter.warning(other.pos, s"Stainless does not support `$other` in class $id")
```
romac, line 513: Same here.

frontends/scalac/src/main/scala/stainless/frontends/scalac/CodeExtraction.scala
```diff
@@ -297,7 +297,7 @@ trait CodeExtraction extends ASTExtractors {
         outOfSubsetError(t, "Mutable fields in static containers such as objects are not supported")
 
       case other =>
-        reporter.warning(other.pos, "Could not extract tree in static container: " + other)
+        reporter.warning(other.pos, s"Stainless does not support `$other` in static containers")
```
romac, line 300: Same here.

frontends/scalac/src/main/scala/stainless/frontends/scalac/CodeExtraction.scala
```diff
@@ -474,7 +474,7 @@ trait CodeExtraction extends ASTExtractors {
         // ignore
 
       case other =>
-        reporter.warning(other.pos, "Could not extract tree in class: " + other + " (" + other.getClass + ")")
+        reporter.warning(other.pos, s"Stainless does not support `$other` in class $id")
```
romac, line 477: Same here.

jad-hamza:
Thanks for the feedback! I also removed `regularRunCycle` (which was calling `internalError` and displaying stack traces), and I added a call to the error handler for options parsing exceptions

PR #768, opened 2020-06-03
Add ListMap implementation by @jad-hamza

romac:
This PR depends on #746


jad-hamza:
Trying to debug what's going on, reopened here https://github.com/epfl-lara/stainless/pull/794

PR #781, opened 2020-06-12
Rearrange debugging options

jad-hamza:
* Move printing of symbols to `--debug=frontend`
* Move printing of trace to `--debug=frontend`
* Move `stainless-debug.txt` to `stainless-stack-trace.txt` (do you think we should even keep this file, since we have `--debug=frontend`?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,7 +38,7 @@ pri-sessions/
 stainless.out/
 repairs.dat
 tmp
-stainless-debug.txt
+stainless-stack-trace.txt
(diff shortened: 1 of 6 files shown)


vkuncak:
Can we also extend section 

https://epfl-lara.github.io/stainless/wrap.html

with a subsection to add some advice on how to adapt Scala code to work with stainless, including things that would have been useful to @saraswat ? This can be e.g. FAQ at the end of that section.

PR #788, opened 2020-06-17
Recommend using Z3 4.8.8

jad-hamza:
I changed the larabot version to z3 4.8.8 to see if the pull requests is accepted by larabot, will revert to 4.8.6 if it fails.

### DIFF
--- a/bin/package-standalone.sh
+++ b/bin/package-standalone.sh
@@ -14,7 +14,7 @@ if [[ $(git diff --stat) != '' ]]; then
 fi
 
 SCALA_VERSION="2.12"
-Z3_VERSION="4.8.6"
+Z3_VERSION="4.8.8"
 
 SBT_PACKAGE="sbt stainless-scalac-standalone/assembly"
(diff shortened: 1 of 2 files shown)


jad-hamza:
One test failed, I'll have a look. I reverted z3 to 4.8.6 for now on larabot.

jad-hamza:
The test that fails is:
```
sbt it:testOnly *ImperativeSuite* -- -z "imperative/valid/WhileAsFun2"
```

Quite strange because it works with the command-line, and the VC that fails in the sbt test goes through in the command-line (maybe symbols are different).

jad-hamza:
My bad, the example actually also fails with the `stainless-scalac`, the issue is due to: https://github.com/Z3Prover/z3/issues/4539

jad-hamza:
The `WhileAsFun2` example is still unstable but we still upgraded to Z3 4.8.10.
For these examples, we can try the solvers: `smt-z3:z3-4.8.6` (if `z3-4.8.6` is in the path) or `no-inc:smt-z3:z3-4.8.10` to try Z3 4.8.10 in non-incremental mode.

PR #790, opened 2020-06-17
Add drganam to the list of trusted larabot users

drganam:


jad-hamza:
Opened https://github.com/epfl-lara/stainless/pull/791 because I can't merge here with the missing checks

PR #805, opened 2020-07-06
wip: windows attempt fix

jad-hamza:


core/src/main/scala/stainless/frontend/FrontendFactory.scala
```diff
@@ -28,7 +28,7 @@ trait FrontendFactory {
       // JAR URL syntax: jar:<url>!/{filepath}, Expected path syntax: file:/path/a.jar!/{filepath}
       assert(path startsWith "file:")
       val Array(_, filepath) = path split "!/"
-      val filename = filepath.replaceAllLiterally(File.separator, "_")
+      val filename = filepath.replaceAllLiterally(File.separator, "_").replaceAllLiterally("/", "_")
```
mwookawa, line 31: this works!

build.sbt
```diff
@@ -145,7 +145,7 @@ lazy val assemblySettings: Seq[Setting[_]] = {
 lazy val libraryFiles: Seq[(String, File)] = {
   val libFiles = ((root.base / "frontends" / "library") ** "*.scala").get
   val dropCount = (libFiles.head.getPath indexOfSlice "library") + ("library".size + 1 /* for separator */)
-  libFiles.map(file => (file.getPath drop dropCount, file)) // Drop the prefix of the path (i.e. everything before "library")
+  libFiles.map(file => (file.getPath.replaceAll("\\\\", "/") drop dropCount, file)) // Drop the prefix of the path (i.e. everything before "library")
```
mwookawa, line 148: sbt publishLocal now has the following error:

`(stainless-scalac / assembly) deduplicate: different file contents found in the following:
stainless/BuildInfo$.class
stainless/BuildInfo$.class
...`

that said, I can build the jars in linux and (somewhat laboriously, but it works!) copy them onto the windows node and, as per below, the build-time error for other projects are gone.

jad-hamza: Great that this works! Do you get this error with the Mingw-w64 terminal? I'll try to reproduce. Could it be an issue with an unclean build? (`sbt clean`)

mwookawa: i did try cleaning, same error.

i'm getting this with powershell -> sbt universal:stage. i can try with mingw if you can't duplicate

jad-hamza: `universal:stage` worked for me on powershell, but I did get the same error as you (`deduplicate`) for `publishLocal`. You get the same error with `universal:stage` and `publishLocal`?

After a search I found that this is related to `assemblyMergeStrategy`. In our [`build.sbt`](https://github.com/epfl-lara/stainless/blob/0b28e1e05af4b522425cee0d6ce6d0ee2e385b47/build.sbt#L132), it looks like we do set the right strategy for `BuildInfo$.class` but for some reason these settings seem to be ignored.

mwookawa: on my stainless build machine i've resorted to using a linux vm. will try again in powershell asap

PR #809, opened 2020-07-10
Check initialization of fields

jad-hamza:
* Prevent reference to future fields from a field in a class
* Prevent reference to `this` from a field in a class


core/src/main/scala/stainless/extraction/methods/Trees.scala
```diff
@@ -94,6 +94,39 @@ trait Trees extends throwing.Trees { self =>
     override protected def ensureWellFormedClass(cd: ClassDef) = {
       super.ensureWellFormedClass(cd)
 
+      val fids = cd.methods.filter(id => !getFunction(id).getFieldDefPosition.isEmpty)
```
samarion, line 97: Are `cd.methods` ever re-ordered during transformations? I'm surprised you need to store the field position explicitly.

jad-hamza: I'll try :)

Issue #808, opened 2020-07-09
Prevent impure lazy vals in function bodies


jad-hamza:

vkuncak:
Here is a test case:
```scala
  def f(): Int = {
    var counter = 0
    lazy val x = {
      counter = counter + 1
      counter
    }
    counter + x + x
  } ensuring (_ == 2)
```
For now (1) we can reject and report cleanly, since Scala style probably does not rely on local lazy val-s much.
Later (2) as @jad-hamza suggested, we could factor out into a method. We already pull out methods for while loops, so doing it for lazy vals seems fine. It means that if you use a lazy val and the computation is pure, there will be some verification penalty.
Eventually (3) we could optimize some of the lazy vals into vals as long as we are sure there are no side effects and termination still works.

jad-hamza:
(1) implemented by https://github.com/epfl-lara/stainless/pull/814

PR #810, opened 2020-07-13
Check that field initializations do not refer to future fields

jad-hamza:


core/src/main/scala/stainless/extraction/methods/Trees.scala
```diff
@@ -94,6 +94,19 @@ trait Trees extends throwing.Trees { self =>
     override protected def ensureWellFormedClass(cd: ClassDef) = {
       super.ensureWellFormedClass(cd)
 
+      // Check that fields do not refer to future fields
+      val fids = cd.methods.filter(id => getFunction(id).isField)
```
samarion, line 98: Thanks for the alternative implementation!

Just a small nit: you probably need to check that the method isn't a lazy field (since they're allowed to refer to later definitions). I believe the laziness information is stored within the `IsField` annotation.

jad-hamza: Thanks for having a look! Normal field to lazy field is ok as well?

frontends/benchmarks/verification/valid/ConcRope.scala
```diff
@@ -68,32 +68,32 @@ object ConcRope {
...
+              lr.level() > r.level()
             case _ =>
-              l.level > r.level
+              l.level() > r.level()
           })
       case _ => true
     }
 
-    val level: BigInt = {
+    def level(): BigInt = {
```
romac, line 78: Why the additional parens here, in `size`, and in the ConcTree benchmark? Those are usually used in Scala for indicating side effects in methods which don't take any arguments, which is not the case here AFAICT.

jad-hamza: The dependency computation thinks that `level` depends on `size`. Maybe I shouldn't use `dependencies` but something more precise?

core/src/main/scala/stainless/extraction/methods/Trees.scala
samarion: Yes, normal to lazy is also ok. The lazy field is internally compiled as a def that stores its value after computing it the first time (so it won't be null even if defined after the current val).

frontends/benchmarks/verification/valid/ConcRope.scala
samarion: The symbols mix in a CallGraph (https://github.com/epfl-lara/inox/blob/master/src/main/scala/inox/ast/CallGraph.scala). You can use `transitivelyCalls` (or maybe `transitiveCallees`).

romac: Oh I just meant why not write it as `def level: BigInt` instead of `def level(): BigInt`. My (admittedly not so useful) remakes was thus just about style rather than functionality.

romac: > The dependency computation thinks that `level` depends on `size`

Any idea why that's the case? I don't see any (transitive) reference to `size` here, so that seems wrong/weird to me.

samarion: Hmm, this is actually a tricky issue: examples like ConcRope should be legal because they're only accessing the val within sub-fields (not the current instance). But it's not so easy to detect that...

Ideally, the following (very contrived) benchmark should pass the ordering check:
```scala
sealed abstract class StrangeList
case class Nil() extends StrangeList
case class Cons(head: Int, tail: StrangeList) extends StrangeList {
  val sumPlus1 = head + trail.sum + 1
  val sum = head + tail.sum
}
```

core/src/main/scala/stainless/extraction/methods/Trees.scala
```diff
@@ -94,6 +94,19 @@ trait Trees extends throwing.Trees { self =>
     override protected def ensureWellFormedClass(cd: ClassDef) = {
       super.ensureWellFormedClass(cd)
 
+      // Check that fields do not refer to future fields
+      val fids = cd.methods.filter(id => getFunction(id).isField(isLazy = false))
+      for ((fid, future) <- fids.zip(fids.tails.toSeq.tail)) {
```
samarion, line 99: Actually, if a field refers to itself, things are already broken in Scala.

I think the following snippet should be illegal as well (but not 100% sure, might be worth checking):
```scala
case class Test() {
  val zeros = SCons(0, () => zeros)
}
```

frontends/benchmarks/verification/valid/ConcRope.scala
jad-hamza: > Oh I just meant why not write it as `def level: BigInt` instead of `def level(): BigInt`. My (admittedly not so useful) remark was thus just about style rather than functionality.

Ah right thanks


> Hmm, this is actually a tricky issue: examples like ConcRope should be legal because they're only accessing the val within sub-fields (not the current instance). But it's not so easy to detect that...
> 
> Ideally, the following (very contrived) benchmark should pass the ordering check:
> 
> ```scala
> sealed abstract class StrangeList
> case class Nil() extends StrangeList
> case class Cons(head: Int, tail: StrangeList) extends StrangeList {
>   val sumPlus1 = head + trail.sum + 1
>   val sum = head + tail.sum
> }
> ```

I can allow self-reference, so `level` and `size` would be ok here (after changing to `transitiveCallees`).

jad-hamza:
About the discussion from #809 

> Are cd.methods ever re-ordered during transformations? I'm surprised you need to store the field position explicitly.

There was an issue with one of the benchmarks (`LazyNumericalRep`), it looks like we lose ordering because symbols are stored in maps.

frontends/benchmarks/verification/valid/ConcRope.scala
samarion: I think self-reference should also be rejected, I don't think the semantics of defs and vals agree.

jad-hamza: Ok so I'll reject that and keep `level` and `size` as `def` or maybe `lazy val`.

samarion: That would work, if a bit strict :)

samarion:
Oh, indeed we're not storing methods within the `ClassDef`: https://github.com/epfl-lara/stainless/blob/00d9ed3483cbd90fef9ab804ef9480b2f675110b/core/src/main/scala/stainless/extraction/methods/Trees.scala#L214 That's too bad...

In that case, we either need your field index annotation or to re-work the `ClassDef` to contain its methods (instead of relying on an annotation within the `FunDef`).

vkuncak:
Thanks @jad-hamza for your work and @samarion and @romac for looking at it.
It is fine if we are more strict than `scalac` regarding checks.
But we should not do transformation inconsistent with operational semantics of Scala.

jad-hamza:
I added a constraint for `this` which might be a bit restrictive (cannot refer to fields), and that requires changes such as:
https://github.com/epfl-lara/stainless/pull/810/files#diff-fd8151de4214ad6b807e433804c6861eR34
Do you want me to relax it?

vkuncak:
Why size inside List cannot also become lazy val instead of def ?

jad-hamza:
> Why size inside List cannot also become lazy val instead of def ?

I changed them to `lazy val` now, this should work:
https://github.com/epfl-lara/stainless/pull/810/files#diff-e3fc71308237c271fa8cdeb818d34f13R20

vkuncak:
This looks promising. 
Do we reject side effects inside all lazy val's in Stainless? A lazy val could be modelled using a boolean and non-aliased fields, so side effects could be modelled and one could argue they are useful, but the naive transformation into `def` of course does not work for side effects.

vkuncak:
What does the branch after the fix do on this test case?
```scala
  def f(): Int = {
    var counter = 0
    lazy val x = {
      counter = counter + 1
      counter
    }
    counter + x + x
  } ensuring (_ == 2)
```
Dynamically, the assertion holds, but in the master branch it fails because lazy val is treated like a val.

core/src/main/scala/stainless/extraction/methods/Trees.scala
```diff
@@ -94,6 +94,46 @@ trait Trees extends throwing.Trees { self =>
...
+            )
+          else if (position2 > position)
+            throw NotWellFormedException(fd,
+              Some(s"future reference in field ${fd.id} to $fid2 is not supported by Stainless")
+            )
+        }
+
+        // Check that fields do not refer to `this`
+        for (
+          id <- callees + fid;
```
samarion, line 120: This seems way too strict. Since field access is still modeled as `this.field` at this point, you're disallowing all field access here. For example, the benchmark `frontends/benchmarks/imperative/valid/TraitVar2.scala` should be valid wrt future/self field access.

I think the logic needs to rather check that all `This` references in the current function (= `fd`) appear in receiver positions of method calls that only involve previous fields.

samarion:
@Viktor: lazy vals inside function bodies are not extracted correctly, so we lose the information that they were lazy. I don't think Jad's fix will help here, we need to fix CodeExtraction so we can actually differentiate between (local) vals and lazy vals first.

jad-hamza:
> @vkuncak: lazy vals inside function bodies are not extracted correctly, so we lose the information that they were lazy. I don't think Jad's fix will help here, we need to fix CodeExtraction so we can actually differentiate between (local) vals and lazy vals first.

Right. Should I make them local function definitions with `@pure` annotation as you suggested above?

core/src/main/scala/stainless/extraction/methods/Trees.scala
jad-hamza: Ok thanks I'll try this.

vkuncak:
> > @vkuncak: lazy vals inside function bodies are not extracted correctly, so we lose the information that they were lazy. I don't think Jad's fix will help here, we need to fix CodeExtraction so we can actually differentiate between (local) vals and lazy vals first.
> 
> Right. Should I make them local function definitions with `@pure` annotation as you suggested above?

I guess this will be addressed in a separate PR that addresses #808.

jad-hamza:
Ah yes, thanks for pointing to the issue.

jad-hamza:
> I think the logic needs to rather check that all `This` references in the current function (= `fd`) appear in receiver positions of method calls that only involve previous fields.

I've relaxed the checks by allowing fields to use `this.x` as long as `x` is an accessor for the (exact) same class (code here: https://github.com/epfl-lara/stainless/pull/810/files#diff-0d88a35004fc4cab46c0cf7f7964857aR102).
This still prevents such examples: https://github.com/epfl-lara/stainless/blob/2372c03de6e9cba91d267a9e63f1504718ea264e/frontends/benchmarks/typechecker/valid/MicroTests/FieldInheritance.scala

Here is a simplified version:

```scala
import stainless.lang._

object SimpleFieldInheritance {
  sealed abstract class Foo[B] {
    val thisIsIt: BigInt = 1
    val y: BigInt
  }

  case class Bar[X](override val thisIsIt: BigInt) extends Foo[X] {
    val y = thisIsIt
  }

  case class Baz[X]() extends Foo[X] {
    val y = thisIsIt + 1
  }
}
```

Should I relax further to accept this?

samarion:
I think allowing access to fields in super-classes should be ok.

But the abstract `val y: BigInt` is a bit worrying here, see https://github.com/epfl-lara/stainless/issues/813#issuecomment-665211228.

jad-hamza:
If I understand correctly [this file](https://github.com/epfl-lara/stainless/blob/master/frontends/benchmarks/typechecker/valid/MicroTests/FieldInheritance.scala#L14) should be rejected because `y` must be overridden in a constructor parameter. Should the field also be rejected if the `y`'s lines 14 and 18 are lazy, or is that ok?

samarion:
Yes, it should be rejected.
If they're lazy it should be ok (but the abstract declaration needs to be a def as well).

jad-hamza:
To summarize, is this correct?
* val in abstract can only be overridden by field parameters
* def/lazy in abstract can be overridden by def/lazy (and they can be interchanged), or by a field parameter, but not by val

jad-hamza:
larabot is happy now! What do you think?

core/src/main/scala/stainless/extraction/methods/Trees.scala
```diff
@@ -94,18 +94,82 @@ trait Trees extends throwing.Trees { self =>
...
+          if (position2 == position)
+            throw NotWellFormedException(fd,
+              Some(s"self-reference in field `${fd.id}` is not supported by Stainless")
+            )
+          else if (position2 > position)
+            throw NotWellFormedException(fd,
+              Some(s"future reference in field `${fd.id}` to `$fid2` is not supported by Stainless")
+            )
+        }
+      }
```
samarion, line 152: Could we maybe merge the two checks? Something like
```scala
for (fd <- fds) {
  val fid = fd.id
  val position = fd.getFieldDefPosition.get

  exprOps.fold[Unit] {
    // allow access to fields defined previously
    case (MethodInvocation(This(_), xid, _, Seq()), _)
      if fids(xid) && lookupFunction(xid).forall {
        fd => (fd.isAccessor || fd.isField) && fd.getFieldDefPosition.forall(_ < position)
      } =>
    // allways allow access to constructor params
    case (ClassSelector(This(_), xid), _) =>
    // check that methods and functions don't access fields defined previously
    case MethodInvocation(rec, xid, _, _) =>
      if (!(transitiveCallees(xid) & fids).forall(tid => getFunction(tid).getFieldPosition.forall(_ < position)))
        throw Exception
    case FunctionInvocation(xid, _, _) =>
      // same check as above
    case (th @ This(_), _) =>
      throw NotWellFormedException(fd,
        Some(s"field `$id` should only refer to `this` to access previous fields")
      )
    case (_, subs) =>
      val _ = subs.toList // force visit to children
      ()
  }(fd.fullBody)
}
```

This could allow examples like ConcRope where access to current field of child ADT should be ok.
I haven't actually tested the code, so this might not work at all. If it doesn't, let's keep your current solution.

vkuncak:
I second the question of @samarion ; it would be sad if we can no longer implement ConcTree and ConcRope as they represent the most complex data structure we verified (and a new one, accompanied by a recent published paper).

core/src/main/scala/stainless/extraction/methods/Trees.scala
```diff
@@ -94,18 +94,82 @@ trait Trees extends throwing.Trees { self =>
...
       if (!cd.isAbstract) {
         val remainingAbstract = (cd +: cd.ancestors.map(_.cd)).reverse.foldLeft(Set.empty[Symbol]) {
           case (abstractSymbols, acd) =>
-            val concreteSymbols = acd.methods.map(_.symbol).toSet
+            val concreteSymbols = acd.methods
+              .map(id => id.symbol -> getFunction(id))
+              .filter { case (_, fd) => !fd.isAbstract }
+              // fd.getFieldDefPosition is empty for lazy val's, non-empty for val's
+              .filter { case (_, fd) =>
+                (!fd.isAccessor || fd.isAccessorOfParam(acd)) && fd.getFieldDefPosition.isEmpty
```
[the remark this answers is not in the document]
jad-hamza: With scalac in this example:

```scala
object Test {
  abstract class A {
    def x: Int
  }
  case class B() extends A {
    val x: Int = 0
  }
}
```

For the second `x` there is an accessor `x$116` for identifier `x$115`, but `x$115` doesn't appear in the `acd.fields`. `x$115` is a method with `@field` and `@fieldDefPosition(1)`

> Also, can we use `fd.isField(isLazy = false)` instead of `fd.getFieldDefPosition.isEmpty` for clarity? Or is there a difference?

I think that didn't work, maybe there is a bug that needs fixing, I'll check again.

jad-hamza:
> Do the ConcRope and ConcTree examples no longer work with the type checker?

> I second the question of @samarion ; it would be sad if we can no longer implement ConcTree and ConcRope as they represent the most complex data structure we verified (and a new one, accompanied by a recent published paper).

They still work (with the lazy addition). I removed the files from `typechecker/valid` because they were duplicates of `verification/valid`, and the type checker testsuite doesn't look anymore at the `typechecker/valid` folder which is unused. I need to completely remove this folder (while making sure we don't lose benchmarks).

core/src/main/scala/stainless/extraction/methods/Trees.scala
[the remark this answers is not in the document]
jad-hamza: Actually I also need to include non-field (normal) methods here (where `getfieldDefPosition` is empty, but `isField` is false).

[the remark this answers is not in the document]
samarion: Ok. These modifiers are becoming difficult to keep track of :)

core/src/main/scala/stainless/extraction/methods/Trees.scala
jad-hamza: http://laraquad4.epfl.ch:9000/epfl-lara/stainless/build/5226

I tried in this build but this doesn't catch references to `this` through function invocations:
https://github.com/jad-hamza/stainless/blob/initialization2/frontends/benchmarks/extraction/invalid/Initialization4.scala

samarion: Accessing `this` should be ok if we don't try to access any fields.

One thing that is missing though is a recursion in `subs` for the `MethodInvocation` and `FunctionInvocation` cases when no exception was thrown.

samarion: The really dangerous part would be if `Initialization4` was
```scala
object Initialization4 {
  case class NoThis() {
    val nothis = f()
    val g = 1

    def f() = this.g
  }
}
```
or
```scala
object Initialization4 {
  case class NoThis() {
    val nothis = h()

    def f() = this
    def h() = f().nothis
  }
}
```

jad-hamza: Thanks, I'll add those.

> One thing that is missing though is a recursion in `subs` for the `MethodInvocation` and `FunctionInvocation` cases when no exception was thrown.

Right but it's not so easy to add, for instance you don't want to recurse on the head of `subs` for `this.f(arg1, ..., argn)` (normal method invocation), but only on the `subs` corresponding to the arguments. I think I need to distinguish those cases explicitly. Recursing in the head of subs would make the check crash with the last case detecting `This()`.

samarion: Do we actually need this `This()` case? Isn't it enough to check field access (on arbitrary objects)?

By the way, we could maybe allow `MethodInvocation(CaseSelector(This(), fid), xid, _, _))` when `xid` is the current field. This would allow the recursive val definitions used in ConcRope. The use-case is a bit over-specific though.

As a general note, it's probably easier to work with a `TreeTraverser` when you want to control the traversal logic.

jad-hamza: You're right we shouldn't need it, trying without the `This()` case here: http://laraquad4.epfl.ch:9000/epfl-lara/stainless/build/5232

> By the way, we could maybe allow `MethodInvocation(CaseSelector(This(), fid), xid, _, _))` when `xid` is the current field. This would allow the recursive val definitions used in ConcRope. The use-case is a bit over-specific though.

You said in this comment https://github.com/epfl-lara/stainless/pull/810#discussion_r453680685 that self-reference in val was broken?

> As a general note, it's probably easier to work with a `TreeTraverser` when you want to control the traversal logic.

Yes thanks

samarion: Self-reference to the current field is broken, but you can refer to all fields of your parameters since they must be fully constructed before they are given to you.

Anyway, while trying to build an example I realized that we would run into the restriction of vals overriding defs in any useful code so we can just ignore this case.

jad-hamza: The build failed because of a difference between scalac and dotty extraction on this file: https://github.com/epfl-lara/stainless/blob/75260e70d885237a6fee77ebc768cd917a65e34a/frontends/benchmarks/extraction/invalid/Initialization2.scala. I remember we talked about this difference once, should we fix code extraction to make sure they extract to the same trees? Or adapt the code to make up for the difference?

Dotty:
```scala
@field
@fieldDefPosition(0)
@method(Hello$0)
def message$0: Int = this.answer$0 + 3

@field
@fieldDefPosition(1)
@method(Hello$0)
def answer$0: Int = 42
```

Scalac
```scala
@private
@field
@fieldDefPosition(3)
@method(Hello$0)
def answer$1: Int = 42

@accessor(answer$1)
@method(Hello$0)
def answer$0: Int = this.answer$1

@private
@field
@fieldDefPosition(1)
@method(Hello$0)
def message$0: Int = this.answer$0 + 3

@accessor(message$0)
@method(Hello$0)
def message$1: Int = this.message$0
```

jad-hamza: I kept extraction as is for now, and changed the call to `transitiveCallees` to include `xid`. larabot happy again :)
https://github.com/epfl-lara/stainless/pull/810/files#diff-0d88a35004fc4cab46c0cf7f7964857aR118

vkuncak:
Let's merge.

core/src/main/scala/stainless/extraction/methods/Trees.scala
```diff
@@ -94,18 +94,64 @@ trait Trees extends throwing.Trees { self =>
...
+          case (MethodInvocation(rec, xid, _, _), subs) =>
+            val _ = subs.toList // force visit to children
+            for (tid <- ((transitiveCallees(xid) + xid) & fids).find(tid => getFunction(tid).getFieldDefPosition.exists(_ >= position)))
+              throw NotWellFormedException(fd,
+                Some(s"field `$fid` cannot only refer to previous fields, not to `$tid`")
+              )
+
+          case (FunctionInvocation(xid, _, _), subs) =>
+            val _ = subs.toList // force visit to children
+            for (tid <- ((transitiveCallees(xid) + xid) & fids).find(tid => getFunction(tid).getFieldDefPosition.exists(_ >= position)))
```
samarion, line 125: Micro nit: we shouldn't need the `+ xid` here since static functions can't have a field position.

jad-hamza:
Merged, thanks for the comments!

core/src/main/scala/stainless/extraction/methods/Trees.scala
jad-hamza: Fixed, thanks

PR #814, opened 2020-07-29
Prevent lazy val's from having effects

jad-hamza:
Add a `Lazy` flag to `Let` and make sure that such val's do no have side-effects.


jad-hamza:
When I add the `Lazy` flag to `Let`, the variable `x` on this line `https://github.com/epfl-lara/stainless/pull/814/files#diff-49d19b13e605a31214d7285adf0c7befR8` becomes unbound. I get the Inox error "Unkown variables" from https://github.com/epfl-lara/inox/blob/master/src/main/scala/inox/ast/Definitions.scala#L238

What is the file that constructs the `path` which is used in `collectWithPC`?

samarion:
You're adding the flag in the wrong place in CodeExtraction, it should be added [here](https://github.com/epfl-lara/stainless/blob/342342180e1254cbf9068523a06eaa9e18d9d118/frontends/scalac/src/main/scala/stainless/frontends/scalac/CodeExtraction.scala#L817) to make sure the `Variable` and `ValDef` ASTs match up.

jad-hamza:
Thanks!

jad-hamza:
@samarion Does it look good now?

samarion:
Yep, looks good to me.

PR #818, opened 2020-08-03
Fix type encoding erasing error[Nothing] to unit

### LINKED ISSUE #817: Type encoding erases `error[Nothing]` to `()`
In the following example
```scala
import stainless.lang._

object Error {
  def foo(x: Int): Unit = {
    if (x == 0) {
      error[Nothing]("Panic")
    } else {
      ()
    }
  }
}
```
type encoding is triggered, because we instantiated `error` at `Nothing`. After type encoding, that tree has been replaced by just `()`, so we miss the counterexample `x == 0`.

This presumably has something to do with the Unit type, in particular, as I couldn't recreate the same problematic behavior using any other primitive type. The problem also doesn't occur when we explicitly instantiate at the right type (`error[Unit]`).

gsps:
Fixes #817

### DIFF
--- a/core/src/main/scala/stainless/extraction/oo/TypeEncoding.scala
+++ b/core/src/main/scala/stainless/extraction/oo/TypeEncoding.scala
@@ -144,7 +144,7 @@ trait TypeEncoding
     case s.CharType() => getRefField(e, charValue)
     case s.RealType() => getRefField(e, realValue)
     case s.StringType() => getRefField(e, strValue)
-    case s.UnitType() => t.UnitLiteral()
+    case s.UnitType() => let(("u" :: ref.copiedFrom(e)).copiedFrom(e), e) { _ => t.UnitLiteral().copiedFrom(e) }
   }).copiedFrom(e)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/stainless/extraction/oo/TypeEncoding.scala
```diff
@@ -144,7 +144,7 @@ trait TypeEncoding
     case s.CharType() => getRefField(e, charValue)
     case s.RealType() => getRefField(e, realValue)
     case s.StringType() => getRefField(e, strValue)
-    case s.UnitType() => t.UnitLiteral()
+    case s.UnitType() => let(("u" :: ref.copiedFrom(e)).copiedFrom(e), e) { _ => t.UnitLiteral().copiedFrom(e) }.copiedFrom(e)
```
samarion, line 147: Micro nit: the last `copiedFrom` shouldn't be necessary since we're calling `copiedFrom` on the result of the function anyway.

Issue #811, opened 2020-07-16
bug in isMutableClassType leading to an infinite loop for immutable recursive class types


vkuncak:
Thanks to @samarion for diagnosing:
 
an infinite loop involving this code path:
 

[error] java.lang.StackOverflowError

[error] stainless.extraction.imperative.SymbolOps.$anonfun$isMutableClassType$1(SymbolOps.scala:62)

[error] stainless.extraction.imperative.SymbolOps.$anonfun$isMutableClassType$1$adapted(SymbolOps.scala:61)

[error] scala.collection.LinearSeqOptimized.exists(LinearSeqOptimized.scala:95)

[error] scala.collection.LinearSeqOptimized.exists$(LinearSeqOptimized.scala:92)

[error] scala.collection.immutable.List.exists(List.scala:89)

[error] stainless.extraction.imperative.SymbolOps.stainless$extraction$imperative$SymbolOps$$isMutableClassType(SymbolOps.scala:61)

[error] stainless.extraction.imperative.SymbolOps.isMutableType(SymbolOps.scala:78)

[error] stainless.extraction.imperative.SymbolOps.isRealMutableType(SymbolOps.scala:68)

[error] stainless.extraction.imperative.SymbolOps.$anonfun$isMutableClassType$1(SymbolOps.scala:62)

[error] stainless.extraction.imperative.SymbolOps.$anonfun$isMutableClassType$1$adapted(SymbolOps.scala:61)

[error] scala.collection.LinearSeqOptimized.exists(LinearSeqOptimized.scala:95)

[error] scala.collection.LinearSeqOptimized.exists$(LinearSeqOptimized.scala:92)

[error] scala.collection.immutable.List.exists(List.scala:89)

[error] stainless.extraction.imperative.SymbolOps.stainless$extraction$imperative$SymbolOps$$isMutableClassType
>  > Looks like an infinite loop involving this code path.
> 
> 
> Yes, there's a bug in isMutableClassType where it can enter an infinite 
> loop for immutable recursive class types: 
> https://github.com/epfl-lara/stainless/blob/2372c03de6e9cba91d267a9e63f1504718ea264e/core/src/main/scala/stainless/extraction/imperative/SymbolOps.scala#L62
>

jad-hamza:
Fixed in #824

PR #824, opened 2020-08-13
Avoid infinite loop in isMutableClassType

jad-hamza:
Fix for https://github.com/epfl-lara/stainless/issues/811

Is that ok?

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/SymbolOps.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/SymbolOps.scala
@@ -32,11 +32,11 @@ trait SymbolOps extends oo.SymbolOps { self: TypeOps =>
   }
 
   def isMutableType(tpe: Type): Boolean = {
-    isMutableType(tpe, mutableClasses)
+    isMutableType(tpe, mutableClasses, Set())
   }
 
   def isMutableClassType(ct: ClassType): Boolean = {
-    isMutableClassType(ct, mutableClasses)
+    isMutableClassType(ct, mutableClasses, Set())
   }
 
   @inline def mutableClasses: Set[Identifier] = _mutableClasses.get
@@ -48,7 +48,7 @@ trait SymbolOps extends oo.SymbolOps { self: TypeOps =>
     inox.utils.fixpoint[Set[Identifier]] { mutableClasses =>
       mutableClasses ++
       symbols.classes.collect {
-        case (id, cd) if isMutableClassType(cd.typed.toType, mutableClasses) => id
+        case (id, cd) if isMutableClassType(cd.typed.toType, mutableClasses, Set()) => id
       } ++
       mutableClasses.flatMap { id =>
         val cd = symbols.getClass(id)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/stainless/extraction/imperative/SymbolOps.scala
```diff
@@ -57,10 +57,17 @@ trait SymbolOps extends oo.SymbolOps { self: TypeOps =>
     } (initialClasses)
   })
 
+  var visited: Set[Identifier] = Set()
+  def visit(ct: ClassType): Boolean = {
+    val canVisit = !visited(ct.id)
+    visited += ct.id
+    canVisit
+  }
```
samarion, line 65: Seems a bit unpredictable since `isMutableClassType` will be called multiple times during the fixpoint above. Could you instead add a `visited: Set[Identifier]` argument to the mutually recursive functions below?

core/src/main/scala/stainless/extraction/imperative/SymbolOps.scala
```diff
@@ -31,12 +31,12 @@ trait SymbolOps extends oo.SymbolOps { self: TypeOps =>
     classForField(ct, id).flatMap(_.fields.find(_.id == id))
   }
 
-  def isMutableType(tpe: Type): Boolean = {
-    isMutableType(tpe, mutableClasses)
+  def isMutableType(tpe: Type, visitedCt: Set[Identifier] = Set()): Boolean = {
```
samarion, line 34: Do we need the `visited` argument here? I'd keep it out of the public API if possible

core/src/main/scala/stainless/extraction/imperative/SymbolOps.scala
```diff
@@ -31,12 +31,12 @@ trait SymbolOps extends oo.SymbolOps { self: TypeOps =>
...
+    isMutableType(tpe, mutableClasses, visitedCt)
   }
 
-  def isMutableClassType(ct: ClassType): Boolean = {
-    isMutableClassType(ct, mutableClasses)
+  def isMutableClassType(ct: ClassType, visitedCt: Set[Identifier]): Boolean = {
```
samarion, line 38: Same comment as above.

core/src/main/scala/stainless/extraction/imperative/SymbolOps.scala
jad-hamza: No you're right, thanks!

samarion:
Looks good to me!

Could we maybe also add an extraction test?

jad-hamza:
> Could we maybe also add an extraction test?

Do I also need to stop loops in `isMutableADTType`?
Why where "normal" ADTs like lists not looping?
What's the simplest way to make a non-ADT class to make it loop?

samarion:
I think there aren't really any ADTs at that point in the pipeline because `AdtSpecialization` takes place after the imperative phases. But still, it won't hurt to also add the ADT ids to the visited list just to be sure.

Since `AdtSpecialization` hasn't run, you should be able to test the correct behavior by using something like a right-leaning list with mutable fields, something like:
```scala
sealed abstract class RList
case class RCons(tail: RList, var head: BigInt) extends RList
case class RNil() extends RList
```

jad-hamza:
I think there was a reason but I don't remember why we needed to have two different mutability analyses. Should we merge them?

samarion:
Oh no I think it's fine to keep both, just make sure to handle the recursive case in `isMutableADTType`.

I think we may have a few synthetic ADTs at this point because some earlier phases inject them. However, they're probably never mutable nor recursive, so they wouldn't lead to any non-termination in existing tests.

samarion:
Ah, sorry, I misunderstood your question... I had forgotten about the other analyzer...

Yes, merging them (_i.e._ having one extend the other) would make sense I think.

jad-hamza:
> Yes, merging them (_i.e._ having one extend the other) would make sense I think.

Ok I'll keep this in mind thanks. The other one might have the same non-termination issue.

The example you gave doesn't loop because the class `RCons` goes directly in the `initialClasses` and then `RList` goes directly in `mutableClasses` (as parent of `RCons`). Removing the `var` also doesn't make the example loop, but we can have a loop when the type of a field is the same as its class (like the not well-formed `Bad` example I added)

jad-hamza:
The loop can be obtained with this example, but is not a well-formed ADT, so I had to remove it from the valid extractions. I'll merge for now without an example, but if someone thinks of a well-formed example that was triggering the old loop, we can add it.

```scala
case class Bad(tail: Bad)

object Bad {
  def f(s: Bad) = {
    val x = s
  }
}
```

PR #828, opened 2020-08-26
Generalize specification helpers

gsps:
This PR unifies the handling of specification helpers, so it becomes easier to add new kinds of specs.


gsps:
To be done: Investigate why the assertion that ensures we only have one contract of each kind triggers.

core/src/main/scala/stainless/ast/ExprOps.scala
```diff
@@ -46,34 +51,121 @@ trait ExprOps extends inox.ast.ExprOps {
...
+        case Let(_, _, rest) => bodyMissing(rest)
+        case _ => false
+      }
+
+      gatherSpecs(fullBody, Seq.empty) match {
+        case Some((specs, body)) =>
+          val lets = gatherLets(fullBody, Seq.empty)
+          assert(!body.isInstanceOf[Let] || !bodyMissing(body),
+            "Body is missing, but there are let bindings irrelevant to specs")
+          assert(lets.isEmpty || specs.nonEmpty)
```
brunnerant, line 139: I am not sure what this assertion means. Does this means that this function body is not accepted ?
```Scala
def f(): Unit = {
  val x = 1
  ()
}
```

brunnerant (approved):
LGTM, it is way cleaner that way !
If you want, I can try to debug the missing tests so that this PR is ready to be merged (because I will definitely need your nice infrastructure to add the two new contracts for my project).

core/src/main/scala/stainless/ast/ExprOps.scala
gsps: This part is unfortunately rather cryptic, but those particular assertions are just trying to establish some basic sanity:
- When the body is missing (`???` / `NoTree`, as we use for `extern` functions), then there are not let bindings *underneath* specs. E.g., you can't have `{ require(x>0); val y = x; ??? } ensuring { res => ... }`.
- When `lets` (i.e. the lets that are wrapping the specs) is non-empty, then we have some `specs`. (Otherwise those lets might as well be part of `body`.)

gsps:
@jad-hamza Feel free to review superficially, mostly adding you to keep you in the loop on changes to specs infrastructure.

Issue #496, opened 2019-05-17
Cannot find field `x` of class A


jad-hamza:
Watch mode is currently not working very well, for instance on that example, commenting the function `f` during watch mode leads to error

```scala
object WatchMissingField {
  case class A(x: BigInt)
  def f() = () // commenting this line during watch mode leads to an error
}
```

> Cannot find field `x` of class A

PR #829, opened 2020-08-27
Fix missing field error in watch mode

jad-hamza:
Fix https://github.com/epfl-lara/stainless/issues/496


jad-hamza:
@romac I made some minor changes to fix some watch issues
(see also https://github.com/epfl-lara/stainless/pull/830 and https://github.com/epfl-lara/stainless/pull/831)
What do you think?

jad-hamza:
Merging for now, will revert if someone notices an issue

PR #833, opened 2020-09-03
Recommend using CVC4 1.8

jad-hamza:


jad-hamza:
We should try again to switch the binaries on larabot once https://github.com/epfl-lara/inox/pull/132 is merged.

PR #840, opened 2020-09-16
Error reporting for misplaced specifications

### LINKED ISSUE #838: Strange reporting
The combination of these two methods:
```scala
  def onlyBool: Boolean = {
    true
  } ensuring (_ => false)

  def done: Unit = {
    onlyBool
  } ensuring(_ => false)
```
gives a report like this:
```scala
done      body assertion: Postcondition failed @37:5  valid    nativez3  ?:?:?       
onlyBool  postcondition                               invalid  nativez3  ListExt.scala:33:5  
```
In my view, Scala should reject this because the body of `done` should be returning boolean. But, unlike ocaml it does not reject it. What is strange is the message ` Postcondition failed` followed by `valid`. In fact, the VC should be valid (`false => false`), but it
looks like something gets misinterpreted due to the non-sensical types. Maybe we should just reject `done` with type error.

jad-hamza:
Fix #838


jad-hamza:
Merging this for now, @gsps I remember you had a more relaxed transformation during extraction, feel free to revert this if you add yours.

PR #841, opened 2020-09-22
Fix unapplyAccessor not instantiating with refinement type

gsps:
This PR fixes an issue I ran into when trying to add support for some synthesized unapplies. The bug can also be triggered on master though, and I [included a test case](https://github.com/epfl-lara/stainless/compare/fix-unapplyAccessor-inst?expand=1#diff-2160f6987b6b86c1669d115880f761c0) demonstrating the underlying issue through ADT specialization. The suggested fix is minimal, but I'd be happy to hear whether @samarion sees any problem with it.

As to why the given test case fails:

Initially, AdtSpecialization creates an unapply for `B` whose return type is `Option[{v: A | v is C || v is D}]`. However, during `matchToIfThenElse` we end up instantiating `isEmpty` and `get` with `A` rather than the refinement type. That is, we end up with `isEmpty[A](ourUnapplyForB(a))`.

Later, in the type checker (the stainless backend), we try to the check the argument passed in `isEmpty[A](ourUnapplyForB(a))`. For some reason, typer insists on proving *equality* between `Option[{v: A | v is C || v is D}]` (the result of `ourUnapplyForB(a)`) and `Option[A]` (the expected type for isEmpty's argument). (*Edit: Actually, I might only have seen the equality check between the respective type arguments to `Option[_]`. That would make more sense, as our `Option` is invariant.*)

Why did we instantiate `isEmpty` with `A` rather than the refinement type? In [`unapplyAccessor` helper](https://github.com/epfl-lara/stainless/blob/6e421baca37c97b9bd227ecbf0643505b6156292/core/src/main/scala/stainless/ast/Expressions.scala#L169) we call `TypedFunDef#getType`, which returns the non-dependent version of the result type. Replacing it with the original (`TypedFunDef#returnType`) fixes the immediate issue. (Edit: If anything, I should probably also replace the `fd.params.head.getType` by `fd.params.head.tpe`?)

### DIFF
--- a/core/src/main/scala/stainless/ast/Expressions.scala
+++ b/core/src/main/scala/stainless/ast/Expressions.scala
@@ -166,7 +166,7 @@ trait Expressions extends inox.ast.Expressions with Types { self: Trees =>
       .filter(_.params.size == 1)
       .getOrElse(throw extraction.MalformedStainlessCode(up, "Invalid unapply accessor"))
     val unapp = up.getFunction
-    val tpMap = s.instantiation(fd.params.head.getType, unapp.returnType)
+    val tpMap = s.instantiation(fd.params.head.tpe, unapp.returnType)
       .getOrElse(throw extraction.MalformedStainlessCode(up, "Unapply pattern failed type instantiation"))
     fd.typed(fd.typeArgs map tpMap).applied(Seq(unapplied))
   }
--- a/core/src/main/scala/stainless/ast/Expressions.scala
+++ b/core/src/main/scala/stainless/ast/Expressions.scala
@@ -193,7 +193,7 @@ trait Expressions extends inox.ast.Expressions with Types { self: Trees =>
       getUnapplied(unapplyScrut(scrut, this).copiedFrom(this))
 
     def subTypes(in: Type)(implicit s: Symbols): Seq[Type] =
-      unwrapTupleType(s.unapplyAccessorResultType(getGet, getFunction.getType).get, subPatterns.size)
+      unwrapTupleType(s.unapplyAccessorResultType(getGet, getFunction.returnType).get, subPatterns.size)
   }
 
   /** Symbolic I/O examples as a match/case.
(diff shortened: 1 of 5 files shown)


core/src/main/scala/stainless/ast/Expressions.scala
```diff
@@ -166,7 +166,7 @@ trait Expressions extends inox.ast.Expressions with Types { self: Trees =>
       .filter(_.params.size == 1)
       .getOrElse(throw extraction.MalformedStainlessCode(up, "Invalid unapply accessor"))
     val unapp = up.getFunction
-    val tpMap = s.instantiation(fd.params.head.getType, unapp.getType)
+    val tpMap = s.instantiation(fd.params.head.tpe, unapp.returnType)
```
samarion, line 169: Do we possibly have the same issue in the `subTypes` method below?

samarion (approved):

core/src/main/scala/stainless/ast/Expressions.scala
gsps: Good catch, there seem to be a few related crashes in the CI run. Let's see if that fixes it.

gsps:
Turns out there's another aspect that we have to address: `SyntheticSorts` (which contributes `Option` when it has been pruned due to lack of code explicitly depending on it) misbehaves when used in `AdtSpecialization`. Namely, it doesn't pick up the library-provided `Option` when present, since the latter is still a class at that point (whereas `SyntheticSorts` only looks for the corresponding ADT). After a discussion with Nicolas the conclusion was that we could either add specialized versions of `SyntheticSorts` pre/post ADTSpecialization. We might also need some tweaks to make it work in AdtSpecialization itself (which is at the boundary). Another solution would be to always inject an Option early in the pipeline and never prune it after that.

Here's an adapted test case that triggers the behavior (resulting in the pattern not type-checking anymore, since it mixes the two `Option` types):
```scala
import stainless.lang.Option

object AdtSpecializationUnapply {
  sealed trait Root
  sealed trait B extends Root { val x: BigInt }
  case class C(x: BigInt) extends B
  case class D(x: BigInt) extends B

  // Ensures that the stainless library Option will be retained
  def keepIsEmpty(o: Option[BigInt]) =
    o.isEmpty

  def f(a: Root): BigInt =
    a match {
      // Will be translated to an unapply that essentially checks `b is C || b is D`
      case b: B => b.x
      case _ => 0
    }
}
```

gsps:
@samarion LGTY?

samarion:
Other than the comment in `patternIsTyped`, LGTM. Thanks for fixing this!

gsps:
Sure, thanks for the help!

PR #844, opened 2020-09-24
Fixed errors in docs

Yawolf:
Hello! I found a couple of things that can be fixed in laws docs:
1) In the _"typeclasses"_ section, there was a `:::`, so on the site `::` is seeing. Changing it for `::` should be enough.
2) In the _"under the hood"_ section, the `law_leftIdentity` is defined twice, being one of them the `law_rightIdentity` implementation


jad-hamza:
Merged, thanks!

PR #862, opened 2020-11-23
Add a flag to chose between the current and the full imperative phase

brunnerant:
With the development of the new imperative phase, we want to start writing some benchmarks, while keeping the old ones. To that end, this PR adds the flag `--full-imperative=[true|false]` to switch to the new implementation (by default the old implementation is used, since the new one is WIP). A new test suite was also added to start writing some serious benchmarks for the new phase.

For now, the new benchmark fails, but I don't know why. When running the benchmark manually though, with the identical options, it succeeds, which I do not currently understand.


brunnerant:
Part of the issue of tests failing is due to [this code](https://github.com/epfl-lara/stainless/blob/explicit-heap-imperative/core/src/main/scala/stainless/extraction/imperative/Trees.scala#L139-L154), where the cached definitions stay across runs, causing stale information to be returned in the next runs. For now I have removed caching altogether, which is not ideal, but this allows me to investigate further.

vkuncak:
I suggest the name `full-imperative` instead of `new` imperative. New things soon stop being new.

brunnerant:
It seems that `ComponentTestSuite` is extracting more stuff than when running stainless with `run`. For that reason, the tests fail, because the `copy` method on case classes is extracted and that involves allocation, which our phase does not currently support. I will see if I should pause this PR for now and start handling allocation, or if I can find a temporary workaround.

brunnerant:
I have made a temporary fix to the full imperative test suite, until we add support for allocation. I simply filtered out the `copy` and `copy$default` methods from the functions to be verified in the test. Once allocation is properly handle, we have to remember to revert those temporary changes.

gsps:
Merging this as is, so we can play with the CI setup on `explicit-heap-imperative`.

PR #879, opened 2020-12-12
Frontend for bitvectors

jad-hamza:
@vkuncak @gsps @samarion Any comment in addition to what we discussed?


frontends/scalac/src/main/scala/stainless/frontends/scalac/ASTExtractors.scala
```diff
@@ -433,6 +439,94 @@ trait ASTExtractors {
       }
     }
 
+    object FrontendBVType {
+      def unapply(tpe: Type): Option[(Boolean, Int)] = tpe match {
+        case TypeRef(_, sym, FrontendBVKind(signed, size) :: Nil) if isBVSym(sym) =>
+          Some((signed, size))
+        case TypeRef(_, sym, Nil) if isBVSym(sym) =>
+          val R = """type (UInt|Int)(\d+)""".r
```
gsps, line 447: You could avoid recreating this regex on every unapply call.

frontends/scalac/src/main/scala/stainless/frontends/scalac/ASTExtractors.scala
```diff
@@ -433,6 +439,94 @@ trait ASTExtractors {
...
+          None
+      }
+
+      def unapply(tr: Tree): Option[(Boolean, Int)] = unapply(tr.tpe)
+    }
+
+    object FrontendBVKind {
+      def unapply(tpe: Type): Option[(Boolean, Int)] = tpe match {
+        case SingleType(_, sym) =>
+          val R = """object ([ui])(\d+)""".r
```
gsps, line 462: Same here.

frontends/scalac/src/main/scala/stainless/frontends/scalac/ASTExtractors.scala
gsps: This will probably be called quite often from `extractType`, since the vast majority of types will be `TypeRef`s, I think. Not sure if it actually makes a difference, but for `TupleSymbol` we have a dedicated cache. That being said, I'm not sure the way we're caching there is ideal either. We also store all the negative results, which means that map is going to be very large with a super low hit rate. Ideally we'd actually enumerate all relevant symbols upfront and put them in the map, then only check for membership during `CodeExtraction`.

gsps (commented):
A few notes on the BVType extractors.

jad-hamza:
Thanks, I've moved the regular expressions outside the unapply calls.

jad-hamza:
I've added `widen` and `narrow` operations, and some more examples. There's still no implementation (for Scala compilation), but verification seems to work fine. I'll merge if there is no objection (and if bot is happy)!

frontends/scalac/src/main/scala/stainless/frontends/scalac/CodeExtraction.scala
```diff
@@ -1518,22 +1518,36 @@ trait CodeExtraction extends ASTExtractors {
...
-          case (StrictBVType(_, _), ">=",     Seq(rhs)) => xt.GreaterEquals(extractTree(lhs), extractTree(rhs))
-          case (StrictBVType(_, _), "<",      Seq(rhs)) => xt.LessThan(extractTree(lhs), extractTree(rhs))
-          case (StrictBVType(_, _), "<=",     Seq(rhs)) => xt.LessEquals(extractTree(lhs), extractTree(rhs))
-          case (StrictBVType(_, _), "|",      Seq(rhs)) => xt.BVOr(extractTree(lhs), extractTree(rhs))
-          case (StrictBVType(_, _), "&",      Seq(rhs)) => xt.BVAnd(extractTree(lhs), extractTree(rhs))
-          case (StrictBVType(_, _), "^",      Seq(rhs)) => xt.BVXor(extractTree(lhs), extractTree(rhs))
-          case (StrictBVType(_, _), "<<",     Seq(rhs)) => xt.BVShiftLeft(extractTree(lhs), extractTree(rhs))
-          case (StrictBVType(_, _), ">>",     Seq(rhs)) => xt.BVAShiftRight(extractTree(lhs), extractTree(rhs))
-          case (StrictBVType(_, _), ">>>",    Seq(rhs)) => xt.BVLShiftRight(extractTree(lhs), extractTree(rhs))
+          case (StrictBVType(_, _), "unary_-",  Seq())    => xt.UMinus(extractTree(lhs))
```
gsps, line 1521: Not sure this new case is actually reachable. Might be [caught here](https://github.com/epfl-lara/stainless/blob/a42b82bf60447f2cb960ea503c21685876761628/frontends/scalac/src/main/scala/stainless/frontends/scalac/CodeExtraction.scala#L1236) before?

gsps:
:+1: apart from the small comment

frontends/scalac/src/main/scala/stainless/frontends/scalac/CodeExtraction.scala
jad-hamza: No it was missed, but perhaps I could adjust the extractor so that it can catch it.

jad-hamza:
Merged, thanks!

PR #892, opened 2021-01-26
Minor type-checker improvements

jad-hamza:
* More text in tooltips for the HTML derivation
* Safer VCKind assignment (we reset the VCKind when entering the value side of a `Let`)
* Don't generate VCs at all when all conjuncts are marked `@unchecked` (previously a `true` VC was generated)

### DIFF
--- a/core/src/main/scala/stainless/verification/TypeChecker.scala
+++ b/core/src/main/scala/stainless/verification/TypeChecker.scala
@@ -635,7 +635,7 @@ trait TypeChecker {
       case IfExpr(b, e1, e2) =>
         val (tpe1, tr1) = inferType(tc.withTruth(b).setPos(e1), e1)
         val (tpe2, tr2) = inferType(tc.withTruth(Not(b)).setPos(e2), e2)
-        (ite(b, tpe1, tpe2), checkType(tc.setPos(b), b, BooleanType()) ++ tr1 ++ tr2)
+        (ite(b, tpe1, tpe2), checkType(tc.setPos(b).withVCKind(VCKind.CheckType), b, BooleanType()) ++ tr1 ++ tr2)
 
       case Error(tpe, descr) =>
         val tr = isType(tc, tpe)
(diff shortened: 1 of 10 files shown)


core/src/main/scala/stainless/verification/TypeChecker.scala
```diff
@@ -699,7 +699,7 @@ trait TypeChecker {
         }
 
       case Let(vd, value, body) =>
-        val trValue = checkType(tc.setPos(value), value, vd.tpe)
+        val trValue = checkType(tc.setPos(value).withVCKind(VCKind.NoKind), value, vd.tpe)
```
samarion, line 702: We could have `VCKind.Let` for this case, it's a reasonable VC check no?

jad-hamza: You would like this to be assigned to a VC generated for the refinement of a `Let` type annotation? (if `vd.tpe` is a refinement type)

When we build a VC on this line, we inherit from the previously set `VCKind`:
https://github.com/epfl-lara/stainless/blob/1892fac574110f1074d2ba60791fb24151d89d34/core/src/main/scala/stainless/verification/TypeChecker.scala#L1038

We could reset to `NoKind` when checking `checkType(tc, e, BooleanType())` here?

jad-hamza: I guess there are several other places where we should reset as well, for instance when entering the condition of an `if then else` or the expression of a `match`.

samarion: Hmm, that's a good point. Maybe we should just have a general `CheckType` VC? Resetting to `NoKind` makes it seem like an error in the tool rather than a deliberate context choice.

jad-hamza: Sure, and I'll use it on `Let` values, `if` conditions, expressions of `match`'s, and the `checkType` of L1038 above? With a generic "type-checking" message for the VC description displayed to the user?

samarion: Yes, I think that would be great. Thanks!

jad-hamza: Done thanks!

Issue #893, opened 2021-01-27
Bug in FunctionClosure phase?


jad-hamza:
```scala
import stainless.lang._
import stainless.math._

import SortedArray._

object SortedArray {
  val sizeLimit: Int = 1000000
}

case class C[T](array: Array[T], lessThan: (T, T) => Boolean) {
  require(array.size < sizeLimit)

  def insert(elem: T): Unit = {
    var l: Int = -1
    var h: Int = array.size
    val n: Int = array.size

    (while (l + 1 != h) {
      decreases(max(0, h - l))

      val m = (l + h) >> 1

      if (lessThan(elem, array(m))) {
        h = m
      } else {
        l = m
      }

    }) invariant (-1 <= l && l < h && h <= n)

    var i = n - 1
    while (i > h) {
      array(i) = array(i-1)
      i = i - 1
    }

  }
}
```

Running with:
```
stainless-scalac --solvers=smt-z3 --timeout=10 --watch --debug=stack,trees --debug-objects=insert --infer-measures=false --check-measures=false --print-ids File.scala
```

I get the output:

```scala
[ Debug  ] Symbols before FunctionClosure
[ Debug  ] 
[ Debug  ] -------------Functions-------------
[ Debug  ] def insert$7[T$435](thiss$88: C$42[T$435], elem$33: T$435): (Unit, C$42[T$435]) = {
[ Debug  ]   val thiss$90: C$42[T$435] = thiss$88
[ Debug  ]   val l$340: Int = -1
[ Debug  ]   val h$257: Int = thiss$90.array$42.length
[ Debug  ]   val n$85: Int = thiss$90.array$42.length
[ Debug  ]   def insertWhile$6(l$341: Int, h$258: Int, thiss$91: C$42[T$435]): (Unit, Int, Int, C$42[T$435]) = {
[ Debug  ]     require({
[ Debug  ]       val t$305: Boolean = if (-1 <= l$341) {
[ Debug  ]         l$341 < h$258
[ Debug  ]       } else {
[ Debug  ]         false
[ Debug  ]       }
[ Debug  ]       val res$435: Boolean = t$305
[ Debug  ]       val t$304: Boolean = if (res$435) {
[ Debug  ]         val t$303: Boolean = if (h$258 <= n$85) {
[ Debug  ]           l$341 + 1 ≠ h$258
[ Debug  ]         } else {
[ Debug  ]           false
[ Debug  ]         }
[...]

PR #894, opened 2021-01-30
Fix 893 + minor fixes

jad-hamza:
* Fix issue #893 (wrong type parameters substitutions in function closures)     
* Remove Let case in withPath to avoid introducing unnecessary specs below lets
* Fix issue where the same unique identifier appears twice in a formula (because of shared lets in specs)

### DIFF
--- a/core/src/main/scala/stainless/ast/SymbolOps.scala
+++ b/core/src/main/scala/stainless/ast/SymbolOps.scala
@@ -341,7 +341,6 @@ trait SymbolOps extends inox.ast.SymbolOps { self: TypeOps =>
     }
 
     expr match {
-      case Let(i, e, b) => withPath(b, path withBinding (i -> e))
       case Require(pre, b) => withShared(path, Seq(b, pre), spec, expr.getPos)
       case Ensuring(Require(pre, b), post) => withShared(path, Seq(b, pre, post), spec, expr.getPos)
       case Ensuring(b, post) => withShared(path, Seq(b, BooleanLiteral(true).copiedFrom(expr), post), spec, expr.getPos)
--- a/core/src/main/scala/stainless/ast/SymbolOps.scala
+++ b/core/src/main/scala/stainless/ast/SymbolOps.scala
@@ -330,7 +330,7 @@ trait SymbolOps extends inox.ast.SymbolOps { self: TypeOps =>
       val speccedWithPre =
         specced.withSpec(
           specced.getSpec(PreconditionKind).map(pre =>
-            Precondition(And(pre.expr, cond).setPos(pre.expr))
+            Precondition(And(cond, pre.expr).setPos(pre.expr))
           ).getOrElse(
             Precondition(cond)
           )
--- a/core/src/main/scala/stainless/ast/SymbolOps.scala
+++ b/core/src/main/scala/stainless/ast/SymbolOps.scala
@@ -325,27 +325,18 @@ trait SymbolOps extends inox.ast.SymbolOps { self: TypeOps =>
     * @see [[Expressions.Require]]
     */
   def withPath(expr: Expr, path: Path): Expr = {
-    def unwrap(e: Expr): Lambda = e match {
-      case Let(i, e, b) =>
-        val Lambda(args, body) = unwrap(b)
-        Lambda(args, Let(i, e, body)).copiedFrom(e)
-      case l: Lambda => l
-      case _ => scala.sys.error("Should never happen!")
-    }
-
-    def spec(cond: Expr, es: Seq[Expr]): Expr = es match {
-      case Seq(e) => Require(cond, e).copiedFrom(cond)
-      case Seq(e, pre) => Require(And(cond, pre).copiedFrom(cond), e).copiedFrom(cond)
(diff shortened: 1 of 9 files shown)


jad-hamza:
@samarion Does that look ok?

core/src/main/scala/stainless/extraction/innerfuns/FunctionClosure.scala
```diff
@@ -52,36 +52,49 @@ trait FunctionClosure
       val tpFresh = outer.tparams map { _.freshen }
       val tparamsMap = outer.typeArgs.zip(tpFresh map {_.tp}).toMap
 
+      val oldInst = new typeOps.TypeInstantiator(tparams.map(td => td.tp -> td.tp).toMap)
```
samarion, line 55: Isn't this a no-op transformer?

core/src/main/scala/stainless/ast/SymbolOps.scala
```diff
@@ -341,7 +341,6 @@ trait SymbolOps extends inox.ast.SymbolOps { self: TypeOps =>
     }
 
     expr match {
-      case Let(i, e, b) => withPath(b, path withBinding (i -> e))
```
samarion, line 344: Is the `withPath` method actually compatible with Georg's new spec handling? Should the logic be updated?

samarion (approved):
The change looks reasonable to me. Could you maybe add a test for the FunctionClosure fix?

core/src/main/scala/stainless/extraction/innerfuns/FunctionClosure.scala
jad-hamza: Yes, thanks!

core/src/main/scala/stainless/ast/SymbolOps.scala
jad-hamza: I'm not sure, `withPath` seems to be used only there, and it seems to be integrating the path in the pre/body/post of the inner function using some sharing for the initial let's. I could remove `withPath` completely and integrate the path manually in the specs and body, but I might introduce new bugs :)

core/src/main/scala/stainless/termination/Strenghtener.scala
```diff
@@ -76,6 +82,9 @@ trait Strengthener { self: OrderingRelation =>
         // @nv: one must also check that variablesOf(formula) is non-empty as
         //      we may proceed to invalid strenghtening otherwise
 
+        println("THE FORMULA")
```
vkuncak: Is this println always active? Is it leftover debug info?

jad-hamza: It was leftover, it's removed now, thanks!

core/src/main/scala/stainless/ast/SymbolOps.scala
samarion: Hmm, I think it would probably be worth trying to use `BodyWithSpecs` in `withPath`. Otherwise, the function closure transformation won't play nice with other specifications (unless they're luckily in the right order).

Maybe something like:
```scala
def withPath(expr: Expr, path: Path): Expr = {
  withShared(path, Seq(expr), { case (cond, Seq(e)) =>
    val specced = BodyWithSpecs(e)
    val speccedWithPre = ... // insert cond into the precondition if it exists or add one otherwise
    speccedWithPre.reconstructed
  })
}
```

jad-hamza: Added, thanks!

jad-hamza:
> The change looks reasonable to me. Could you maybe add a test for the FunctionClosure fix?

Thanks! I've added the example (I had to adapt several parts of the pipeline to the `BodyWithSpecs` API)

jad-hamza:
One benchmark broke, not sure why, but this seems to fix it: https://github.com/epfl-lara/stainless/pull/894/commits/20267936342b577c75be1e2ee7cc5ad4455884e7
Maybe a change of behavior between old and new `withPath`?

samarion:
Could you maybe compare the old and new trees to make sure the difference in behavior makes sense? It could have been working due to a bug :)

jad-hamza:
Good idea, I looked at the trees and something seemed out of order. I think I had an ordering error in `withPath` (fixed now).

core/src/main/scala/stainless/extraction/innerfuns/FunctionClosure.scala
```diff
@@ -55,33 +55,45 @@ trait FunctionClosure
...
         }
 
       val freeMap = freshVals.toMap
       val freshParams = freshVals.filterNot(p => reqPC.bindings exists (_._1.id == p._1.id)).map(_._2)
 
-      val instBody = inst.transform(withPath(fullBody, reqPC))
+      val oldBody = withPath(fullBody, reqPC)
+
+      object bodyTransformer extends SelfTreeTransformer {
+        override val s: self.s.type = self.s
```
samarion, line 71: Nit: is this needed? IIRC the `SelfTreeTransformer` actually expands to `s.SelfTreeTransformer` so it should already have the right tree type.

samarion (approved):
Other than the small nit, LGTM. Thanks!

PR #895, opened 2021-02-02
Use Enter key to force watch mode to reload

jad-hamza:
I'm trying to make it so that pressing Enter force reload in `watch` mode (`watch` mode doesn't seem to work on some file systems, this needs to be fixed separately)

I guess that will only work when parallelism is enabled? Should I pass directly the `multiThreadedExecutionContext` (even when parallelism is disabled?)


jad-hamza:
This seems to work (I tested the multi-threaded context which works even when `useParallelism` is `false`),so I'll merge.
There is a strange delay when I press `Enter` for the first time, but that's good enough for now.

Issue #572, opened 2019-06-28
Aliasing and accessors


jad-hamza:
Reported by @samarion in https://github.com/epfl-lara/stainless/pull/526#issuecomment-500335792

```scala
import stainless.lang._

case class A(var x: BigInt)
trait B { var a: A }

object Test {
  def test(b: B): Boolean = {
    val a = b.a
    a.x += 1
    b.a.x += 1
    a == b.a
  }.holds
}
```

Related to:
* [Disable aliasing in lets PR (not merged)](https://github.com/epfl-lara/stainless/pull/541)
* https://github.com/epfl-lara/stainless/issues/413

PR #915, opened 2021-02-24
Prevent aliasing in val's

jad-hamza:
Fix for https://github.com/epfl-lara/stainless/issues/572

Just noticed my PR is similar to my previous (unmerged) PR https://github.com/epfl-lara/stainless/pull/541 where @samarion said:
> This looks too strict to me (although it's a valid fix).
> The ideal fix would be to correctly compute the aliased fields that are returned by accessors. Didn't you start working on that at some point @romac?

What should we do?
Can you give more details on how to compute those aliased fields?
After a valid transformation, what should the function from issue #572 look like?


samarion:
I think @romac actually solved the accessor question since `getEffects` inlines non-recursive functions.

@gsps noticed that `getEffects` and `isExpressionFresh` are not always consistent, but they should be (`getEffects(e).isEmpty` iff `isExpressionFresh(e)`). Would it be possible to re-define `isExpressionFresh` in terms of `getEffects`?

Also, it would probably be worth investigating whether `getExactEffects` and `getKnownEffects` actually make sense. They seem a bit suspicious to me.

jad-hamza:
See alternative fix in #920

PR #918, opened 2021-02-24
Add map duplicate and snapshot as fresh expressions, and remove accessors calls

jad-hamza:


jad-hamza:
Do the changes for accessors, map duplicate and snapshot make sense?

PR #920, opened 2021-02-25
Allow aliasing when the first alias is not used

jad-hamza:
Contains PRs #918 and #919, and a change that saves us from computing effects in an let-bound expression when its free variables are disjoint from the free variables of the rest of the `let`.

There is also another change for computing when `Let` or `LetVar` expressions are fresh.

This is for the new example `VarAliasing.scala`.

Any thoughts?


core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -424,7 +425,10 @@ trait EffectsAnalyzer extends oo.CachingPhase {
 
     def rec(expr: Expr, env: Map[Variable, Effect]): Set[Effect] = expr match {
       case Let(vd, e, b) if symbols.isMutableType(vd.tpe) =>
-        rec(e, env) ++ rec(b, env ++ effect(e, env).map(vd.toVariable -> _))
+        if ((variablesOf(e) & variablesOf(b)).isEmpty)
```
samarion, line 428: I don't understand this change. How does this resolve the effect on `b.x` in
```scala
val a = b // b is some mutable variable in context
a.x += 1
```

jad-hamza: You're right, this doesn't make sense.
I was trying to make that example go through (which is one way to make the Tokenizer example go through):

https://github.com/epfl-lara/stainless/pull/920/files#diff-fec2f03f53fbb95f9831bcdc5e0dfd9d51d23318f9e1a632099f06c04e81ae04

I wanted to handle the case in `val a = A(data)` where `data` is no longer used afterwards.

On master this fails because `effect()` fails on the expression `A(data)`.

samarion: Hmm, I think the `Effect` abstraction can't be used to represent `A(data)`.

Maybe what you could do when `(variablesOf(e) & variablesOf(b)).forall(v => !isMutableType(v.tpe))` is call `rec(b, env)` and then apply the effects on `b`.

jad-hamza: What do you mean "then apply the effects on `b`"? Do you still want to call `effect(e, env)`?

jad-hamza: Oh did you mean changing the effects that we get from `rec(b, env)` to make them apply to the free variables of `e`? (which is just `b` in the example?).

In that example it doesn't work, since `rec(b, env)` is empty given how the function is written for the other cases (it's non-empty if you enrich the environment as in the original code).

samarion: > What do you mean "then apply the effects on b"? Do you still want to call effect(e, env)?

No, `effect` won't have the correct behavior. I mean something like `Path.on` but more strict. Something like:
```scala
def applyEffect(env: Map[Variable, Effect], accessors: Seq[Accessor], expr: Expr): Set[Effect] = (accessors, expr) match {
  case (ADTFieldAccessor(fid) +: rest, ADT(id, _, args)) =>
    applyEffect(rest, args(symbols.getConstructor(id).fields.indexWhere(_.id == fid)))
  // Handle other accessor / expression combinations for which precise analysis is possible
  case (_, v: Variable) =>
    // I'm not sure `inEnv` is the right behavior here, but it should be something like that
    Set(inEnv(Effect(v, Path(accessors)), env))
  case _ => throw MalformedStainlessCode(expr, ...)
}

// In the Let case of expressionEffects.rec
val newEnv = ... // TBD
rec(b, newEnv).flatMap { case ef @ Effect(receiver, path) =>
  if (receiver == vd.toVariable) applyEffect(env, path.toSeq, e)
  else Set(ef)
}
```

> In that example it doesn't work, since rec(b, env) is empty given how the function is written for the other cases (it's non-empty if you enrich the environment as in the original code).

Hmm, how running with a clean environment? Something like `rec(b, variablesOf(b).map(...).toMap)`.

jad-hamza: Thanks a lot for these suggestions, I pushed new commits with changes that look like this. Is that what you had in mind?

I got this error on the `AccessorAliasing.scala` file:

```scala
[ Debug  ] Symbols before AntiAliasing
[ Debug  ] 
[ Debug  ] def test$0(b$18: B$75): Unit = {
[ Debug  ]   val a$23: A$117 = a$22(b$18)
[ Debug  ]   val old$0: BigInt = a$23.x$128
[ Debug  ]   a$23.x$128 = (a$23.x$128 + 1)
[ Debug  ]   val ev$1$0: A$117 = a$22(b$18)
[ Debug  ]   ev$1$0.x$128 = (ev$1$0.x$128 + 1)
[ Debug  ]   assert(a$22(b$18).x$128 == old$0 + 2)
[ Debug  ]   assert(a$23.x$128 == old$0 + 2)
[ Debug  ]   assert(a$23 == a$22(b$18))
[ Debug  ]   ()
[ Debug  ] }
```

```
[ Fatal  ] AccessorAliasing.scala:9:5: Could not apply effects List(ClassFieldAccessor(x$128)) to a$22(b$18)
```

samarion: You can support more cases in `applyEffect`, like
```scala
case (_, fi: FunctionInvocation) if !fi.tfd.isRecursive =>
  applyEffect(env, accessors, fi.inlined)
case (_, IfExpr(_, t, e)) =>
  applyEffect(env, accessors, t) ++ applyEffect(env, accessors, e)
...
```

I wonder if the default case shouldn't rather be:
```scala
case (_, expr) =>
  // I'm still not sure `inEnv` is the right behavior here, but it should be something like that
  getEffects(expr).flatMap(target => inEnv(target.toEffect ++ accessors, env))
```

Actually, looking at the `rec` implementation within `getEffects`, it seems to have pretty much the exact behavior we want here (I think). Maybe you should just expose that function?

jad-hamza: I tried to use that (renamed to `getTargets`) and it seems to work on the small examples. I also changed `getExactEffects` and `getKnownEffects`. But we ignore `env` (and `newEnv`) when calling `getTargets`?

Next I need to understand why this works :)

Thanks

samarion: Yeah, I guess you'll have to map with `inEnv(_, env)` afterwards.

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -301,93 +315,96 @@ trait EffectsAnalyzer extends oo.CachingPhase {
...
 
-    rec(expr, Seq.empty)
+  def getTargets(expr: Expr)(implicit symbols: Symbols): Set[Target] = {
+    getTargets(expr, Seq.empty)
   }
 
-  def getExactEffects(expr: Expr)(implicit symbols: Symbols): Set[Target] = getEffects(expr) match {
-    case effects if effects.nonEmpty => effects
-    case _ => throw MalformedStainlessCode(expr, s"Couldn't compute exact effect targets in: $expr")
+  def getExactTargets(expr: Expr)(implicit symbols: Symbols): Set[Target] = try {
```
samarion, line 397: Let's just drop these functions (`getExactTargets` and `getExactTargetsOpt`), I don't like them. We should be able to just call `getTargets(expr)` instead.

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -418,13 +435,23 @@ trait EffectsAnalyzer extends oo.CachingPhase {
...
       }
 
     def rec(expr: Expr, env: Map[Variable, Effect]): Set[Effect] = expr match {
       case Let(vd, e, b) if symbols.isMutableType(vd.tpe) =>
-        rec(e, env) ++ rec(b, env ++ effect(e, env).map(vd.toVariable -> _))
+        if ((variablesOf(e) & variablesOf(b)).forall(v => !isMutableType(v.tpe))) {
+          val effe = rec(e, env)
+          val newEnv = (variablesOf(b) ++ freeVars).map(v => v -> Effect(v, Path.empty)).toMap
+          val effb = rec(b, newEnv)
+          effb.flatMap { case ef @ Effect(receiver, path) =>
```
samarion, line 448: `effe ++ effb.flatMap { ... }.map(inEnv(_, env))`

jad-hamza: Fixed, thanks

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
jad-hamza: Then this call will fail on some examples:
https://github.com/jad-hamza/stainless/blob/0814ae8c2e0f30f2986b535beb38afb0a36d39fb/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L273
Do you prefer to wrap the call here in a `try`?

jad-hamza:
What should we do with the call here:
https://github.com/jad-hamza/stainless/blob/719e203e636bf88a0e83449429542d17ee90a731/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala#L49-L55

Is it safe to completely remove it? And the `LetVar` one?

This makes the `VarAlias` example fail because of the computation of targets on `val a = A(data)`:
https://github.com/jad-hamza/stainless/blob/better-aliasing-fix2/frontends/benchmarks/imperative/valid/VarAlias.scala

Initially I had changed the way `isExpressionFresh` works to avoid the call above, but actually I was ignoring free variables (returning `true`) in the expression. This was wrongly making the `A(data)` expression fresh. Now that I fixed that, the example fails again.

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
samarion: Which examples will this fail on? I don't understand why that transformation would make sense.

jad-hamza: It's again on the `VarAlias.scala` example, because the call to `getTargets` on the line above fails on `A(data)` which is bound in a `Let`.

```
Couldn't compute effect targets in class constructor: A(data)
```

samarion:
> Is it safe to completely remove the call in EffectsChecker? And the one in LetVar?

No I don't think we can remove the call completely, but we can probably restrict the check to only cases where `(variablesOf(e) & variablesOf(b)).exists(v => isMutableType(v.tpe))`.

samarion:
Also IIRC we discussed with Georg at some point that there shouldn't be any need for both `getTargets` and `isExpressionFresh`. You should be able to implement `isExpressionFresh(e) as `getTargets(e).isEmpty`.

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
samarion: Hmm, I'd really prefer we didn't rely on exception existence to decide on how something should be transformed... We can probably replace the use of `getExactTargetsOpt` by `variablesOf(e).exists(v => isMutableType(v.tpe))`.

jad-hamza:
> Also IIRC we discussed with Georg at some point that there shouldn't be any need for both `getTargets` and `isExpressionFresh`. You should be able to implement `isExpressionFresh(e)` as `getTargets(e).isEmpty`.

We have many calls to `isExpressionFresh`, and unfortunately `getTargets` crashes on a lot of those (for instance `A(data)` again). Could we instead make `getTargets` natively return an `Option` (instead of having the other layer with the exception?)).

samarion:
> We have many calls to isExpressionFresh, and unfortunately getTargets crashes on a lot of those (for instance A(data) again). Could we instead make getTargets natively return an Option (instead of having the other layer with the exception?)).

In general, `getTargets` throwing an exception really corresponds to invalid Stainless code (so throwing is reasonable behavior). You could change `getExpressionFresh` to `checkExpressionFresh` and only call it in positions where throwing is expected. The transformation anyway shouldn't depend on whether an expression is fresh or not.

Alternatively, you could catch the `MalformedStainlessException` in `isExpressionFresh` and return `false`, but I think it would probably be cleaner to not produce a boolean at all since it will only be an under-approximation.

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
jad-hamza: I wrote this in the `Let` case:

```scala
          case l @ Let(vd, e, b) if isMutableType(vd.tpe) =>
            val newExpr = transform(e, env)
            // if the bound expression doesn't contain any mutable subexpression, we can use it
            // as a rewriting
            if (!exprOps.exists { e2 => isMutableType(e2.getType) }(e)) {
              val newBody = transform(b, env withRewritings Map(vd -> newExpr))
              Let(vd, newExpr, newBody).copiedFrom(l)
            } else {
              val newBody = transform(b, env withBinding vd)
              LetVar(vd, newExpr, newBody).copiedFrom(l)
            }
```

On an example, checking just the free variables wasn't enough, because a binding such as `Array(1, 2, 3)` was going in to the `withRewritings` branch. Does this make sense?

I'm now wondering how this example is supposed to work:

```scala
case class A(var x: BigInt)
trait B { var a: A }

object Test {
  def test(b: B): Unit = {
    val old = b.a.x
    b.a.x += 1
    assert(b.a.x == old + 1)
  }
}
```

Before `AntiAliasing`, we have:

```scala
[ Debug  ] Symbols before AntiAliasing
[ Debug  ] 
[ Debug  ] def test$0(b$18: B$75): Unit = {
[ Debug  ]   val old$0: BigInt = a$22(b$18).x$128
[ Debug  ]   val ev$1$0: A$117 = a$22(b$18)
[ Debug  ]   ev$1$0.x$128 = (ev$1$0.x$128 + 1)
[ Debug  ]   assert(a$22(b$18).x$128 == old$0 + 1)
[ Debug  ]   ()
[ Debug  ] }
```

It looks like we lost information, and we don't know that we also need to apply the effect to: `Target(b, None, .a)`
When we arrive at the transformation of the `FieldAssignment`: `ev$1$0.x$128 = (ev$1$0.x$128 + 1)`, we have `ev$1$0` in the bindings, but not in the rewritings.

Should we perhaps keep more information in `env.bindings` and propagate the effects there?

samarion: > On an example, checking just the free variables wasn't enough, because a binding such as Array(1, 2, 3) was going in to the withRewritings branch. Does this make sense?

Isn't the `withRewritings` branch supposed to handle the case where we have aliasing? I would expect the condition to be `variablesOf(e).exists(v => isMutableType(v.tpe))` (without negation), which would let `Array(1, 2, 3)` go into the `withBinding` branch (which handles the case where the let body is a fresh mutable value).

> I'm now wondering how this example is supposed to work [...]

I think we should have `ev -> a(b)` in the rewrittings. In the current implementation, we have `getKnownEffects(a(b)) == Set(Target(b, None, .a))` so we'd go into the `withRewritings` branch.

jad-hamza:
```scala
object Array7 {
  def foo(a: Array[Int]): Unit = {
    require(a.length > 0)
    val a2 = Array.fill(a.length)(0)
    a2(0) = a(0)
    assert(a2(0) == a(0))
  }
}
```

Should this go to the `withRewritings` branch? Currently it does, and then when we arrive at `ArrayUpdate`, we have the rewriting `a2 -> Array(0, ..., 0) (of size a.length)`, and `getTargets` on the right-hand-side is empty. So the assignment gets transformed to `()`. Seeing this example, it now feels strange to change the transformation depending on whether a free variable appears or not.

In the rest of the test suite, I got errors on `getTargets` not supporting enough trees:
* `ArrayLength` in `While4` (so I added it as `Set.empty` in `getTargets`)
* `Block` (so I added it as `getTargets(last)`)

Is it accurate to say that given an expression `e`, `getTargets` returns the set of targets that would be affected if `e` is bound to some `x` and `x` is modified? If so, the additions for `ArrayLength` and `Block` should be ok?

I also got an issue with the `Snapshot` tree, so I changed the condition so that transformation of let bindings with snapshot go to the `withBindings` case.

I changed the `getTargets` of `IfExpr`, there was a conjunct missing, was that on purpose?
https://github.com/epfl-lara/stainless/pull/920/files#diff-8e88ff59f4e5533bd8fa25ed53c11010adbe901f39faa5647962bb39fdf45548R364

samarion: > Should the val a2 binding go to the withRewritings branch?

Hmm, no, probably not.

> Seeing this example, it now feels strange to change the transformation depending on whether a free variable appears or not.

Yes, you're right. Maybe using the exception is actually better. But I'm a bit worried we won't correctly reject all bad cases if we do it that way...

Maybe we need the condition to be something like:
```scala
if (Try(getTargets(e)).toOption == Some(Set.empty)) {
  // If we are sure `e` has no aliasing, we can use `withBinding` and allow mutation of `vd`
  val newBody = transform(b, env withBinding vd)
  LetVar(vd, newExpr, newBody).copiedFrom(l)
} else {
  // Otherwise, we stay conservative and use rewritings,
  // which means we'll check target resolution at mutation points
  val newBody = transform(b, env withRewritings Map(vd -> newExpr))
  Let(vd, newExpr, newBody).copiedFrom(l)
}
```

> Is it accurate to say that given an expression e, getTargets returns the set of targets that would be affected if e is bound to some x and x is modified? If so, the additions for ArrayLength and Block should be ok?

Yes, that sounds correct.

> I changed the getTargets of IfExpr, there was a conjunct missing, was that on purpose?

Looks to me like that was an error.

jad-hamza: Thanks! Now the ImperativeSuite seems to be working, let's see what happens with larabot.

* You say the `withRewritings` branch is more conservative, but then the example above (`Array7.foo`) should succeed also when going to that branch?
* I removed these two lines, does this make sense? I don't understand why we failed when the targets were empty? (I think the same logic was used in `getExactEffects/getKnownEffects)` https://github.com/epfl-lara/stainless/pull/920/files#diff-8e88ff59f4e5533bd8fa25ed53c11010adbe901f39faa5647962bb39fdf45548L370-L371

samarion: > You say the withRewritings branch is more conservative, but then the example above (Array7.foo) should succeed also when going to that branch?

You're right, it's not really more conservative...

IIUC there are 3 cases which can be supported for let-bindings:
1. the rhs expression is fresh: in that case we can use `withBinding`.
2. the rhs expression is an alias: in that case we can use `withRewritings`.
3. the rhs expression contains an aliased variable, but we can resolve it in `getTargets` given the current path (_e.g._ for the `A(b)` case). This case can also be handled with `withRewritings`.

The issue is that if there's a mix between fresh and aliased expressions, the transformation is broken. For example, if we have `val a = if (cond) A(b) else A(1)`, then neither bindings nor rewritings can handle the expression. This might somewhat explain the empty targets check (but it's not sufficient, anyway). This mix can also occur in case 3), for example if we had `val a = A(b, Array(1, 2, 3))` and then tried to mutate `a.array(0) = 2`. Using rewritings would fail in this case.

I wonder if rewritings are maybe just fundamentally broken for aliasing. We probably have all the information we need in `getTargets` to correctly resolve variables. Given a let-binding `val a = b in e` with `getTargets(b) = target`, we can probably transform references to `a` into something like `if (target.cond) target.path.on(target.receiver) else a`.

Could you also add the following benchmark to the test set? I'm not sure we correctly support this. (It's fine if it gets rejected, as long as it doesn't return a counterexample.)
```scala
case class B(var i: BigInt)
case class A(var b: B)

def test(): Unit = {
  val b = B(0)
  val a = A(b)
  a.b = B(1)
  assert(a.b.i == 1)
}
```

jad-hamza: > the rhs expression is an alias: in that case we can use withRewritings.

An alias means just `val x = y` for another variable `y`?

By the way why don't we have the same cases in `LetVar`?

> Could you also add the following benchmark to the test set? I'm not sure we correctly support this. (It's fine if it gets rejected, as long as it doesn't return a counterexample.)

Sure! The example fails on the following path:
* first `Let` goes to `withBindings` branch
* second `Let` goes to `withRewritings`
* then we get to `transform` the `FieldAssignment` `a.b`, we use the rewriting on `a` to get `A(b)`, and `getTargets` fails on `A(b)`
https://github.com/jad-hamza/stainless/blob/232139920e793e62b361d140aa2f88c9ba5bbe7a/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L347

jad-hamza: The test suite currently runs fast but that's because I added too much filtering and some things like `LibrarySuite` aren't verified.

jad-hamza: I changed the place where `UserFiltering` occurs (now `LibrarySuite` doesn't use it), and the test suite still runs faster (but slower than when the filtering was everywhere), I hope nothing else is disabled now.

samarion: > An alias means just val x = y for another variable y?

Yes, or something like `val x = y.z` where we alias a field/array index for which the target can be computed precisely.

> By the way why don't we have the same cases in LetVar?

For `LetVar`, we should disallow aliasing because it can't be tracked across assignments. So only the `withBinding` case is legal (which is how the case is implemented in master).

> Sure! The example fails on the following path:

Hmm, I'm not sure that's the way we want it to fail... let's just keep the example in the extraction/invalid suite to make sure it doesn't start unexpectedly working after we make more changes :)

---

In order to support mixing fresh and aliased rhs, I wonder if the rewriting should look something like:
```scala
val rewriting = getTargets(newExpr).foldRight(vd.toVariable : Expr) {
  case (Target(r, cond, path), rewriting) => IfExpr(cond, applyEffect(Target(r, None, path), r), rewriting)
}

// And then use the following env in the recursive transform call
env withRewritings Map(vd -> rewriting)
```

This won't allow your original use case of `val a = A(aliased)` though...

I'm now thinking we should maybe handle that case by copying over the effect _after_ the let body has been evaluated. This would be closer to the effect analysis where the concrete effects are computed a posteriori. Something like
```scala
val newBody = transform(b, env withBinding vd)
val copyEffects = effects(b).filter(_.receiver == vd.toVariable).map { e =>
  applyEffect(e.on(newExpr), vd)
}

Let(
  resVd,
  LetVar(vd, newExpr, newBody).copiedFrom(l),
  Block(copyEffects, resVd.toVariable).copiedFrom(l)
).copiedFrom(l)
```

Does that make sense?

jad-hamza: Don't we want the effects to be applied exactly at the place they occur (otherwise some assertions inside the `Let` body `b` could fail because they don't see the effects yet?)

samarion: Well, we're relying on the fact that we don't used the aliased variable inside the let body, so applying the effect later (after the body evaluation completes) should be ok.

jad-hamza: In that example we refer to both `a` and `b.a`, wouldn't that be a problem here?
https://github.com/jad-hamza/stainless/blob/91297951c4107af206cc78fde283b1dd278efe6a/frontends/benchmarks/imperative/valid/AccessorAliasing.scala

samarion: Ah, but that's a case where the targets can be computed precisely. So we can (and should) directly apply the effects. I was talking about the case where we _can't_ compute the targets (_e.g._ `val a = A(b) in body`) _but_ the aliased variable (`b`) is never used in `body`. In that case, we can apply the effects to `b` after executing `body`.

jad-hamza: Oh I see, so should we add this in a new branch of the top-level `if (Try(getTargets(e)).toOption == Some(Set.empty))`?

samarion: Yes, I think so. The branch condition should look something like what we have in `EffectsAnalyzer` for let-bindings.

Also, I think the top-level if-expression should now be something like:
```scala
if (Try(getTargets(e)).isSuccess) {
  // This branch handles all cases when targets can be precisely computed, namely when
  // 1. newExpr is a fresh expression
  // 2. newExpr is a precise alias to an existing variable
  val rewriting = getTargets(newExpr).foldRight(vd.toVariable : Expr) {
    case (Target(r, cond, path), rewriting) =>
      IfExpr(cond, applyEffect(Target(r, None, path), r), rewriting)
  }

  val newBody = transform(body, env withRewritings Map(vd -> rewriting) withBinding vd)
  LetVar(vd, newExpr, newBody).copiedFrom(l)
} else {
  // This branch handles the cases where we can apply effects afterwards.
  // TODO: check that variablesOf(e) & variablesOf(body) all have immutable types,
  // similarly to the check in EffectsAnalyzer (just to be safe).
  val newBody = transform(body, env withBinding vd)
  
  val copyEffects = effects(body).filter(_.receiver == vd.toVariable).map { e =>
    applyEffect(e.on(newExpr), vd)
  }

  val resVd = ValDef.fresh("res", body.getType).copiedFrom(body)
  Let(
    resVd,
    Let(vd, newExpr, newBody).copiedFrom(l),
    Block(copyEffects, resVd.toVariable).copiedFrom(l)
  ).copiedFrom(l)
}
```

(My use of `Effect` and `Target` APIs in the above snippet may be quite wrong, you'll need to double check.)

jad-hamza: I modified a bit the code for it to compile (pushed), and it's failing on two examples: `VarAlias`, and `AccessorAliasing`.

For `VarAlias`, `copyEffects` generates an expression which is has undefined variable (see `a$23` below the definition `val res$73 = ...`). I changed the name `applyEffect` to `updatedTarget` because the function uses `Target` and not `Effect` and also it looks like that it doesn't create an assignment, which we need to create here?
If I understand correctly, instead of `a$23` here, we want an `ArrayUpdate` on `data$3`?

```scala
[ Debug  ] Symbols before AntiAliasing
[ Debug  ] 
[ Debug  ] def f$39(a$22: A$117): Unit = {
[ Debug  ]   require(a$22.data$0.length > 0)
[ Debug  ]   a$22.data$0(0) = 10
[ Debug  ] } ensuring {
[ Debug  ]   (x$1$14: Unit) => a$22.data$0.length > 0
[ Debug  ] }
[ Debug  ] 
[ Debug  ] def g$1: Unit = {
[ Debug  ]   val data$3: Array[Int] = Array(100)
[ Debug  ]   val a$23: A$117 = A$117(data$3)
[ Debug  ]   f$39(a$23)
[ Debug  ]   assert(a$23.data$0(0) == 10)
[ Debug  ]   ()
[ Debug  ] }
[ Debug  ] 
[ Debug  ] 
[ Debug  ] 
[ Debug  ] Symbols after AntiAliasing
[ Debug  ] 
[ Debug  ] def f$39(a$22: A$117): (Unit, A$117) = {
[ Debug  ]   require(a$22.data$0.length > 0)
[ Debug  ]   var a$24: A$117 = a$22
[ Debug  ]   ({
[ Debug  ]     a$24 = A$117(@unchecked a$24.asInstanceOf[A$117].data$0.updated(0, 10))
[ Debug  ]     ()
[ Debug  ]   }, a$24)
[ Debug  ] } ensuring {
[ Debug  ]   (x$1$15: (Unit, A$117)) => x$1$15._2.data$0.length > 0
[ Debug  ] }
[ Debug  ] 
[ Debug  ] def g$1: Unit = {
[ Debug  ]   var data$3: Array[Int] = Array(100)
[ Debug  ]   val res$73: Unit = {
[ Debug  ]     val a$23: A$117 = A$117(data$3)
[ Debug  ]     var res$72: (Unit, A$117) = f$39(a$23)
[ Debug  ]     (if (res$72._2.isInstanceOf[A$117]) {
[ Debug  ]       a$23 = @unchecked A$117(@unchecked res$72._2.asInstanceOf[A$117].data$0).asInstanceOf[A$117]
[ Debug  ]     } else {
[ Debug  ]       ()
[ Debug  ]     })
[ Debug  ]     res$72._1
[ Debug  ]     assert(a$23.data$0(0) == 10)
[ Debug  ]     ()
[ Debug  ]   }
[ Debug  ]   a$23
[ Debug  ]   res$73
[ Debug  ] }
```

For `AccessorAliasing.scala` (simplified version):

```scala
[ Debug  ] Symbols before AntiAliasing
[ Debug  ] 
[ Debug  ] def test(b: B): Unit = {
[ Debug  ]   val old: BigInt = a(b).x
[ Debug  ]   val ev$1: A = a(b)
[ Debug  ]   ev$1.x = (ev$1.x + 1)
[ Debug  ]   assert(a(b).x == old + 1)
[ Debug  ]   ()
[ Debug  ] }
```

There is a crash in the `getTargets` invoked in `FieldAssignment` case for `ev$1.x = (ev$1.x + 1)`.
The rewriting here is `ev$1 -> BExt(@unchecked b.asInstanceOf[BExt].__x, b)`, and `getTargets` fails on the right-hand-side.

While we're in the `FieldAssignment` case, don't we want to distinguish whether we have to do the targets modifications here, or whether we let them be done in the `Let` case? It seems hard to distinguish here in which case we are because the receiver of the assignment can be a complex expression.

Also didn't you want to avoid `withRewritings` in the new solution?

samarion: > For VarAlias, copyEffects generates an expression which is has undefined variable (see a$23 below the definition val res$73 = ...)

I don't see which variable is undefined. It seems like `a$23` should be a `LetVar` though, so the `Let(vd, newExpr, newBody)` in the second case should be `LetVar(vd, newExpr, newBody)` instead.

> [...] it looks like that it doesn't create an assignment, which we need to create here? If I understand correctly, instead of a$23 here, we want an ArrayUpdate on data$3?

No I think this shape is correct. After `AntiAliasing` we shouldn't have any `ArrayUpdate` or `FieldAssignment` statements anymore, they are all converted into variable assignments where unchanged fields are copied over.

> The rewriting here is ev$1 -> BExt(@unchecked b.asInstanceOf[BExt].__x, b), and getTargets fails on the right-hand-side.

I don't understand why we're getting this rewriting. Maybe I need the rest of the symbols as context?

> While we're in the FieldAssignment case, don't we want to distinguish whether we have to do the targets modifications here, or whether we let them be done in the Let case? It seems hard to distinguish here in which case we are because the receiver of the assignment can be a complex expression.

I don't think we need to care? Every `FieldAssignment` should be transformed into a let-var assignment. The complex expression we see (because of rewritings) should have the shape of a big `if-then-else` with precise targets in the branches, and `getTargets` should be able to handle it when required (hopefully). Maybe I'm being optimistic here and we'll really have to do away with rewritings and track targets instead.

jad-hamza: > I don't see which variable is undefined.

It's the occurrence of `a$23` in the line before last.

> It seems like `a$23` should be a `LetVar` though, so the `Let(vd, newExpr, newBody)` in the second case should be `LetVar(vd, newExpr, newBody)` instead.

Thanks I'll fix that

> No I think this shape is correct. After `AntiAliasing` we shouldn't have any `ArrayUpdate` or `FieldAssignment` statements anymore, they are all converted into variable assignments where unchanged fields are copied over.

Shouldn't `data$3` be reassigned then?

samarion: > It's the occurrence of a$23 in the line before last.

Ah I see. I was too focused on the block where it was declared...

> Shouldn't data$3 be reassigned then?

Yes, that's true.

I think the `copyEffect` block I sketched out isn't correct. It should at the very least have an `Assignment` wrapping the `updatedTarget`, and `on` probably isn't the right method to call. Maybe it should rather be something like
```scala
val copyEffects = effects(b).filter(_.receiver == vd.toVariable).flatMap { e =>
  getTargets(newExpr, e.path.accessors).map { case Target(receiver, _, path) =>
    Assignment(receiver, updatedTarget(Target(receiver, None, path), vd.toVariable))
  }
}
```

Anyway, I can't run the code so you'll probably have to fiddle around the implementation a bit.

By the way, `updatedTarget` is broken for targets with conditions (already in master). I'm not sure it makes sense to actually support those, could you please try to adapt the API to only support `Effect`?

jad-hamza: For the `data$3` example, what should be the assignment in the result of the transformation?

Is it `data$3 = data$3.updated(0, 10)`? I'm sure how to get that information in the code (effects contain accessors, but I don't think they contain this information; in this example `effects(b).filter(_.receiver == vd.toVariable)` contains just `Effect(a$23.data$0)`)

Otherwise, I was thinking `data$3 = a$23.data$0`, but that's also not possible since `a$23` is not in scope here.

samarion: I think it should be `data$3 = a$23.data$0`.

Maybe you need to swap the `Let`s around, something like
```scala
LetVar(
  vd, newExpr,
  Let(resVd, newBody, Block(copyEffects.toSeq, resVd.toVariable)))
```

jad-hamza: Oh right, thanks that makes sense, and that works! You're inventing a new programming paradigm (by snippets :))

About the `updatedTarget` or `updatedEffect` (I think I need a better name!), I changed it to use `Effect` but now an example that used to work fails. See below, it's a simplified version of `AccessorMutation2.scala`

With Stainless master:
```scala
[ Debug  ] Symbols before AntiAliasing
[ Debug  ] 
[ Debug  ] def t2(tata: Tata): Unit = box(tata).array(0).value = 2
[ Debug  ] 
[ Debug  ] 
[ Debug  ] Symbols after AntiAliasing
[ Debug  ] 
[ Debug  ] def t2(tata: Tata): (Unit, Tata) = {
[ Debug  ]   var tata: Tata = tata
[ Debug  ]   ({
[ Debug  ]     tata = @unchecked (if (tata.isInstanceOf[Bar]) {
[ Debug  ]       Bar(ArrayBox(@unchecked @unchecked tata.asInstanceOf[Bar].box.asInstanceOf[ArrayBox].array.updated(0, IntBox(2))))
[ Debug  ]     } else {
[ Debug  ]       tata
[ Debug  ]     }).asInstanceOf[Tata]
[ Debug  ]     tata = @unchecked (if (¬tata.isInstanceOf[Bar] && tata.isInstanceOf[Foo]) {
[ Debug  ]       Foo(ArrayBox(@unchecked @unchecked tata.asInstanceOf[Foo].box.asInstanceOf[ArrayBox].array.updated(0, IntBox(2))))
[ Debug  ]     } else {
[ Debug  ]       tata
[ Debug  ]     }).asInstanceOf[Tata]
[ Debug  ]     tata = @unchecked (if (¬tata.isInstanceOf[Bar] && ¬tata.isInstanceOf[Foo]) {
[ Debug  ]       Baz(ArrayBox(@unchecked @unchecked tata.asInstanceOf[Baz].box.asInstanceOf[ArrayBox].array.updated(0, IntBox(2))))
[ Debug  ]     } else {
[ Debug  ]       tata
[ Debug  ]     }).asInstanceOf[Tata]
[ Debug  ]     ()
[ Debug  ]   }, tata)
[ Debug  ] }
```

With this branch:

```scala
[ Debug  ] Symbols after AntiAliasing
[ Debug  ] 
[ Debug  ] def t2(tata: Tata): (Unit, Tata) = {
[ Debug  ]   var tata: Tata = tata
[ Debug  ]   ({
[ Debug  ]     tata = Bar(ArrayBox(@unchecked @unchecked tata.asInstanceOf[Bar].box.asInstanceOf[ArrayBox].array.updated(0, IntBox(2))))
[ Debug  ]     tata = Foo(ArrayBox(@unchecked @unchecked tata.asInstanceOf[Foo].box.asInstanceOf[ArrayBox].array.updated(0, IntBox(2))))
[ Debug  ]     tata = Baz(ArrayBox(@unchecked @unchecked tata.asInstanceOf[Baz].box.asInstanceOf[ArrayBox].array.updated(0, IntBox(2))))
[ Debug  ]     ()
[ Debug  ]   }, tata)
[ Debug  ] }
```

samarion: > About the updatedTarget or updatedEffect (I think I need a better name!), I changed it to use Effect but now an example that used to work fails. See below, it's a simplified version of AccessorMutation2.scala

Oh my bad, I mis-understood part of the implementation... It should be fine to keep as is/was.

As a follow-up, we could look into completely doing away with target conditions and instead computing them inside `updatedTarget`. For example, the `ADTFieldAccessor` case would become something like
```scala
case ADTFieldAccessor(id) :: fs =>
  val adt @ ADTType(_, tps) = receiver.getType
  val tcons = adt.getSort.constructors.find(_.fields.exists(_.id == id)).get
  val r = rec(Annotated(ADTSelector(receiver, id).copiedFrom(newValue), Seq(Unchecked)).copiedFrom(newValue), fs)
  val newExpr = ADT(tcons.id, tps, tcons.definition.fields.map { vd =>
    if (vd.id == id) r
    else Annotated(ADTSelector(receiver, vd.id).copiedFrom(receiver), Seq(Unchecked)).copiedFrom(receiver)
  }).copiedFrom(newValue)

  if (adt.getSort.constructors.size > 1) {
    IfExpr(IsConstructor(receiver, tcons.id), newExpr, receiver)
  } else {
    newExpr
  }
```

jad-hamza: So I keep my change for `updatedTarget` to only use `Effect`, and add your `ADTFieldAccessor` change to the function to add the conditionals?

About the `Let` transformation, I added a `wrap` method on `Effect` (extracted from the `Path.on` function), does it make sense? I wasn't able to get `a$23.data$0` for the assignment otherwise, and locally the test suite was happy (except on the `Tata` that shoud be fixed with the `updatedTarget` change)
https://github.com/epfl-lara/stainless/pull/920/files#diff-0165b6310b74b20ca5f074f69069a8f81a1280c6cbb95a486ca5ab004b7dae15R325

jad-hamza: (I reverted to use `updatedTarget` for now)

jad-hamza:
Merging for now, thanks a lot for the help @samarion :+1: 
Please notify me if you want me to modify some things and I'll open another PR.

PR #923, opened 2021-02-27
Transformation for return keyword

jad-hamza:
This implements @romac transformation idea (with suggestions from @samarion) for the `return` keyword.

Feedback is welcome! (Loops are not supported yet, I can add them to this PR or a different one)

Here are the examples we discussed (added to imperative benchmarks in this PR):

<details>
  <summary> SimpleReturn.scala </summary>

```scala
[ Debug  ] Symbols before ReturnElimination
[ Debug  ] 
[ Debug  ] -------------Functions-------------
[ Debug  ] def example1: Int = {
[ Debug  ]   return 0
[ Debug  ]   1
[ Debug  ] }
[ Debug  ] 
[ Debug  ] def example2((x: Int) @pure): Int = {
[ Debug  ]   val a: Int = if (x > 0) {
[ Debug  ]     return 1
[ Debug  ]   } else {
[ Debug  ]     2
[ Debug  ]   }
[ Debug  ]   3
[ Debug  ] }
[ Debug  ] 
[ Debug  ] def example3((cond: Boolean) @pure): Boolean = {
[ Debug  ]   val a: Int = if (cond) {
[ Debug  ]     return true
[ Debug  ]   } else {
[ Debug  ]     1
[ Debug  ]   }
[ Debug  ]   false
[ Debug  ] }
[ Debug  ] 
[ Debug  ] def example4: Boolean = {
[ Debug  ]   val x: Int = return false
[ Debug  ]   true
[ Debug  ] }
[ Debug  ] 
[ Debug  ] def example5((cond: Boolean) @pure): Int = {
[ Debug  ]   val x: Int = if (cond) {
[ Debug  ]     return 100
[ Debug  ]   } else {
[ Debug  ]     5
[ Debug  ]   }
[ Debug  ]   x
[ Debug  ] }
[ Debug  ] 
[ Debug  ] 
[ Debug  ] Symbols after ReturnElimination
[ Debug  ] 
[ Debug  ] -------------Functions-------------
[ Debug  ] def example1: Int = {
[ Debug  ]   val topLevelCF: ControlFlow[Int, Int] = {
[ Debug  ]     val cf: ControlFlow[Int, Nothing] = Return[Int, Nothing](0)
[ Debug  ]     cf match {
[ Debug  ]       case Return(retValue) =>
[ Debug  ]         Return[Int, Int](retValue)
[ Debug  ]       case Proceed(proceedValue) =>
[ Debug  ]         Proceed[Int, Int](1)
[ Debug  ]     }
[...]

PR #925, opened 2021-02-27
Add support for return in while loops

jad-hamza:
This implements @romac transformation idea (with suggestions from @samarion and @vkuncak) for the return keyword.

Extends the PR https://github.com/epfl-lara/stainless/pull/923 (which doesn't have loops)

Some examples:
https://github.com/jad-hamza/stainless/blob/93159702480b8303039d2f416d1c74f0f445edb2/frontends/benchmarks/imperative/valid/ControlFlow2.scala
https://github.com/jad-hamza/stainless/blob/93159702480b8303039d2f416d1c74f0f445edb2/frontends/benchmarks/imperative/valid/ControlFlow3.scala
https://github.com/jad-hamza/stainless/blob/93159702480b8303039d2f416d1c74f0f445edb2/frontends/benchmarks/imperative/valid/FindIndex.scala

I'm not sure how to prove `containsFromLemma` in that last example.


vkuncak:
This variant 
```scala
import stainless.annotation._
import stainless.collection._
import stainless.lang._
import stainless.io._

object FindIndex {
  def findIndex[T](a: Array[T], t: T): Option[Int] = {
    val N: Int = a.length
    var i: Int = 0
    (while (i < N) {
      decreases(N - i)
      if (a(i) == t) return Some[Int](i)
      i = i + 1
    }).invariant(0 <= i && i <= N)
    None[Int]()
  }
}
```
fails with 
```
findIndex       refinements checks for subtyping            invalid    frontends/benchmarks/imperative/valid/FindIndex.scala:40:5 
[Warning ]  => INVALID
[Warning ] Found counter-example:
[Warning ]   a: Array[T]    -> Array[T]()
[Warning ]   v: Option[Int] -> Some[Int](0)
[Warning ]   t: T           -> T#1
```
I don't understand why. Moreover, if I run it with `--debug=trees` I get
```
[ Error  ]       retValue
[ Error  ]     case Proceed(proceedValue) =>
[ Error  ]       proceedValue
[ Error  ]   } is of type Option[Int]
[ Error  ]     topLevelCF is of type ControlFlow[Option[Int], Option[Int]]
[ Error  ]     retValue is of type Option[Int]
[ Error  ]     proceedValue is of type Option[Int]
               val N: Int = a.length
               ^^^^^^^^^^^^^^^^^^^^^
[ Fatal  ] Well-formedness check failed after phase ReturnElimination
```

jad-hamza:
Thanks for the report! I simplified the example to:

```scala
import stainless.lang._

object FindIndex {
  def findIndex(): Option[Int] = {
    while (true) {
      return Some[Int](0)
    }
    None[Int]()
  }
}
```

```scala
[ Debug  ] Symbols before ReturnElimination
[ Debug  ] 
[ Debug  ] def findIndex$0: Option$0[Int] = {
[ Debug  ]   while (true) {
[ Debug  ]     return Some$0[Int](0)
[ Debug  ]     ()
[ Debug  ]   }
[ Debug  ]   None$0[Int]()
[ Debug  ] }
[ Debug  ] 
[ Debug  ] 
[ Debug  ] 
[ Debug  ] 
[ Debug  ] Symbols after ReturnElimination
[ Debug  ] 
[ Debug  ] def findIndex$0: Option$0[Int] = {
[ Debug  ]   val topLevelCF$0: ControlFlow$0[Option$0[Int], Option$0[Int]] = {
[ Debug  ]     val cf$0: ControlFlow$0[Option$0[Int], Unit] = {
[ Debug  ]       def findIndexWhile$0: ControlFlow$0[Option$0[Int], Unit] = {
[ Debug  ]         {
[ Debug  ]           val cf$1: ControlFlow$0[Option$0[Int], Nothing] = Return$0[Option$0[Int], Nothing](Some$0[Int](0))
[ Debug  ]           cf$1 match {
[ Debug  ]             case Return$0(retValue$0) =>
[ Debug  ]               Return$0[Option$0[Int], Unit](retValue$0)
[ Debug  ]             case Proceed$0(proceedValue$0) =>
[ Debug  ]               Proceed$0[Option$0[Int], Unit](())
[ Debug  ]           }
[ Debug  ]         } match {
[ Debug  ]           case Return$0(retValue$1) =>
[ Debug  ]             Return$0[Option$0[Int], Unit](retValue$1)
[ Debug  ]           case Proceed$0(proceedValue$1) =>
[ Debug  ]             if (true) {
[ Debug  ]               findIndexWhile$0
[ Debug  ]             } else {
[ Debug  ]               Proceed$0[Option$0[Int], Unit](())
[ Debug  ]             }
[ Debug  ]         }
[ Debug  ]       } ensuring {
[ Debug  ]         (cfWhile$0: ControlFlow$0[Option$0[Int], Unit]) => cfWhile$0 match {
[ Debug  ]           case Return$0(retValue$2) =>
[ Debug  ]             true
[ Debug  ]           case Proceed$0(proceedValue$2) =>
[ Debug  ]             ¬true
[ Debug  ]         }
[ Debug  ]       }
[ Debug  ]       if (true) {
[ Debug  ]         findIndexWhile$0
[ Debug  ]       } else {
[ Debug  ]         Proceed$0[Option$0[Int], Unit](())
[ Debug  ]       }
[ Debug  ]     }
[ Debug  ]     cf$0 match {
[ Debug  ]       case Return$0(retValue$3) =>
[ Debug  ]         Return$0[Option$0[Int], None$0[Int]](retValue$3)
[ Debug  ]       case Proceed$0(proceedValue$3) =>
[ Debug  ]         Proceed$0[Option$0[Int], None$0[Int]](None$0[Int]())
[ Debug  ]     }
[ Debug  ]   }
[ Debug  ]   topLevelCF$0 match {
[ Debug  ]     case Return$0(retValue$4) =>
[ Debug  ]       retValue$4
[ Debug  ]     case Proceed$0(proceedValue$4) =>
[ Debug  ]       proceedValue$4
[ Debug  ]   }
[ Debug  ] }
```

One issue is the match at the end
```scala
[ Debug  ]     cf$0 match {
[ Debug  ]       case Return$0(retValue$3) =>
[ Debug  ]         Return$0[Option$0[Int], None$0[Int]](retValue$3)
[ Debug  ]       case Proceed$0(proceedValue$3) =>
[ Debug  ]         Proceed$0[Option$0[Int], None$0[Int]](None$0[Int]())
[ Debug  ]     }
```

where we should have `Return$0[Option$0[Int], Option$0[Int]](retValue$3)` instead of `Return$0[Option$0[Int], None$0[Int]](retValue$3)` (`retValue$3` is an `Option[Int]`)

vkuncak:
This works:
```scala
sealed abstract class Opt[T]
case class Non[T]() extends Opt[T]
case class Som[T](x: T) extends Opt[T]

object FindIndex {
  def findIndex(): Opt[Int] = {
    while (true) {
      return Som[Int](0)
    }
    Non[Int]().asInstanceOf[Opt[Int]]
  }
}
```
This suggests that there may be an issue with built-in Option you are injecting during transformation and how its encoded into refinements.

An orthogonal issue is that we cannot end a method with `return` if it expects a value without also providing a dummy value, (presumably because we do not manage to show we can cast the Nothing into the desired type).

vkuncak:
This does not work. Could we use `@inlineOnce` for those methods generated by loops, would it help?
```scala
import stainless.lang._

sealed abstract class Opt[T]
case class Non[T]() extends Opt[T]
case class Som[T](x: T) extends Opt[T]

object FindIndex {

  def findIndex[T](a: Array[T], t: T): Opt[Int] = {
    val N: Int = a.length
    var i: Int = 0
    if (0 < N) {
      (while (i < N) {
        decreases(N - i)
        if (a(i) == t) { return Som[Int](i).asInstanceOf[Opt[Int]] }
        i = i + 1
      }).invariant(0 <= i && i <= N)
      Non[Int].asInstanceOf[Opt[Int]]
    } else Non[Int].asInstanceOf[Opt[Int]]
  } ensuring(res =>
    (res match {
      case Non() => true
      case Som(i) => 0 <= i && i < a.length && a(i)==t
    }))
}
```

jad-hamza:
> This works:
> 
> ```scala
> sealed abstract class Opt[T]
> case class Non[T]() extends Opt[T]
> case class Som[T](x: T) extends Opt[T]
> 
> object FindIndex {
>   def findIndex(): Opt[Int] = {
>     while (true) {
>       return Som[Int](0)
>     }
>     Non[Int]().asInstanceOf[Opt[Int]]
>   }
> }
> ```
> 
> This suggests that there may be an issue with built-in Option you are injecting during transformation and how its encoded into refinements.

The injected option was for the previous method, this example works as well with the built-in option (when using asInstanceOf)

> An orthogonal issue is that we cannot end a method with `return` if it expects a value without also providing a dummy value, (presumably because we do not manage to show we can cast the Nothing into the desired type).

Do you mean something like this?

```scala
object Return {
  def f: Int = {
    return 0
  }
}
```

Perhaps this fails like the example above due to (lack of) variance in the  `ControlFlowSort`, I'll have a look.

samarion:
Here's the issue: https://github.com/epfl-lara/stainless/pull/925/files#diff-27e7c2f127e91f5de44a2e796eb0d6eb7a9414d756a828d43de84d6b4d45a48fR238

You need to pass the expected type to `ReturnTransformer`

jad-hamza:
> Here's the issue: https://github.com/epfl-lara/stainless/pull/925/files#diff-27e7c2f127e91f5de44a2e796eb0d6eb7a9414d756a828d43de84d6b4d45a48fR238
> 
> You need to pass the expected type to `ReturnTransformer`

Fixed, thanks!

jad-hamza:
@vkuncak Thanks to @samarion's fix, your examples don't crash anymore, but your last example still times out (I tried locally with `@inlineOnce` but it didn't help, even with a timeout of 60 seconds).

Here's the program again, this time with the built-in `Option`:

```scala
import stainless.lang._

object FindIndex {
  def findIndex[T](a: Array[T], t: T): Option[Int] = {
    val N: Int = a.length
    var i: Int = 0
    if (0 < N) {
      (while (i < N) {
        decreases(N - i)
        if (a(i) == t) { return Some[Int](i) }
        i = i + 1
      }).invariant(0 <= i && i <= N)
      None[Int]
    } else None[Int]
  } ensuring((res: Option[Int]) =>
    (res match {
      case None() => true
      case Some(i) => 0 <= i && i < a.length && a(i) == t
    }))
}
```

<details>
  <summary> And here are the trees after before and after `ReturnElimination` </summary>

  ```scala
[ Debug  ] Symbols before ReturnElimination
[ Debug  ] 
[ Debug  ] def findIndex[T](a: Array[T], (t: T) @pure): Option[Int] = {
[ Debug  ]   val N: Int = a.length
[ Debug  ]   var (i: Int) @var = 0
[ Debug  ]   (if (0 < N) {
[ Debug  ]     @invariant((0 <= i && i <= N))
[ Debug  ]     while ((i < N)) {
[ Debug  ]       {
[ Debug  ]         decreases(N - i)
[ Debug  ]         (if (a(i) == t) {
[ Debug  ]           return Some[Int](i)
[ Debug  ]         } else {
[ Debug  ]           ()
[ Debug  ]         })
[ Debug  ]         i = (i + 1)
[ Debug  ]         ()
[ Debug  ]       }
[ Debug  ]     }
[ Debug  ]     None[Int]()
[ Debug  ]   } else {
[ Debug  ]     None[Int]()
[ Debug  ]   })
[ Debug  ] } ensuring {
[ Debug  ]   (res: Option[Int]) => res match {
[ Debug  ]     case None() =>
[ Debug  ]       true
[ Debug  ]     case Some(i) =>
[ Debug  ]       (0 <= i && i < a.length) && a(i) == t
[ Debug  ]   }
[ Debug  ] }
[ Debug  ] 
[ Debug  ] 
[ Debug  ] 
[ Debug  ] 
[ Debug  ] Symbols after ReturnElimination
[ Debug  ] 
[ Debug  ] def findIndex[T](a: Array[T], (t: T) @pure): Option[Int] = {
[ Debug  ]   val topLevelCF: ControlFlow[Option[Int], Option[Int]] = {
[ Debug  ]     val N: Int = a.length
[ Debug  ]     var (i: Int) @var = 0
[ Debug  ]     (if (0 < N) {
[ Debug  ]       val cf: ControlFlow[Option[Int], Unit] = {
[ Debug  ]         @inlineOnce
[ Debug  ]         def findIndexWhile: ControlFlow[Option[Int], Unit] = {
[ Debug  ]           require(0 <= i && i <= N && i < N)
[ Debug  ]           decreases(N - i)
[ Debug  ]           {
[ Debug  ]             val cf: ControlFlow[Option[Int], Unit] = if (a(i) == t) {
[ Debug  ]               Return[Option[Int], Unit](Some[Int](i))
[ Debug  ]             } else {
[ Debug  ]               Proceed[Option[Int], Unit](())
[ Debug  ]             }
[ Debug  ]             cf match {
[ Debug  ]               case Return(retValue) =>
[ Debug  ]                 Return[Option[Int], Unit](retValue)
[ Debug  ]               case Proceed(proceedValue) =>
[ Debug  ]                 Proceed[Option[Int], Unit]({
[ Debug  ]                   i = (i + 1)
[ Debug  ]                   ()
[ Debug  ]                 })
[ Debug  ]             }
[ Debug  ]           } match {
[ Debug  ]             case Return(retValue) =>
[ Debug  ]               Return[Option[Int], Unit](retValue)
[ Debug  ]             case Proceed(proceedValue) =>
[ Debug  ]               if (i < N) {
[ Debug  ]                 findIndexWhile
[ Debug  ]               } else {
[ Debug  ]                 Proceed[Option[Int], Unit](())
[ Debug  ]               }
[ Debug  ]           }
[ Debug  ]         } ensuring {
[ Debug  ]           (cfWhile: ControlFlow[Option[Int], Unit]) => cfWhile match {
[ Debug  ]             case Return(retValue) =>
[ Debug  ]               0 <= i && i <= N
[ Debug  ]             case Proceed(proceedValue) =>
[ Debug  ]               ¬(i < N) && 0 <= i && i <= N
[ Debug  ]           }
[ Debug  ]         }
[ Debug  ]         if (i < N) {
[ Debug  ]           findIndexWhile
[ Debug  ]         } else {
[ Debug  ]           Proceed[Option[Int], Unit](())
[ Debug  ]         }
[ Debug  ]       }
[ Debug  ]       cf match {
[ Debug  ]         case Return(retValue) =>
[ Debug  ]           Return[Option[Int], Option[Int]](retValue)
[ Debug  ]         case Proceed(proceedValue) =>
[ Debug  ]           Proceed[Option[Int], Option[Int]](None[Int]())
[ Debug  ]       }
[ Debug  ]     } else {
[ Debug  ]       Proceed[Option[Int], Option[Int]](None[Int]())
[ Debug  ]     })
[ Debug  ]   }
[ Debug  ]   topLevelCF match {
[ Debug  ]     case Return(retValue) =>
[ Debug  ]       retValue
[ Debug  ]     case Proceed(proceedValue) =>
[ Debug  ]       proceedValue
[ Debug  ]   }
[ Debug  ] } ensuring {
[ Debug  ]   (res: Option[Int]) => res match {
[ Debug  ]     case None() =>
[ Debug  ]       true
[ Debug  ]     case Some(i) =>
[ Debug  ]       (0 <= i && i < a.length) && a(i) == t
[ Debug  ]   }
[ Debug  ] }
```
</details>

I thought the issue came from the `&& a(i) == t` in the postcondition of `findIndex` but actually the example still times out after without it. I'll have a look a bit later.

**Comment by samarion** [comment]
Isn't the invariant slightly too weak? I think we need `@invariant((0 <= i && i < N))`

**Comment by jad-hamza** [comment]
> Isn't the invariant slightly too weak? I think we need `@invariant((0 <= i && i < N))`

Hmm I don't think so, because `i` becomes equal to `N` in the last iteration, I think the invariant has to account for that.
I tried with just `0 <= i` for the postcondition of `findIndex` and even this fails, I think the issue is that we have no information at all on the returned value when we look at the postcondition of the recursive function `findIndexWhile`.

The postcondition is:
```scala
[ Debug  ]         } ensuring {
[ Debug  ]           (cfWhile: ControlFlow[Option[Int], Unit]) => cfWhile match {
[ Debug  ]             case Return(retValue) =>
[ Debug  ]               0 <= i && i <= N
[ Debug  ]             case Proceed(proceedValue) =>
[ Debug  ]               ¬(i < N) && 0 <= i && i <= N
[ Debug  ]           }
```   

I was thinking of adding a `.returns((indexOpt: Option[Int]) => ...)` clause for the while loop at the source level so that the user can specify what's supposed to be returned by the while loop, and we can add that clause to the postcondition of `findIndexWhile` to the `retValue` in the `Return` case.

samarion:
What happened to my comment about supporting the general case... Is github becoming too web 2.0 and dropping comments? Anyway, here's it is again (from memory).

I think we can support the general case instead of crashing on unexpected expressions. Something like:
```scala
// Somewhere in transformer context:
private val deconstructor = s.getDeconstructor(t)

// Last case in ReturnTransformer.transform(expr, ...):
case _ =>
  val (ids, vs, es, tps, flags, recons) = deconstructor.deconstruct(expr)
  val tvs = vs.map(NoReturnCheck.transform)
  val ttps = tps.map(NoReturnCheck.transform)
  val tflags = flags.map(NoReturnCheck.transform)
  
  def rec(es: Seq[s.Expr], tes: Seq[t.Expr]): t.Expr = es match {
    case Seq() => recons(ids, tvs, tes, ttps, tflags)
    case e +: rest if !tc.hasReturnExpr(e) =>
      // We use a let-binding here to preserve execution order.
      val vd = t.ValDef.fresh("x", NoReturnCheck.transform(e.getType), true).copiedFrom(e)
      t.Let(vd, NoReturnCheck.transform(e), rec(rest, tes :+ vd.toVariable)).copiedFrom(e)
    case e +: rest =>
      val firstType = NoReturnCheck.transform(e.getType)
      ControlFlowSort.andThen(
        retTypeChecked, firstType, firstType,
        transform(e, e.getType),
        (v: t.Variable) => rec(rest, tes :+ v)
        e.getPos
      )
  }

  // I think here you also need to add a `ControlFlowSort.andThen` to convert the result if `expr.getType != currentType`.
  rec(es, Seq.empty)
```

I'm not 100% sure there won't be an issue with losing the expected type in this transform case, but we can always add more dedicated cases if necessary.

samarion:
> Hmm I don't think so, because i becomes equal to N in the last iteration, I think the invariant has to account for that ...

Oh right, I didn't realize the two `i` were disconnected.

> I was thinking of adding a .returns((indexOpt: Option[Int]) => ...) clause for the while loop at the source level so that the user can specify what's supposed to be returned by the while loop, and we can add that clause to the postcondition of findIndexWhile to the retValue in the Return case.

Yeah, I agree that makes sense.

jad-hamza:
Perhaps there is no need for a `.returns` clause for the while loop, we can use postcondition of the top-level function. I added the postcondition of the top-level function to the postcondition of the function corresponding to the while loop (in the case the while loop returns), and now @vkuncak's example works: https://github.com/jad-hamza/stainless/blob/monadic-return-while/frontends/benchmarks/imperative/valid/FindIndexOption.scala

samarion:
Oh yeah that makes sense, great idea!

vkuncak:
It's great that the example works. I will play with this a bit.
By the way, sometimes you get expressions that pattern match on `ControlFlow` but in the end just compute identity. There are two in the `FindIndexOption` example. Maybe they could be optimized away as they add unnecessary case splits?

vkuncak:
Looks great! I also checked on this example. It works, since you add the postcondition only on the return path, as you should!
```scala
import stainless.lang._
object LifeAfterWhile {
  def f(N: BigInt): BigInt = {
    require(0 <= N)
    var i: BigInt = 0
    (while (i < N) {
      decreases(N - i)
      if (i == 42)
        return N - 200
      else
        i = i + 1
    }).invariant(0 <= i && i <= N)
    i = i - 100
    i
  } ensuring (res => res <= N - 100)
}
```

vkuncak:
@jad-hamza , you seem to forbid all local definitions when your phase runs, even if they don't have anything to do with `return`. Is that the plan? I imagine they could stay non-transformed? Try, e.g.
```scala
  def findIndexOpt[T](a: Array[T], t: T): Option[Int] = {
    val N: Int = a.length
    var i: Int = 0
    
    def notYet(upto: Int): Boolean = true

    if (0 < N) {
      (while (i < N) {
        decreases(N - i)
        if (a(i) == t) { return Some[Int](i) }
        i = i + 1  
      }).invariant(0 <= i && i <= N && a.length == N)
      None[Int]
    } else None[Int]
  } ensuring((res: Option[Int]) =>
    (res match {
      case None() => true
      case Some(i) => 0 <= i && i < a.length && a(i) == t
    }))
```

jad-hamza:
> By the way, sometimes you get expressions that pattern match on `ControlFlow` but in the end just compute identity. There are two in the `FindIndexOption` example. Maybe they could be optimized away as they add unnecessary case splits?

I just removed the top level control-flow variable (@samarion suggested this as well), should I simplify more?

> you seem to forbid all local definitions when your phase runs, even if they don't have anything to do with return. Is that the plan? I imagine they could stay non-transformed? Try, e.g.

Good catch thanks, I added a fix, I now check they do not contain `return` and continue.

> Looks great! I also checked on this example. It works, since you add the postcondition only on the return path, as you should!

Thanks for the example, I've added it!

samarion:
You should be able to support `return` inside inner functions as well. You can probably use the `FunAbstraction` construct from innerfuns.

Also, I don't think there's much point splitting out the non-return while transformation, it makes the triggering logic a bit messy. Why not just keep the transformation within `ReturnElimination`?

vkuncak:
> Also, I don't think there's much point splitting out the non-return while transformation, it makes the triggering logic a bit messy. Why not just keep the transformation within `ReturnElimination`?

Seems fine to keep. There is a discontinuity when one uses `return`, but it's like that with lots of other features already (we can call it: on-demand monadic semantics for verification).

@jad-hamza , you should document the translation with the example, as it is done, e.g., for laws.

jad-hamza:
> I think we can support the general case instead of crashing on unexpected expressions. Something like:

Wouldn't that have been dangerous for the missing `LetRec` case? It feels safer to know where we're applying the transformation, maybe we could use guard your general case by the explicit list of trees for which we want to apply it?

> Seems fine to keep. There is a discontinuity when one uses return, but it's like that with lots of other features already (we can call it: on-demand monadic semantics for verification).

You mean keep them split? Initially I split them because I thought they didn't apply in the same places (more restrictions for the while with return case), as there's no particular restriction for the normal while loops. But maybe in the end we want to apply the transformations in the same places? In that case I could merge them back (and the monadic transformation would trigger only when there are returns).

> it makes the triggering logic a bit messy

Which part do you mean?

> You should be able to support return inside inner functions as well.

Always ambitious :) I'll try to add it.

> @jad-hamza , you should document the translation with the rst docs, as it is done, e.g., for laws.

Sure I'll add a section or a page.

samarion:
> Wouldn't that have been dangerous for the missing LetRec case? It feels safer to know where we're applying the transformation, maybe we could use guard your general case by the explicit list of trees for which we want to apply it?

Well, I guess we can explicitly list allowed cases if you prefer but that's a lot of cases :) I would expect we'd eventually notice if some tree needs special handling (like other transformations). This risk exists for all transformations but I'm afraid that restricting transformations to expected cases will make the input language fragmented.

> You mean keep them split? Initially I split them because I thought they didn't apply in the same places (more restrictions for the while with return case), as there's no particular restriction for the normal while loops. But maybe in the end we want to apply the transformations in the same places? In that case I could merge them back (and the monadic transformation would trigger only when there are returns).

Hmm, yes I guess the triggering logic is indeed slightly different. I would suggest implementing this by building a `WhileTransformer` which implements the default case (without returns) and having the `ReturnTransformer` extend it. You can then use `WhileTransformer` instead of `NoReturnCheck`.

>> You should be able to support return inside inner functions as well.
>
> Always ambitious :) I'll try to add it.

You'll have to slightly change the logic when computing `exprHasReturn` to make sure returns inside inner functions don't make the transformation trigger on the outer trees. Maybe attach the wrapping function id to the expression and make it a constructor parameter of `ReturnTransformer`.

core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
```diff
@@ -0,0 +1,501 @@
...
+          SimpleWhileTransformer.transform(fa.returnType),
+          newBodyWithSpecs,
+          fa.flags.map(SimpleWhileTransformer.transform)
+        )
+      }
+
+      private def proceedOrTransform(expr: s.Expr, currentType: s.Type): t.Expr = {
+        val currentTypeChecked = SimpleWhileTransformer.transform(currentType)
+        if (exprHasReturn(expr)) transform(expr, currentType)
+        else ControlFlowSort.proceed(retTypeChecked, currentTypeChecked, SimpleWhileTransformer.transform(expr))
```
samarion, line 207: I think this isn't correct. Imagine the case where you have a `LetRec` with a `return` inside of a "normal" while loop.

You need to have `SimpleWhileTransformer` be a class extended by `ReturnTransformer`. Something like
```scala
class SimpleWhileTransformer extends extends TransformerWithType {
  // Keep the current implementation, just ignore the `currentType` parameter in `transform`.
}

// Use this value wherever you're using the SimpleWhileTransformer currently
val noReturnTransformer = new SimpleWhileTransformer

class ReturnTransformer(fa: s.FunAbstraction) extends SimpleWhileTransformer {
  // Mostly keep the current implementation.
  private def proceedOrTransform(expr: s.Expr, currentType: s.Type): t.Expr = {
    val currentTypeChecked = noReturnTransformer.transform(currentType)
    if (exprHasReturn(expr)) {
      transform(expr, currentType)
    } else {
      // Note that we use `transform` here and not noReturnTransformer.transform because there
      // might still be legal return statements inside let-recs within `expr`.
      ControlFlowSort.proceed(retTypeChecked, currentTypeChecked, transform(expr, currentType))
    }
  }
}
```

core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
```diff
@@ -0,0 +1,501 @@
...
+        val s.MatchCase(pattern, optGuard, rhs) = mc
+        t.MatchCase(
+          SimpleWhileTransformer.transform(pattern),
+          optGuard.map(SimpleWhileTransformer.transform),
+          proceedOrTransform(rhs, currentType)
+        ).setPos(mc)
+      }
+
+      override def transform(expr: s.Expr, currentType: s.Type): t.Expr = expr match {
+        case wh @ s.While(cond, body, optInv) if exprHasReturn(body) =>
```
samarion, line 220: I think the case should be
```scala
case wh @ s.While(cond, body, optInv) if exprHasReturn(wh) =>
```

core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
```diff
@@ -0,0 +1,501 @@
...
+              condChecked,
+              t.ApplyLetRec(id, Seq(), tpe, Seq(), Seq()).copiedFrom(wh),
+              ControlFlowSort.proceed(retTypeChecked, t.UnitType(), t.UnitLiteral()).copiedFrom(wh)
+            ).copiedFrom(wh)
+          ).copiedFrom(wh)
+
+        case s.Assert(e, err, rest) =>
+          t.Assert(SimpleWhileTransformer.transform(e), err, transform(rest, currentType)).setPos(expr)
+
+        case s.Assume(e, rest) =>
```
samarion, line 294: You can probably group these all in a single case, something like
```scala
case (_: s.Assert | _: s.Assume | _: s.Lambda | _: s.Forall | ...) =>
  noReturnCheck.transform(expr)
```

Alternatively, if you want to allow returns within nested functions in those trees (since the transformation will deal with them correctly), you can instead do
```scala
case (_: s.Assert | _: s.Assume | _: s.Lambda | _: s.Forall | ...) =>
  if (exprHasReturn(expr))
    throw MalformedStainlessException(...)
  super.transform(expr, currentType)
```

core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
```diff
@@ -0,0 +1,501 @@
...
+          t.Assume(SimpleWhileTransformer.transform(e), transform(rest, currentType)).setPos(expr)
+
+        case s.Return(e) =>
+          ControlFlowSort.ret(
+            retTypeChecked,
+            SimpleWhileTransformer.transform(currentType),
+            SimpleWhileTransformer.transform(e)
+          )
+
+        case s.IfExpr(cond, e1, e2)
```
samarion, line 304: Same as the while case, the match guard should probably be `hasReturnExpr(expr)`

core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
```diff
@@ -0,0 +1,501 @@
...
+          if exprHasReturn(e1) || exprHasReturn(e2) =>
+
+          t.IfExpr(SimpleWhileTransformer.transform(cond),
+            proceedOrTransform(e1, currentType),
+            proceedOrTransform(e2, currentType)
+          ).setPos(expr)
+
+        case s.IfExpr(cond, e1, e2) =>
```
samarion, line 312: Then you can remove this case

core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
```diff
@@ -0,0 +1,501 @@
...
+          t.IfExpr(SimpleWhileTransformer.transform(cond),
+            transform(e1, currentType),
+            transform(e2, currentType)
+          ).setPos(expr)
+
+        case s.MatchExpr(scrut, cases)
```
samarion, line 318: Same comments as for `IfExpr` in the two `MatchExpr` cases.

core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
```diff
@@ -0,0 +1,501 @@
...
+              (v: t.Variable) =>
+                t.exprOps.replaceFromSymbols(
+                  Map(vdChecked -> v),
+                  proceedOrTransform(body, currentType)
+                )(t.convertToVal),
+              body.getPos
+            )
+          ).setPos(expr)
+
+        case s.Let(vd, e, body) =>
```
samarion, line 363:
```scala
case s.Let(vd, e, body) => super.transform(expr, currentType)
```

core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
```diff
@@ -0,0 +1,501 @@
...
+              val (nonReturnEs, others) = es.span(e => !exprHasReturn(e))
+              val nonReturnsEsChecked = nonReturnEs.map(SimpleWhileTransformer.transform(_))
+              if (others.isEmpty)
+                t.Block(nonReturnsEsChecked.init, nonReturnsEsChecked.last).copiedFrom(expr)
+              else
+                t.Block(nonReturnsEsChecked, processBlockExpressions(others)).copiedFrom(expr)
+          }
+          processBlockExpressions(es :+ last)
+
+        case _ =>
```
samarion, line 446: I think this case should only trigger if `exprHasReturn(expr)`

samarion (commented):
Thanks for addressing all my picky comments!

jad-hamza:
> Thanks for addressing all my picky comments!

And thanks for the quick review!

core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
jad-hamza: You're right this doesn't work at the moment, but did you see that `SimpleWhileTransformer` invokes the `ReturnTransformer` under `LetRec`?

core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
jad-hamza: Good point, that's to avoid the "bad" cases going to the wildcard case?

core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
jad-hamza: `Assert` and `Assume` are accepted, but you're right I need to catch the others.

core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
samarion: Oh I hadn't noticed, sorry. In that case, it should be fine to keep the two transformations separate.

core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
samarion: Yes, exactly.

core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
samarion: Oh, right :)

samarion:
Urgh, Github keeps dropping my file comments. Maybe force pushing changes the commit hash the comments get lost?

Anyway, I have a small nit in the `ReturnFinder` default case: if there's a return within a `LetRec` itself within a type, then it will never be visited. I though you previously had `super.traverse(expr, currentId)`, is there a reason to prefer the new variant?

jad-hamza:
> Urgh, Github keeps dropping my file comments. Maybe force pushing changes the commit hash the comments get lost?

I still see the comments but I marked them as resolved

> Anyway, I have a small nit in the `ReturnFinder` default case: if there's a return within a `LetRec` itself within a type, then it will never be visited. I though you previously had `super.traverse(expr, currentId)`, is there a reason to prefer the new variant?

Hm no I'll revert, thanks! I was debugging the issue that `LetRec` expression was not added to the set (I needed to add the expression to the set in the `LetRec` case in addition to the `Operator` case)

romac:
:tada:

PR #927, opened 2021-03-02
update trusted for larabot

vkuncak:
Updating github users who can run larabot


vkuncak:
Settings change seems ok. (Complain on mattermost if you have issues with larabot.)

Issue #917, opened 2021-02-24
Illegal aliasing is allowed by AntiAliasing


brunnerant:
`AntiAliasing` is supposed to report illegal uses of aliasing and crash Stainless in the later case. If I am not mistaken, it is pretty permissive and supports functions returning aliased results if they are non-recursive (it handles those by inlining the function to detect the exact aliasing).
But on the following example, it doesn't report any illegal aliasing, when it actually should:
```Scala
object Bug {
  case class A(var a: BigInt)

  def rec(a1: A, a2: A, n: BigInt): A = {
    if (n == 0) a1
    else rec(a2, a1, n - 1)
  }

  def bug() = {
    val a1 = A(1)
    val a2 = A(2)
    val a = rec(a1, a2, 5)
    a.a = 3
  }
}
```
The code gets translated to (after `AntiAliasing`):
```Scala
def bug: Unit = {
  var a1: A = A(1)
  var a2: A = A(2)
  var a: A = rec(a1, a2, 5)
  a = A(3)
  ()
}
```
Which is wrong, since `a1` and `a2` should be (conditonally) mutated.

jad-hamza:
Thanks for the report! Can you post the full code with `A`?

brunnerant:
I think that the issue might be coming from [there](https://github.com/epfl-lara/stainless/blob/master/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala#L353). Instead of returning an empty set of effects, I think that we should be throwing an exception indicating that the effects of a recursive functions are not known. I tried changing it and it does not seem to break the integration tests, and it makes the specific example that I posted crash, which is what we want.

jad-hamza:
Thanks for the report and the fix!
Fixed in https://github.com/epfl-lara/stainless/pull/928 https://github.com/epfl-lara/stainless/pull/985

PR #928, opened 2021-03-03
Fix issue #917 about allowed illegal aliasing

brunnerant:
This fixes issue #917 and stops stainless when aliasing occurs from a recursive function, since Stainless doesn't support that.

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
@@ -350,7 +350,7 @@ trait EffectsAnalyzer extends oo.CachingPhase {
           .map(rec(_, path))
           .getOrElse(Set.empty)
 
-      case fi: FunctionInvocation => throw MalformedStainlessCode(expr, s"Couldn't compute effect targets in: $expr")
+      case fi: FunctionInvocation => Set.empty
       case (_: ApplyLetRec | _: Application) => Set.empty
       case (_: FiniteArray | _: LargeArray | _: ArrayUpdated | _: MutableMapUpdated) => Set.empty
       case IsInstanceOf(e, _) => rec(e, path)
--- a/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
@@ -350,7 +350,7 @@ trait EffectsAnalyzer extends oo.CachingPhase {
           .map(rec(_, path))
           .getOrElse(Set.empty)
 
-      case fi: FunctionInvocation => Set.empty
+      case fi: FunctionInvocation => throw MalformedStainlessCode(expr, s"Couldn't compute effect targets in: $expr")
       case (_: ApplyLetRec | _: Application) => Set.empty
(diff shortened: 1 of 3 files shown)


brunnerant:
Is there a way to test that a particular file triggers a `MalformedStainlessCodeException`, to verify that Stainless doesn't accept some programs ? It seems that `invalid` benchmarks only test that some verification condition fails, but they still require the program to be valid.

jad-hamza:
> Is there a way to test that a particular file triggers a `MalformedStainlessCode`, to verify that Stainless doesn't accept some programs ? It seems that `invalid` benchmarks only test that some verification condition fails, but they still require the program to be valid.

Thanks for the PR! Yes there is the `extraction/invalid` folder

samarion:
Hmm, recursive functions are supposed to be allowed to return fresh mutable values, e.g. the following example should be supported by the AntiAliasing phase:
```scala
sealed abstract class MutableList
case class MutableCons(var head: Int, tail: MutableList) extends MutableList
case class MutableNil() extends MutableList

def fill(size: Int, value: Int): MutableList[Int] =
  if (size <= 0) MutableNil()
  else MutableCons(value, fill(size - 1, value))
```

In practice, I guess this isn't really a useful paradigm though...

samarion:
As an alternative fix, could you please try adding the following check
```scala
if (isRecursive(fd) && !exprOps.withoutSpecs(fd.fullBody).forall(isExpressionFresh))
  throw ImperativeEliminationException(...)
```
after
https://github.com/epfl-lara/stainless/blob/70764a80f6072bbde4f7e48a1bf5669c1e1e8b0c/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala#L42

You'll have to implement `isRecursive(fd: FunAbstraction)`, I don't think it exists yet in the Stainless code base.

brunnerant:
Sorry, I didn't have time the last few days. But good catch, I didn't think about that case. I have added your check and it indeed works :+1: .

vkuncak:
It looks like this was merged as part of another PR? @jad-hamza , did you do this?

vkuncak:
I tried the example on master ( https://github.com/epfl-lara/stainless/commit/965b0ec485e62715003738e73605bd4dbb5da1da ) and it reports an error now, but it's a stack trace. I noticed in general that effect checker errors are reported very differently than VC gen errors. Can we unify them or is there a limitation in our architecture that only verification can report errors properly?

jad-hamza:
Yes indeed, I forgot to close this.
I got this for the reporting:

```scala
[ Error  ] Issue917.scala:4:3: Illegal recursive functions returning non-fresh result
             def rec(a1: A, a2: A, n: BigInt): A = {
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
[ Error  ] Stainless terminated with an error.
[ Error  ] Debug output is available in the file `stainless-stack-trace.txt`. If the crash is caused by Stainless, you may report your issue on https://github.com/epfl-lara/stainless/issues
[ Error  ] You may use --debug=stack to have the stack trace displayed in the output.
```

You prefer that we hide the "Debug output" message?

I can try fixing discrepancies when they're identified, there shouldn't be architecture issues (except maybe for errors outside Stainless core such as `FragmentChecker.scala`).

jad-hamza:
Merged as part of https://github.com/epfl-lara/stainless/pull/985, thanks!

PR #931, opened 2021-03-11
Clustering

drganam:
Introduces two command line options: --comparefuns and --models, for equivalence checking of specified functions


core/src/main/scala/stainless/extraction/trace/Trace.scala
```diff
@@ -115,19 +170,86 @@ trait Trace extends CachingPhase with SimpleFunctions with IdentitySorts { self
     // body, pre and post for the tactFun
 
     val body = andJoin(Seq(inductScheme, prevBody))
-    val precondition = function.precondition
-    val postcondition = function.postcondition
- 
+    val precondition = exprOps.preconditionOf(function.fullBody)  //function.precondition
+    val postcondition = exprOps.postconditionOf(function.fullBody) //function.postcondition
     val bodyPre = exprOps.withPrecondition(body, precondition)
     val bodyPost = exprOps.withPostcondition(bodyPre,postcondition)
```
jad-hamza, line 176: Could you try to manipulate specifications (getting pre/postconditions, and changing them), using the new `BodyWithSpecs` API?

vkuncak:
@drganam : I just realized that `@traceInduct` uses `.holds` pattern for specifying these statements.
We have moved away from it to make it easier to not inline proofs and only keep lemma statements.
So, in the running example, it would be better to have
```scala
@opaque
def zero_check(arg: BigInt): Boolean = {
  if (arg > 0) zero_check(arg - 1)
  else true
} ensuring (res =>   zero1(arg) == zero2(arg))
```

core/src/main/scala/stainless/extraction/trace/Trace.scala
```diff
@@ -115,19 +170,86 @@ trait Trace extends CachingPhase with SimpleFunctions with IdentitySorts { self
...
+    functions map CheckFilter.fullNameToPath
   }
 
-}
+  private def shouldBeChecked(fid: Identifier): Boolean = pathsOpt match {
+    case None => false
+
+    case Some(paths) =>
+      // Support wildcard `_` as specified in the documentation.
+      // A leading wildcard is always assumes.
```
jad-hamza, line 191: Small typo: `assumed`

core/src/main/scala/stainless/extraction/trace/Trace.scala
```diff
@@ -115,19 +170,86 @@ trait Trace extends CachingPhase with SimpleFunctions with IdentitySorts { self
...
+  private lazy val pathsOptModels: Option[Seq[Path]] = context.options.findOption(optModels) map { functions =>
+    functions map CheckFilter.fullNameToPath
+  }
 
+  private def isModel(fid: Identifier): Boolean = pathsOptModels match {
+    case None => false
+
+    case Some(paths) =>
+      // Support wildcard `_` as specified in the documentation.
+      // A leading wildcard is always assumes.
```
jad-hamza, line 208: Same here for "assumes", perhaps you could factor this code into a common function

core/src/sphinx/trace.rst
```diff
@@ -0,0 +1,84 @@
+.. _trace:
+
+Induction and equivalence checking
+==================================
+
+Induction and @traceInduct annotation
+-------------------------------------
+
+We introduce the @traceInduct annotation for automating proofs using induction. Stainless will transform the annotated lemma by generating the inductive proof, based on the structure of one of the functions that appear in the lemma. This approach is useful for functions that have simple inductive form and are easily mapped into inductive proofs.
```
jad-hamza, line 9: You can surround `@traceInduct` (and other code) with backquotes like this \`\`@traceInduct\`\`.

drganam:
@jad-hamza @vkuncak the latest implementation (with `ensurings` and the new `BodyWithSpecs` API) is here: https://github.com/epfl-lara/stainless/pull/1052

PR #942, opened 2021-03-22
Allow immutable classes with mutable type parameter

jad-hamza:
Does this change make sense?


samarion:
Hmm, this might just work because if there exists a mutable type in `tps` which would appear in a class field type, then the corresponding type parameter would have to be annotated with `@mutable`.

jad-hamza:
> Hmm, this might just work because if there exists a mutable type in `tps` which would appear in a class field type, then the corresponding type parameter would have to be annotated with `@mutable`.

Right, and I've added more checks to make sure we don't instantiate non-mutable type-parameter with mutable types. I think these checks should probably be moved elsewhere, but I'll leave it as is for now.

PR #953, opened 2021-03-26
Remove unsound postcondition on Set.map

### LINKED ISSUE #881: Wrong ensuring clause for map in stainless/lang/Set.scala
Here :
https://github.com/epfl-lara/stainless/blob/0ebb00139688b6214909b939dc7c932b41c99bf8/frontends/library/stainless/lang/Set.scala#L36-L40

Isn't the ensuring clause false if the function f is a non-injective function ?
Shouldn't the clause be like this instead :
```scala
forall((a: A) => (!set.contains(a) || res.contains(f(a))))
&& forall((b: B) => (!res.contains(b) || exists((a: A) =>
  set.contains(a) && f(a) == b
)))
```

Zyfarok:
fixes #881

### DIFF
--- a/frontends/library/stainless/lang/Set.scala
+++ b/frontends/library/stainless/lang/Set.scala
@@ -34,8 +34,6 @@ object Set {
     @extern @pure
     def map[B](f: A => B): Set[B] = {
       new Set(set.theSet.map(f))
-    } ensuring { res =>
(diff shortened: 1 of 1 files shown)


romac:
Superseded by #964

romac:
@Zyfarok Thanks by the way!

PR #956, opened 2021-03-29
Remove size injection from help list

rjraya:
the phase was integrated in SizedADTExtraction


romac:
FYI you are not getting a larabot report because you are not yet in the trusted larabot users list. Will be fixed in https://github.com/epfl-lara/stainless/pull/957.

Issue #961, opened 2021-03-30
`UInt.widen` not yet supported by GenC


deiruch:
Widening of UInt's is not yet supported in  GenC:

```
  [ Fatal  ] ASW\src\Stainless\global\commonFSW.scala:12:10: Unsupported integral cast to 16-bit integer
                     (buf(i).widen[UInt16]<<8) | buf(i+1).widen[UInt16]
                      [31m^^^^^^[0m
```

jad-hamza:
Fixed, thanks! This is now translated to a C cast

Issue #954, opened 2021-03-26
Cannot compare Int to UInt32


deiruch:
The following code fragment does not validate.
```
import stainless.math.BitVectors._
import scala.language.postfixOps

object Constants
{
  val CONSTANT = 4096 * 64
}

case class Test()
{
  def x(x: UInt32): Unit =
  {
    require(x <= Constants.CONSTANT)
  }
}
```

This is the error message:
```
Test.scala:13:18: `intToBV` implicit may only be used on `Int` literals
               require(x <= Constants.CONSTANT)
                            ^^^^^^^^^^^^^^^^^^
```

jad-hamza:
https://epfl-lara.github.io/stainless/purescala.html#bitvectors
You can now do conversions `x.toSigned[Int32].toInt` (or directly `x.toInt`), or conversely you can write `BitVectors.fromInt(Constants.CONSTANT).toUnsigned[UInt32]`
Implicit conversions when doing comparisons or arithmetic operations are not supported at the moment.

Issue #950, opened 2021-03-26
Array indexing does not yet work with UInt8 & Co.


deiruch:
Stainless could support array indexing with UInt8 and other BitVector types.

jad-hamza:
Thanks, now supported!

Issue #948, opened 2021-03-25
(Unsigned) integer types not yet documented


deiruch:
Could the new integer types (e.g. int32) be documented here: https://epfl-lara.github.io/stainless/purescala.html#predefined-types ?

deiruch:
Please describe the following operations :)

`    def widen [X <: BV[_ <: BVKind with Singleton]]: X = { ??? }`
`    def narrow[X <: BV[_ <: BVKind with Singleton]]: X = { ??? }`

jad-hamza:
We now have a little bit of documentation here: https://epfl-lara.github.io/stainless/purescala.html#bitvectors

PR #962, opened 2021-03-30
Add more support for bitvector operations and documentation

jad-hamza:
For issues #950, #954, #961


jad-hamza:
I also added some documentation to address https://github.com/epfl-lara/stainless/issues/948

vkuncak:
@jad-hamza : For documentation, we can think of all these types as finite intervals of mathematical integers. Could you describe the conversion functions from this point of view? This would explain the general cases and would complement the examples that you gave so far.

jad-hamza:
Yes good idea, I've added some explanations. I hope I didn't make mistakes, could someone double check? I'll have a fresh look tomorrow as well.

deiruch:
Very helpful! LGTM

PR #963, opened 2021-03-30
Add Induct case object for `@induct` annotation

romac:
- Add a proper case object to represent the `@induct` annotation
- Remove `@induct` flag on the function and its parameters after `InductElimination` and ensure well-formedness
- Always replace existing measure with measure computed from `@induct` params, even when the type-checker is disabled


jad-hamza (approved):

jad-hamza:
Looks good, thanks! If I understand correctly, the well-formedness exception came from a mismatch between flags in `ValDef`'s and `Variable`'s?

PR #964, opened 2021-03-30
Remove quantified postconditions on Set operations

romac:
Closes: #881

### DIFF
--- a/frontends/benchmarks/verification/valid/SetAxioms.scala
+++ b/frontends/benchmarks/verification/valid/SetAxioms.scala
@@ -3,7 +3,7 @@ import stainless.lang._
 import stainless.collection._
 import stainless.annotation._
 
-object SetAxioms {
+object Iterables {
 
   @extern @pure
   def setToListAxiom[A](set: Set[A], a: A): Unit = {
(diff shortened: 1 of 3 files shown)


frontends/benchmarks/verification/valid/Iterables.scala
```diff
@@ -3,41 +3,78 @@ import stainless.lang._
...
 
+  @extern @pure
+  def setFilterAxiom[A](set: Set[A], p: A => Boolean, a: A): Unit = {
+    ()
+  }.ensuring { _ =>
+    if (set.contains(a) && p(a)) {
+      set.filter(p).contains(a)
+    } else {
+      !set.filter(p).contains(a)
+    }
```
jad-hamza, line 56: Can this be written as `(set.contains(a) && p(a)) == set.filter(p).contains(a)`?

vkuncak:
Oops, I ended up redoing some of this in #1000 ... We should check.

vkuncak:
Ok, so after #1000, all that remains is to port `frontends/benchmarks/verification/valid/SetAxioms.scala` to have call the alternatively named postcondition axioms, to have some extra tests. The other changes are no longer needed.

vkuncak:
After checking some more, this is in fact addressed already in master.

Issue #947, opened 2021-03-24
Wrong imperative translation when mutable object returned (AntiAliasing, effects)


vkuncak:
This verifies on master as of today:
```scala
object Test {
  case class IntRef(var x:Int)
  def f = IntRef(42)

  def g(d1: IntRef, d2: IntRef) = {
    val x1 = d1.x
    d2.x = 33
    assert(d1.x==x1)
  }
  
  def mytest:Unit = {
    val c1 = f
    val c2 = c1
    g(c1, c2)
  }
}
```
Running `Test.mytest` in scala REPL crashes in assertion, as I would expect.

vkuncak:
"The aliasing check in function invocation arguments is pretty useless... 
https://github.com/epfl-lara/stainless/blob/43aa5c786ca5b969e442e8078b1c38fa3dae3a41/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L437

aliasing can occur with fields, e.g. a.b and a.b are aliased which won't be caught by the check
We can probably do something like finding conservative targets using `Try(getTargets(e)).toOption.getOrElse(variablesOf(e))`
"

PR #965, opened 2021-03-31
More checks of argument aliasing

jad-hamza:
For issue https://github.com/epfl-lara/stainless/issues/947

@samarion I tried your suggestion of using `Try(getTargets(e)).toOption.getOrElse(exprOps.variablesOf(arg).map(v => Target(v, None, Path.empty))))` to identify aliases. Did you then want to check whether a target of one argument is a prefix of a target of another argument?

For issue #947, this doesn't seem to work because the targets of `c1` are `Some(Target(c1, None, <empty>))` and the targets of `c2` are `Some(Target(c1, None, <empty>))`. Should `getTargets` on `c2` also return `c1`?

We might have the same issue with `getTargets` returns `None`, because `variablesOf` will not capture the transitive dependencies.

### DIFF
--- a/core/src/main/scala/stainless/MainHelpers.scala
+++ b/core/src/main/scala/stainless/MainHelpers.scala
@@ -186,6 +186,12 @@ trait MainHelpers extends inox.MainHelpers { self =>
       def watchRunCycle() = try {
         baseRunCycle()
       } catch {
+        case e @ extraction.MalformedStainlessCode(tree, msg) =>
+          reporter.debug(e)(frontend.DebugSectionStack)
+          ctx.reporter.error(tree.getPos, msg)
+          reporter.error("There was an error during the watch cycle")
+          reporter.reset()
+          compiler = newCompiler()
         case e: Throwable =>
           reporter.debug(e)(frontend.DebugSectionStack)
           reporter.error("There was an error during the watch cycle")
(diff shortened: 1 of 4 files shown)


samarion:
> I tried your suggestion of using Try(getTargets(e)).toOption.getOrElse(exprOps.variablesOf(arg).map(v => Target(v, None, Path.empty)))) to identify aliases.

You might want to filter out variables that don't have a mutable type in the `orElse` case.

> Did you then want to check whether a target of one argument is a prefix of a target of another argument?

Yes, that sounds right.

> For issue #947, this doesn't seem to work because the targets of c1 are Some(Target(c1, None, <empty>)) and the targets of c2 are Some(Target(c1, None, <empty>)).

You probably need to apply rewritings with `exprOps.replaceFromSymbols(env.rewritings, arg)` before calling `getTargets` and `variablesOf`. That should hopefully fix the issue of transitive dependencies. If it doesn't, then we probably have a more general problem with aliasing in let-bindings.

jad-hamza:
Thanks for the quick review! The rewritings make this example work correctly (make it fail as it should).

jad-hamza:
I disabled the aliasing check for accessors, otherwise it was making simple updates such as `x = x + 1` for a `var x` in a trait fail.

Issue #960, opened 2021-03-30
Report Errors and Warning in a GCC/CLANG-compatible style


deiruch:
To let IDEs automatically pick up the output of Stainless, it should output its error and warning messages this way:

```
...
Path/to/file/SourceCode.scala:15:8: error: This is an error description GCC- and CLANG-style for the 8th character on line 15
...
Path/to/file/SourceCode.scala:3:1: warning: And this is a mere warning
...
```

deiruch:
Tested that Visual Studio 2019 picks these up.

vkuncak:
@jad-hamza : this is a surprisingly lightweight to get basic IDE linkup. We should have done it long ago.

vkuncak:
I think we should just change ``--no-color`` to 
  * remove [WARNING] etc. headers at the beginning of lines and use instead this standardized formatting that works with IDEs
  * remove colors (which the existing option apparently does not quite achieve now)
  * remove the extended ASCII symbols for table boundaries, possibly also use commas or semicolons to separate table entries instead of formatted spaces

One can start with step one. You should be able to test it using `compile` command of emacs. For me, it works with `scalac` which displays errors in this format:

`Test.scala:9: error: SomeTextOfErrorMessage`
 
and then I can click with a mouse in emacs to get to the line.

Maybe we will change the name of the option currently, but it makes sense as it is if you interpret "color" more broadly as having output with no coloring, thus helping tools process it.

PR #966, opened 2021-03-31
Propagate option optNoColors to Inox

jad-hamza:
Fix https://github.com/epfl-lara/stainless/issues/960
Depends on https://github.com/epfl-lara/inox/pull/131


vkuncak:
@samarion could we merge this, please?

Issue #955, opened 2021-03-28
Code duplication in FunctionClosure


jad-hamza:
In the example below, we duplicate the definition of `x`, and the first occurrence of `x` in the transformed `g` is not well-defined (because `z > 0` comes after). I wonder why it goes through, this also needs to be investigated.

Any idea on how to avoid code duplication in `FunctionClosure` when lifting inner functions?
Especially when the definition `x`: 
1) depends on a precondition of the top-level function
2) is used in the precondition of the inner function
3) is used in the body of the inner function

To avoid the duplication, what do you think about allowing multiple `require` (interleaved) with `val`'s? (we might have already had this discussion?)

```scala
object Test {
  def f(a: Int, z: Int) = {
    require(0 <= a && a < 100 && z < 0 && z < 100)
    val x = 2 * a / z
    def g(b: Int) = {
      require(x < b && b < 100)
      x + 2
    }
  }
}
```

```scala
[ Debug  ] Symbols before FunctionClosure
[ Debug  ] 
[ Debug  ] -------------Functions-------------
[ Debug  ] def f(a: Int, z: Int): Unit = {
[ Debug  ]   require(((0 <= a && a < 100) && z < 0) && z < 100)
[ Debug  ]   val x: Int = (2 * a) / z
[ Debug  ]   def g(b: Int): Int = {
[ Debug  ]     require(x < b && b < 100)
[ Debug  ]     x + 2
[ Debug  ]   }
[ Debug  ]   ()
[ Debug  ] }
[ Debug  ] 
[ Debug  ] 
[ Debug  ] Symbols after FunctionClosure
[ Debug  ] 
[ Debug  ] -------------Functions-------------
[ Debug  ] def f(a: Int, z: Int): Unit = {
[ Debug  ]   require(((0 <= a && a < 100) && z < 0) && z < 100)
[ Debug  ]   val x: Int = (2 * a) / z
[ Debug  ]   ()
[ Debug  ] }
[ Debug  ] 
[ Debug  ] @derived(f)
[ Debug  ] def g(a: Int, z: Int, b: Int): Int = {
[ Debug  ]   val x: Int = (2 * a) / z
[ Debug  ]   require((0 <= a && (a < 100 && (z < 0 && (z < 100 && {
[ Debug  ]     val x: Int = (2 * a) / z
[ Debug  ]     true
[ Debug  ]   })))) && (x < b && b < 100))
[ Debug  ]   x + 2
[ Debug  ] }
```

PR #968, opened 2021-04-05
Change BodyWithSpecs API to allow multiple require's

jad-hamza:
cc: @vkuncak @gsps 

Fix https://github.com/epfl-lara/stainless/issues/955


jad-hamza:
I'll merge this for now, please tell me if you'd like to see some things reverted!

PR #969, opened 2021-04-06
Make targets empty when path doesn't match

jad-hamza:


jad-hamza:
@samarion Merging this for now, I'll revert if this blows up something :)

samarion:
This breaks `isExpressionFresh` which will now return true for `A(a)`

PR #985, opened 2021-04-13
Add empty targets for basic trees

jad-hamza:


core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -405,6 +454,8 @@ trait EffectsAnalyzer extends oo.CachingPhase {
         if (be.receiver == vd.toVariable) ee.append(be) else be
       }
 
+    case _ if !symbols.isMutableType(expr.getType) => Set.empty
```
samarion, line 457: Doesn't this subsume almost all cases you added above?

jad-hamza: Right, except for containers?

I also reverted to the old `isExpressionFresh`.

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
```diff
@@ -42,6 +42,10 @@ trait EffectsChecker { self: EffectsAnalyzer =>
       checkEffectsLocations(fd)
       checkPurity(fd)
 
+      // Recursive functions must return fresh results so that no aliasing is possible
+      if (symbols.isRecursive(fd.id) && !exprOps.withoutSpecs(fd.fullBody).forall(isExpressionFresh))
```
samarion, line 46: It seems we assume `LetRecs` to be fresh (also when non-recursive), so we need to check this here as well.

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
```diff
@@ -257,7 +261,53 @@ trait EffectsChecker { self: EffectsAnalyzer =>
...
+
+        //MutableMapDuplicate returns a fresh duplicate by definition
+        case MutableMapDuplicate(IsTyped(_, MutableMapType(from, to))) =>
+          !isMutableType(from) && !isMutableType(to)
+
+        // snapshots are fresh
+        case Snapshot(e) => true
+
+        case Let(vd, e, b) => rec(e, bindings) && rec(b, bindings + vd)
+        case LetVar(vd, e, b) => rec(e, bindings) && rec(b, bindings + vd)
```
samarion, line 300: This is wrong (in combination with the `Block` case). You could have
```scala
var a = A(B(1))
a = A(b)
a
```

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
```diff
@@ -257,7 +261,53 @@ trait EffectsChecker { self: EffectsAnalyzer =>
...
+
+        case Block(_, e) => rec(e, bindings)
+
+        //any other expression is conservately assumed to be non-fresh if
+        //any sub-expression is non-fresh (i.e. an if-then-else with a reference in one branch)
```
samarion, line 305: You could add `IfExpr` as a dedicated case since you can ignore the condition.

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
```diff
@@ -257,7 +261,53 @@ trait EffectsChecker { self: EffectsAnalyzer =>
...
+
+        case FiniteArray(elems, _) => elems.forall(rec(_, bindings))
+        case LargeArray(elems, default, _, _) => elems.forall(p => rec(p._2, bindings)) && rec(default, bindings)
+
+        // We assume `old(.)` is fresh here, although such cases will probably be
+        // rejected later in `ImperativeCleanup`.
+        case Old(_) => true
+
+        //function invocation from accessors are not fresh
+        case FunctionInvocation(id, _, _)
```
samarion, line 277: Is this special case really needed? `A(B(1)).b` is still fresh, no?

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -394,6 +394,24 @@ trait EffectsAnalyzer extends oo.CachingPhase {
...
+
+    case FiniteBag(elements, tpe) => Set.empty
+    case BagUnion(s1, s2) => Set.empty
+    case BagIntersection(s1, s2) => Set.empty
+    case BagDifference(s1, s2) => Set.empty
+    case MultiplicityInBag(element, bag) => Set.empty
+    case BagAdd(bag, element) => Set.empty
+
+    case FiniteMap(pairs, default, from, to) => Set.empty
+    case MapUpdated(m, k, v) => Set.empty
```
samarion, line 413: I don't understand, why doesn't this have targets? Same for the `ArrayUpdated` above, about `a.updated(0, B(1))(0)`?
I also think that `FiniteArray` and `FiniteMap` probably have targets (so they should throw exceptions).

jad-hamza: For `FiniteMap` we don't compute targets for `MapApply` so I thought this would be safe?
For arrays you're right it depends on the path, perhaps we can say the targets are empty when the path doesn't start with `ArrayAccessor`? I also have a special case in another branch, where you can compute the exact target if you have a `FiniteArray` with a concrete `ArrayAccessor` index.

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
jad-hamza: Right, I noticed we don't compute targets for `LetVar` either. Should we just return `false` here?

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
jad-hamza: I added these changes for arrays

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
jad-hamza: Added, thanks

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
jad-hamza: Fixed, thanks

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
jad-hamza: Fixed, thanks

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
samarion: I guess you could return `rec(b, bindings)`?

I just realized that `Let` is also non-trivial, what about the following case:
```scala
val a = A(B(C(1)))
a.b = B(c)
a
```

Maybe both cases are not relevant because they correspond to unsupported cases of aliasing? Do we correctly check this?

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
samarion: > For FiniteMap we don't compute targets for MapApply so I thought this would be safe?

Hmm, I feel like we're rather missing the effects computation for `MapApply`. Unless we check that we never mutate values inside the map?

> For arrays you're right it depends on the path, perhaps we can say the targets are empty when the path doesn't start with ArrayAccessor?

Yes, that sounds reasonable. For some reason I felt that we needed to be more strict but I think you're right, checking the path should be enough. Note that we probably have to throw an exception if the path is `ArrayAccessor` (unless we have a constant index and can actually do the array lookup).

jad-hamza: So we need to add a new kind of accessors for immutable maps? I'll open an issue to keep track of this

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
jad-hamza: I put `rec(b, bindings)` for `Let` and `LetVar`

samarion: So it seems that we can keep `Let` because we check freshness during field assignment: https://github.com/epfl-lara/stainless/blob/391088ee6ea1154db23e1320d8f637b350c593c0/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala#L103
Maybe just add a comment about why it's safe so we don't have to think about it again next time? :)

However, we don't check freshness in `Assignment` so `LetVar` should indeed be `rec(b, bindings)`.

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
samarion: Yes, that SGTM.

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
jad-hamza: Fixed, thanks

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
jad-hamza: Ok thanks, I opened this: https://github.com/epfl-lara/stainless/issues/991

PR #996, opened 2021-04-19
tips on --no-colors and emacs

vkuncak:
Documentation: advice on using --no-colors inside installation tips. Also concrete illustration with emacs commands.


vkuncak:
Documentation is always good

PR #999, opened 2021-04-20
Move several collections to bolts repo

vkuncak:


vkuncak:
Moved out some collections to bolts to have less code to extract by default

PR #1000, opened 2021-04-22
Library cleanup: remove quantifiers from posts in Set and Map (fixed two unsound ones for function image). Option to load library files from a config file.

vkuncak:
Currently library files are computed in build.sbt and put into auto-generated top-level file.
This change instead puts this list of files into a config file, allowing user to modify library dependencies.


vkuncak:
@romac @gsps : does this configurability of paths look like a good idea? Now I added a fallback onto the precomputed list of paths if the config file is not found, so I think it should not result in too mysterious behavior in practice. I also added an explanation in the docs.

gsps:
What's the use case, exactly? Do you ever want to include any library files outside of stainless' usual standard library? If the answer is no, I think I'd prefer some command-line flag that acts as a filter on the library file paths. Wdyt, @samarion ?

vkuncak:
> What's the use case, exactly? Do you ever want to include any library files outside of stainless' usual standard library? If the answer is no, I think I'd prefer some command-line flag that acts as a filter on the library file paths. Wdyt, @samarion ?

I'd also like to be able to switch versions of the library for larger projects. Command line processing is between stainless and scalac and this is solved early on with the proposed solution. It should also work with packaging files, I think, but should test.

samarion:
> Wdyt, @samarion ?

I kind of agree with Georg, I feel like controlling this with a flag would make more sense. You should be able to change the `Main.libraryFiles` val into a def which takes an implicit `inox.Context` to access the flag value. I would let the file point to the path of the `libfiles.txt` you want to use.

Also, instead of using a hard-coded boolean to control the regeneration of `libfiles.txt`, you could create an sbt task which takes as argument the path to which the generated library files should be output.

Issue #1007, opened 2021-04-27
GenC: avoid name mangling for some types


deiruch:
Similar to #921 there should be a way to prevent GenC from name mangling types.

```scala
case class MemStream(var data: Array[UInt8], var pos: Int, var limit: Int, var error: Boolean)
{
	@export
	def stream_seek(_offset: Int):Unit =
	{
		//...
	}
}
```

results in the following C code:

```c
typedef struct {
  array_uint8 data_0;
  int32_t pos_5;
  int32_t limit_0;
  bool error_0;
} MemStream_1;

static void stream_seek(MemStream_1* thiss_2, int32_t _offset_0);
```

Because we'd like to interface manually written code with the generated code, the resulting type should be called `MemStream` instead of `MemStream_1`. This is only important for types that are passed to/from `@export`-ed functions.

jad-hamza:
Is it ok if you have to manually annotate `MemStream` as `@export`?

deiruch:
Sure. In facts, that's the first thing I tried :)

PR #1008, opened 2021-04-27
Add support for GenC @export annotation on types

jad-hamza:
Fix https://github.com/epfl-lara/stainless/issues/1007

@deiruch Should we also prevent name mangling for `val`'s and `var`'s of `@export`'ed classes?

### DIFF
--- a/core/src/main/scala/stainless/extraction/package.scala
+++ b/core/src/main/scala/stainless/extraction/package.scala
@@ -26,6 +26,7 @@ import scala.language.existentials
 package object extraction {
 
   val phases: Seq[(String, String)] = Seq(
+    "UserFiltering"             -> "Removes all the library functions not used by the user",
     "Preprocessing"             -> "A preprocessing phase before the pipeline",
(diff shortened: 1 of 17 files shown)


deiruch:
> @deiruch Should we also prevent name mangling for `val`'s and `var`'s of `@export`'ed classes?

I don't think it's scrictly necessary right now, but it'd be consistent and will give us more options for the future.

jad-hamza:
Should be good now, but it has not been tested extensively :)
Please tell me if there's something wrong!

deiruch:
It works *almost* as it should! :)

It seems that all functions in an annotated type get exported - even including `inv()`. I think it should still only export explicitly annotated functions.

jad-hamza:
Hmm I thought I disabled the propagation of `@export`. Are you sure all functions get exported, or is it only the special function `inv`?

deiruch:
Sorry, you're absolutely right. It's just `inv`.

jad-hamza:
Thanks for the report! Fixed in https://github.com/epfl-lara/stainless/pull/1019

PR #1010, opened 2021-04-27
Recommend using Z3 4.8.10

jad-hamza:
* Test suite is changed to use `smt-z3:z3-4.8.10` (needs `z3-4.8.10` in path)
* Bundle Z3 4.8.10 in our releases
* Refer to Z3 4.8.10 in the documentation


jad-hamza:
Already merged

PR #1016, opened 2021-04-29
Catch and fail on mutation of immutable fields

### LINKED ISSUE #1014: Mutation of non-var fields is erased
Currently, if a field of a case class is mutated, that is not declared as `var`, the mutation is just erased by the `AntiAliasing` phase. However, that alters the semantics of a program, so it should at least generate a warning:

```scala
 @sealed
 sealed case class A(b: Boolean)

 def set_false(var0: A): A = {
   var (a: A) @var = var0
   a.b = false
   a
 }
```

Stainless trees:
```
 Symbols before AntiAliasing
 -------------Functions-------------
 def set_false((var0: A) @pure): A = {
   var (a: A) @var = var0
   a.b = false
   a
 }
 -------------Functions-------------
 def set_false((var0: A) @pure): A = {
   var (a: A) @var = var0
   ()
   a
 }
```

yannbolliger:
Closes #1014.

The original solution suggested by @samarion (thanks!) was that `getType` of `FieldAssignment` returns `Untyped` if the field is not annotated as `@var`. 

That approach was dropped in favour of an actual `ensureWellFormedFunction` check, as suggested by @romac.

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/Definitions.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/Definitions.scala
@@ -7,10 +7,23 @@ package imperative
 trait Definitions extends oo.Trees { self: Trees =>
   override type Symbols >: Null <: AbstractSymbols
 
-  trait AbstractSymbols
-    extends super.AbstractSymbols
-      with SymbolOps
-      with TypeOps { self0: Symbols =>
+  trait AbstractSymbols extends super.AbstractSymbols with SymbolOps with TypeOps {
+    self0: Symbols =>
+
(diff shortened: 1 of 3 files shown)


yannbolliger:
After a new suggestion by @romac, I'll try another solution that performs the check in `ensureWellFormedFunction` of the imperative phase and throws `NotWellFormedException`.

vkuncak:
So this is still WIP, @yannbolliger ?

yannbolliger:
> So this is still WIP, @yannbolliger ?

Yes, somehow I couldn't change this PR to _draft_ state however.

yannbolliger:
The new approach is now implemented.

PR #1017, opened 2021-04-29
Move to ScalaZ3 4.8.10

jad-hamza:
Missing Mac package for now


jad-hamza:
I think the test suites run for pull requests all run `smt-z3` now, I'll add one with `nativez3` to see what happens.

vkuncak:
@mario-bucev we seem to have leapfrogged this PR by using latest z3 version. Is there anything useful here that we should consider including?

mario-bucev:
No, not really

PR #1018, opened 2021-04-30
Print verification progress but not verification conditions

jad-hamza:
Opened this after closing https://github.com/epfl-lara/inox/pull/143

This PR displays a counter `Verified: 60 / 87` that gets updated on the same line.
What do people think? It's not very pretty with the global variables, but it seems to work :)
In  `SplitCallBack`, the denominator will increase with each new subprogram.
When starting a new watch cycle, we reset the counters.


jad-hamza:
Merging this for now, please try it and tell me if you want any modification!

PR #1029, opened 2021-05-04
Fix all case object flags in NoxtFrontend

yannbolliger:
This is a small _maintenance_ PR that finally adds _all_ the flags the Rust frontend could possibly emit to the `FIXME` pattern match in `NoxtFrontend`.

### DIFF
--- a/frontends/stainless-noxt/src/main/scala/stainless/frontends/noxt/NoxtFrontend.scala
+++ b/frontends/stainless-noxt/src/main/scala/stainless/frontends/noxt/NoxtFrontend.scala
@@ -69,40 +69,43 @@ object NoxtFrontend {
       case _ => super.transform(pat)
     }
 
+
+    val flagsByName = Seq(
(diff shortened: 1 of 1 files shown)


frontends/stainless-noxt/src/main/scala/stainless/frontends/noxt/NoxtFrontend.scala
```diff
@@ -72,27 +72,37 @@ object NoxtFrontend {
...
+        xt.IsSealed,
+        xt.StrictBV,
+        xt.Synthetic,
+        xt.Template,
+        xt.TraceInduct,
+        xt.Unchecked,
+        xt.ValueClass,
+        xt.IsVar,
+        xt.Wrapping
+      ).map(f => (f.name, f)).toMap.getOrElse(flag.name, super.transform(flag))
```
romac, line 105: Can we pre-compute this map somewhere?

PR #1033, opened 2021-05-05
Add fresh copy primitive

yannbolliger:
This PR adds a new primitive operation (as a tree) to Stainless. The new operation called `freshCopy` (naming ist still wip, your suggestion is welcome) is an exact duplicate of `snapshot` except that it can be used outside of the `@ghost` context.

There is no Scala code extraction for the primitive for now, because it is used via the Rust frontend. There, it solves a problem with AntiAliasing https://github.com/epfl-lara/rust-stainless/issues/154.

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
@@ -289,8 +289,9 @@ trait EffectsChecker { self: EffectsAnalyzer =>
         case MutableMapDuplicate(IsTyped(_, MutableMapType(from, to))) =>
           !isMutableType(from) && !isMutableType(to)
 
-        // snapshots are fresh
-        case Snapshot(e) => true
+        // snapshots & freshCopies are fresh
+        case Snapshot(_) => true
+        case FreshCopy(_) => true
 
         // For `Let`, it is safe to add `vd` as a fresh binding because we disallow
         // `FieldAssignments` with non-fresh expressions in `check(fd: FunAbstraction)` above.
--- a/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
@@ -289,7 +289,7 @@ trait EffectsChecker { self: EffectsAnalyzer =>
         case MutableMapDuplicate(IsTyped(_, MutableMapType(from, to))) =>
           !isMutableType(from) && !isMutableType(to)
 
-        // snapshots & freshCopies are fresh
+        // snapshots & fresh copies are fresh
         case Snapshot(_) => true
         case FreshCopy(_) => true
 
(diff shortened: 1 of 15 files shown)


core/src/main/scala/stainless/utils/Serialization.scala
```diff
@@ -98,7 +98,7 @@ class XLangSerializer(override val trees: extraction.xlang.Trees, serializeProdu
     *
     * The new identifiers in the mapping range from 180 to 259.
```
jad-hamza, line 99: 180 to 260

core/src/main/scala/stainless/extraction/imperative/GhostChecker.scala
```diff
@@ -45,6 +45,7 @@ trait GhostChecker { self: EffectsAnalyzer =>
       case Decreases(_, body) => isGhostExpression(body)
 
       case Snapshot(e) => true
+      case FreshCopy(_) => false
```
jad-hamza, line 48: Should we maybe recurse down through the argument of `FreshCopy`?

jad-hamza (approved):
Looks good to me!
Could you perhaps add in the library a `freshCopy` function (with extraction in scalac) and a few examples in `frontends/benchmarks/imperative/valid` that we expect to work with `freshCopy` to make sure everything works?

core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
```diff
@@ -289,8 +289,9 @@ trait EffectsChecker { self: EffectsAnalyzer =>
         case MutableMapDuplicate(IsTyped(_, MutableMapType(from, to))) =>
           !isMutableType(from) && !isMutableType(to)
 
-        // snapshots are fresh
-        case Snapshot(e) => true
+        // snapshots & freshCopies are fresh
```
romac, line 292:
```suggestion
        // snapshots & fresh copies are fresh
```

jad-hamza (approved):

jad-hamza:
Looks good, I'll leave open for a little while in case others have comments.

romac:
Looks good! Would be good to have an invalid test case which does not use `freshCopy`.

jad-hamza:
Could you also perhaps add this example (in `valid`) https://github.com/epfl-lara/rust-stainless/issues/154 with an additional caller to `set_field` to make sure it behaves correctly (not modifying the argument), and also checking that the result of `set_field` is as expected?

yannbolliger:
Discussion with @vkuncak @romac @gsps:
 - `freshCopy` should be a deep copy

jad-hamza:
Can you show an example that fails for the deep copy issue? We should also fix it for the `Snapshot` tree.

samarion:
> freshCopy should be a deep copy

You could just remove `freshCopy` from the Stainless library and keep it as an internal tree for the Rust frontend.

Otherwise, maybe there's a Scala type class which gives you a deep copy method?

> We should also fix it for the Snapshot tree.

Snapshot should be fine since it's ghost so the library placeholder doesn't need to respect Scala semantics.

jad-hamza:
We could keep `freshCopy` in the library (to be able to easily test it) with the implementation `(??? : T)`

samarion:
> We could keep freshCopy in the library (to be able to easily test it) with the implementation (??? : T)

Yeah, I guess that would be ok. In that case, we should also have a comment in the library which warns this is only for tests.  
I wonder if we could use Viktor's new library configuration to inject the `freshCopy` Scala placeholder only in the relevant test suite.

Issue #1024, opened 2021-05-03
Add support for RefinementType in unapply pattern


jad-hamza:
This file fails:
https://github.com/epfl-lara/bolts/blob/4cf535349deca6f2d27fefe8ba5490e1ae0f7a83/data-structures/seqs/ArrayList.scala

samarion:
You can either
1. add support for dependent types in the `instantiation` method, or
2. erase them with `getType` inside [unapplyAccessor](https://github.com/epfl-lara/stainless/blob/43aa5c786ca5b969e442e8078b1c38fa3dae3a41/core/src/main/scala/stainless/ast/Expressions.scala#L164).

The first option will give better support for dependent type inference, but might be a bit more complicated.

rjraya:
This is the very last thing that went in before crashing:

```
fd...............
@library
def isEmpty[T](thiss: Option[T]): { arg: Boolean | (BigInt("0") <= OptionPrimitiveSize[T](thiss)) } = thiss match {
  case Some(_) =>
    false
  case None() =>
    true
}
unapp............
typed def unapply[BigInt]
```

samarion:
Well those two types certainly look unifiable to me. The only issue I can think of is an id mismatch for the `Option` type, but I don't know where it would come from. Maybe some issue with caching?

rjraya:
It looks like chasing down in Inox and Stainless does the job. I will now try to see how to do it modifying the instantiation method.

rjraya:
It seems like the culprit is rather:  `unapplyAccessorResultType` and its missing getType in `typeOps.instantiateType(fd.returnType.getType, _)`.

PR #1046, opened 2021-05-11
Make unapply patterns work again in some benchmark

rjraya:
Seems to solve #1024 at least in Count.scala, need to check others benchmarks changed.

### DIFF
--- a/core/src/main/scala/stainless/ast/Expressions.scala
+++ b/core/src/main/scala/stainless/ast/Expressions.scala
@@ -166,7 +166,7 @@ trait Expressions extends inox.ast.Expressions with Types { self: Trees =>
       .filter(_.params.size == 1)
       .getOrElse(throw extraction.MalformedStainlessCode(up, "Invalid unapply accessor"))
     val unapp = up.getFunction
-    val tpMap = s.instantiation(fd.params.head.tpe, unapp.returnType)
+    val tpMap = s.instantiation(fd.params.head.tpe.getType, unapp.returnType.getType)
       .getOrElse(throw extraction.MalformedStainlessCode(up, "Unapply pattern failed type instantiation"))
     fd.typed(fd.typeArgs map tpMap).applied(Seq(unapplied))
   }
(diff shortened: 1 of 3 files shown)


core/src/main/scala/stainless/ast/Expressions.scala
```diff
@@ -166,7 +166,7 @@ trait Expressions extends inox.ast.Expressions with Types { self: Trees =>
       .filter(_.params.size == 1)
       .getOrElse(throw extraction.MalformedStainlessCode(up, "Invalid unapply accessor"))
     val unapp = up.getFunction
-    val tpMap = s.instantiation(fd.params.head.tpe, unapp.returnType)
+    val tpMap = s.instantiation(fd.params.head.tpe.getType, unapp.returnType.getType)
```
samarion, line 169: Nit: you can directly call `getType` on `ValDef` and `FunDef`.

So you can write `fd.params.head.getType` and `unapp.getType` which is a bit clearer IMO.

(And same remark in `unapplyAccessorResultType`)

core/src/main/scala/stainless/ast/TypeOps.scala
```diff
@@ -12,9 +12,9 @@ trait TypeOps extends inox.ast.TypeOps {
     lookupFunction(id)
       .filter(_.params.size == 1)
       .flatMap { fd =>
-        instantiation(fd.params.head.tpe, inType)
+        instantiation(fd.params.head.tpe.getType, inType.getType)
           .filter(tpMap => fd.typeArgs forall (tpMap contains _))
-          .map(typeOps.instantiateType(fd.returnType, _))
+          .map(typeOps.instantiateType(fd.returnType.getType, _))
```
samarion, line 17: You can leave it as `fd.returnType` here, and rather use `getType` inside `patternIsTyped` below.

This isn't super important, but it's best to only call `getType` when absolutely needed since it erases potentially useful specifications each time.

PR #1050, opened 2021-05-12
Simplest value API + inlined precondition numbering

jad-hamza:
@samarion Is `simplestValue` for arrays ok here? Or do we need to check the base type is non-empty?


samarion:
Looks fine to me, we can always create an empty array even when the base type is empty.

PR #1060, opened 2021-05-19
Move Rust interop generator from Inox

yannbolliger:
To make Inox completely frontend-independent, the tool that generates AST trees for Rust is moved to Stainless.
This is the dual of https://github.com/epfl-lara/inox/pull/155.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -233,8 +233,7 @@ val scriptSettings: Seq[Setting[_]] = Seq(
 def ghProject(repo: String, version: String) = RootProject(uri(s"${repo}#${version}"))
 
 // lazy val inox = RootProject(file("../inox"))
-// lazy val inox = ghProject("https://github.com/epfl-lara/inox.git", "ec3b77f97da1acf46af21766fb0d845c7f2d9834")
-lazy val inox = ghProject("https://github.com/yannbolliger/inox.git", "4ab8e63213cf52adb322ddeb16c4c0b07bcfe4b8") // rust-interop, latest commit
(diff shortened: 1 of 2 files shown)


samarion:
LGTM, but maybe @gsps should take a look?

gsps:
Re https://github.com/epfl-lara/inox/pull/155#discussion_r634964616 :

> I hadn't realized this was going to live inside Inox. Is there any point having it here rather than in Stainless? I'd prefer if Inox stayed mostly frontend-independent (if possible).

Well, the original intention was to follow the structure of the existing serializer: implement all the Inox-related parts in Inox, and the additions needed to deal with Stainless trees in Stainless. That way we could generate the Rust-interop code for Inox alone as well.

That being said, this code is quite ad-hoc and doesn't actually separate things that cleanly, so I sympathize with the desire to localize it :-)

vkuncak:
@samarion : can we merge this into the rust branch?

samarion:
> @samarion : can we merge this into the rust branch?

You mean in Stainless? Yes, sure.

> Well, the original intention was to follow the structure of the existing serializer: implement all the Inox-related parts in Inox, and the additions needed to deal with Stainless trees in Stainless. That way we could generate the Rust-interop code for Inox alone as well.

I guess it could make sense to introduce Inox bindings for other languages, but maybe we could split them out into their own projects? I would expect the Rust bindings to also include some Rust code which allows idiomatic interaction with Inox.

gsps:
Yes, that's a fair point. Currently both the hand-written and the auto-generated Rust parts live in [a crate that's part of rust-stainless](https://github.com/epfl-lara/rust-stainless/tree/master/stainless_data). For the time being, I'm happy to merge this PR and keep it all in Stainless for now.

PR #1064, opened 2021-05-20
Support tuples with mutable types

### LINKED ISSUE #1063: Tuples cannot contain mutable types
As there is no `TupleAccessor` in the [`EffectsAnalyzer`](https://github.com/epfl-lara/stainless/blob/master/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala#L162-L180) it is currently impossible to have mutable types in tuples and access fields.

The following example:
```scala
object MutableTuple {
  case class Foo(var value: BigInt)
  case class Bar(var value: BigInt)

  def t1(pair: (Foo, Bar)) = {
    pair._1.value = 1
    assert(pair._1.value == 1)
  }
}
```

Fails with: `Couldn't compute effect targets in (class inox.ast.Expressions$TupleSelect): pair._1`

yannbolliger:
Closes #1063.

- Add benchmark for tuple with mutable types
- Implement tuple field accessor

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
@@ -179,6 +179,10 @@ trait EffectsAnalyzer extends oo.CachingPhase {
     def asString(implicit ctx: inox.Context) = s"MutableMapAccessor(${index.asString})"
   }
 
+  case class TupleFieldAccessor(index: Int) extends Accessor {
+    def asString(implicit ctx: inox.Context) = s"TupleFieldAccessor($index)"
+  }
+
   case class Path(path: Seq[Accessor]) {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -356,6 +370,13 @@ trait EffectsAnalyzer extends oo.CachingPhase {
         Set.empty
     }
 
+    case Tuple(exprs) => path match {
+      case TupleFieldAccessor(idx) +: rest =>
+        getTargets(exprs(idx), rest)
```
jad-hamza, line 375: `-1` missing here? Indices in `TupleSelect` start at `1` or `0`?

jad-hamza:
Thanks for the implementation! I'll merge for now, if anyone has comments, please tell me and I'll fix them or revert.

Issue #297, opened 2018-08-14
Replace custom VC cache implementation with H2's MVStore


romac:
http://www.h2database.com/html/mvstore.html

romac:
Currently running into:

> [error] (stainless-core/compile:compileIncremental) java.lang.AssertionError: assertion failed: Can't convert from REF(class EvaluatorRun$FunctionStatus) to REF(trait Product)
> [error] in unit EvaluatorComponent.scala at source=stainless/evaluators/EvaluatorComponent.scala,line=124,offset=4264

which might be related to https://github.com/scala/bug/issues/10815, with no fix in view.

romac:
For large projects, where the cache can easily grow over 2GB, the performance of the current caching backend really suffers and has quite an impact on overall verification time (5-10s of cache loading for each run).

I think it would be valuable to re-prioritize this issue. I don’t know if Sway is still (or even ever was) the best solution, so some investigation into embedded databases is probably still needed.

simerplaha:
Hi, just checking to see if you need help validating SwayDB. 

These links might help - [benchmarks](http://swaydb.io/benchmarks/rocksdb/?language=scala/), [caching](https://dev.to/simerplaha/caching-iostrategy-swaydb-2a8e) & [blogs](https://dev.to/dashboard). Caching is very configurable depending on the use-case. It can be enabled for your entire 2GB file, parts of that file or you can have multiple smaller files with memory-mapped enabled or disabled or just for caching individual key-values on access.

As as being the best solution goes, that's always been the goal but I should point out that it at version zero (`0.13.4`) - see [project status](https://github.com/simerplaha/SwayDB#Project-status).

romac:
We won't go with SwayDB [because of its license](https://github.com/epfl-lara/stainless/pull/1071#issuecomment-848619933).

vkuncak:
Mostly subsumed by using a file with fixed-length hashes. To revisit in case we want to do some machine learning on saved VCs.

PR #1071, opened 2021-05-24
Replace file-based verification cache with SwayDB

romac:
Closes: #297

This is just a quick and dirty draft PR to replace our hand-rolled VC cache database with [SwayDB](https://github.com/simerplaha/SwayDB).

It currently uses a terrible reflection-based hack to work around `SerializableResult`'s constructor being inaccessible, and might therefore require some changes in Inox to get rid of it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -155,7 +155,7 @@ lazy val assemblySettings: Seq[Setting[_]] = {
 lazy val libFilesFile = "libfiles.txt" // file storing list of library file names
 
 lazy val regenFilesFile = false
-  
+
 lazy val libraryFiles: Seq[(String, File)] = {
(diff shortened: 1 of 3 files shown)


samarion:
Would be interesting to see some performance comparisons, _e.g._ total cache load and access time over all valid Stainless regression benchmarks under `frontends/benchmarks/verification/valid/`.

We should probably anyway have some timers debug in the `VerificationCache` as it's useful information.

romac:
As @vkuncak points out, SwayDB is licensed under a [modified version of the AGPL](https://github.com/simerplaha/SwayDB/blob/master/LICENSE.md), which is a no go due to the viral nature of the AGPL.

PR #1072, opened 2021-05-25
Imperative rewritings bug in pattern match

### LINKED ISSUE #1070: Invalid value replacement in AntiAliasing
For the following function with mutable ADTs:
```scala
  @pure
  def remove_from_list(var0: List[Int]): Unit = {
    var list: List[Int] = freshCopy(var0)
    list match {
      case Cons(first_elem, _) =>
        list = remove(list, first_elem)
        if (contains(list, first_elem)) {
          error[Nothing]("still contained")
        }
      case _ => ()
    }
  }
```
You can see that on the line of the `if` the value `first_elem` (which is immutable) is used.
However, after `AntiAliasing` the function looks like this:
```scala
@pure
def remove_from_list(var0: List[Int]): Unit = {
  var (list: List[Int]) @var = freshCopy(var0)
  (list match {
    case Cons(first_elem, _) =>
      list = remove(list, list.asInstanceOf[Cons[Int]].head)
      (if (contains(list, list.asInstanceOf[Cons[Int]].head)) {
        error[Nothing]("still contained")
      } else {
        ()
      })
    case _ =>
      ()
  })
}
```
We see that on the line of the `if` it's not `first_elem` anymore but rather `list.asInstanceOf[Cons[Int]].head`. But as `list` is newly assigned on the line before, this is not valid anymore.

Stainless later fails to prove two `cast correctness` VCs and the body assertion.

<details>
<summary>full Scala example</summary>

```scala
import stainless.lang._
import stainless.annotation._


object CastCorrectness {

  sealed abstract class List[@mutable T]
  case class Nil[@mutable T]() extends List[T]
  case class Cons[@mutable T](var head: T, var tail: List[T]) extends List[T]
  
  @pure
  def contains(self: List[Int], t: Int): Boolean =
    contents(freshCopy(self)).contains(t)

  @pure
  def contents(self: List[Int]): Set[Int] = {
    decreases(self)
    self match {
      case Nil() => Set.empty
      case Cons(head, tail) => contents(freshCopy(tail)) + head
    }
  }

  @pure
  def remove(self: List[Int], t: Int): List[Int] = {
    self match {
      case Nil() => freshCopy(Nil[Int]())
      case Cons(head, tail) if head == t => freshCopy(remove(tail, t))
      case Cons(head, tail) => freshCopy(Cons[Int](head, remove(tail, t)))
    }
  } ensuring {
    (ret: List[Int]) =>
      !contains(freshCopy(ret), t) &&
        contents(freshCopy(ret)).subsetOf(contents(freshCopy(self)))
  }

  @pure
  def remove_from_list(var0: List[Int]): Unit = {
    var list: List[Int] = freshCopy(var0)
    list match {
      case Cons(first_elem, _) =>
        list = remove(list, first_elem)
        if (contains(freshCopy(list), first_elem)) {
          error[Nothing]("still contained")
        }
      case _ => ()
    }
  }
}
````

</details>

yannbolliger:
Closes #1070.

- Add test case for rewritings bug.
- Filter out immutable ValDefs from rewritings.

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
@@ -411,7 +411,9 @@ trait AntiAliasing
             } else {
               val newScrut = transform(scrut, env)
               val newCases = cses.map { case mc @ MatchCase(pattern, guard, rhs) =>
-                val newRewritings = mapForPattern(newScrut, pattern)
+                val newRewritings =
+                  mapForPattern(newScrut, pattern).filterKeys(v => isMutableType(v.getType))
+
                 val newGuard = guard.map(transform(_, env withRewritings newRewritings))
                 val newRhs = transform(rhs, env withRewritings newRewritings)
                 MatchCase(pattern, newGuard, newRhs).copiedFrom(mc)
--- a/frontends/benchmarks/imperative/valid/CastCorrectness.scala
+++ b/frontends/benchmarks/imperative/valid/CastCorrectness.scala
@@ -1,0 +1,49 @@
+import stainless.lang._
+import stainless.annotation._
+
+
+object CastCorrectness {
+
+  sealed abstract class List[@mutable T]
+  case class Nil[@mutable T]() extends List[T]
+  case class Cons[@mutable T](var head: T, var tail: List[T]) extends List[T]
+  
+  @pure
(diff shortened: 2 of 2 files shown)


samarion (approved):

jad-hamza:
@yannbolliger is this ready to be merged?

yannbolliger:
@jad-hamza: 
> is this ready to be merged?

Yes, this can be merged.

PR #1076, opened 2021-05-26
Compute targets of a fresh LetVar

yannbolliger:
A new solution for the problem in https://github.com/epfl-lara/stainless/pull/1033/files#diff-94338a6a254a8818b64bec4aca058a1097dadc8a142418a53fa842b44024e7ecR9-R13. The previous solution was: https://github.com/epfl-lara/stainless/pull/1033/files#diff-8e88ff59f4e5533bd8fa25ed53c11010adbe901f39faa5647962bb39fdf45548R395-R39 which is not generally correct.

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
@@ -413,8 +413,7 @@ trait EffectsAnalyzer extends oo.CachingPhase {
       } yield target
 
     case fi: FunctionInvocation if !symbols.isRecursive(fi.id) =>
-      if (fi.tfd.flags.contains(IsPure)) Set.empty
-      else BodyWithSpecs(symbols.simplifyLets(fi.inlined))
+      BodyWithSpecs(symbols.simplifyLets(fi.inlined))
         .bodyOpt
         .map(getTargets(_, path))
         .getOrElse(Set.empty)
@@ -457,6 +456,8 @@ trait EffectsAnalyzer extends oo.CachingPhase {
         if (be.receiver == vd.toVariable) ee.append(be) else be
       }
 
+    case _ if isExpressionFresh(expr) => Set.empty
+
     case _ if !symbols.isMutableType(expr.getType) => Set.empty
 
     case _ =>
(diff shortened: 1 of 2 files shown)


core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -457,6 +456,8 @@ trait EffectsAnalyzer extends oo.CachingPhase {
         if (be.receiver == vd.toVariable) ee.append(be) else be
       }
 
+    case LetVar(_, e, _) if isExpressionFresh(e) => Set.empty
```
jad-hamza, line 459: I don't think that's correct (there can be targets in the body of the `LetVar`, I was thinking of:
```scala
    case _ if isExpressionFresh(expr) => Set.empty
```

Then you probably need to improve the `LetVar` case in `isExpressionFresh` for your example.

samarion, line 459: I don't think this is correct, how can you decide that there aren't any targets without looking at the body of the `LetVar`? You can return an empty set if the _body_ of the `LetVar` is fresh, but that won't help in your example program.

Is it necessary for `v` to be a `var`? If it was a `val`, then the program would probably be handled correctly by the `Let` case above.

samarion: Haha, you got there first :)

Your case with `isExpressionFresh` should work indeed. You can probably support `LetVar` inside `isExpressionFresh` by checking the var is only assigned fresh values (both in the original `LetVar` and in all subsequent `Assign` statements).

jad-hamza: > Haha, you got there first :)

I had inside information :)

We also need to make sure that the `var` is not passed into a function, where a field could be assigned to a non-fresh object. Or is that taken care of by other checks?

I wonder if we don't have the same issue when checking `isExpressionFresh` for normal `Let`'s that could be passed around to functions.

samarion: > We also need to make sure that the var is not passed into a function, where a field could be assigned to a non-fresh object. Or is that taken care of by other checks?

We check that fields are never assigned non-fresh values, so we should be fine: https://github.com/epfl-lara/stainless/blob/8fa04b5b931d788896585e51e836cf2a715884a4/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala#L107-L111

yannbolliger: Closing this in favour of the other discussion

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -470,6 +471,89 @@ trait EffectsAnalyzer extends oo.CachingPhase {
...
+      case FreshCopy(_) => true
+
+      // For `Let`, it is safe to add `vd` as a fresh binding because we disallow
+      // `FieldAssignments` with non-fresh expressions in `EffetsChecker.check(fd: FunAbstraction)`.
+      // See discussion on: https://github.com/epfl-lara/stainless/pull/985#discussion_r614583479
+      case Let(vd, e, b) => rec(e, bindings) && rec(b, bindings + vd)
+
+      // A `LetVar` can be fresh if `vd` is only assigned fresh values both
+      // here and in all subsequent Assign statements.
+      case LetVar(vd, expr, b) if isExpressionFresh(expr) && allAssignmentsFresh(vd, b) =>
```
samarion, line 537: Instead of `isExpressionFresh(expr)` here, you should rather call `rec(expr, bindings)`. This will allow `expr` to refer to let-bindings which we know are fresh.

The `allAssignmentsFresh` function will miss some assignments (_e.g._ if they appear in blocks nested under other expressions). You can use the `exprOps.exists` utility for this:
```scala
!exprOps.exists {
  case Assignment(v, e) if v.toVal == vd => !rec(e, bindings + vd)
  case _ => false
}(b)
```

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -470,6 +471,80 @@ trait EffectsAnalyzer extends oo.CachingPhase {
...
+
+      // For `Let`, it is safe to add `vd` as a fresh binding because we disallow
+      // `FieldAssignments` with non-fresh expressions in `EffetsChecker.check(fd: FunAbstraction)`.
+      // See discussion on: https://github.com/epfl-lara/stainless/pull/985#discussion_r614583479
+      case Let(vd, e, b) => rec(e, bindings) && rec(b, bindings + vd)
+
+      // A `LetVar` can be fresh if `vd` is only assigned fresh values both
+      // here and in all subsequent Assign statements.
+      case LetVar(vd, expr, b) if rec(expr, bindings) =>
+        !exprOps.exists {
```
samarion, line 525: Small nit: you should probably move the `!exprOps.exists { ... }(b)` condition into the match guard because otherwise you might miss the case where `rec(b, bindings)` is true but `vd` is assigned somewhere in `b`.

samarion (approved):
Could you maybe add a small test as well? Other than that, LGTM. Thanks for the change!

yannbolliger:
> Could you maybe add a small test as well? Other than that, LGTM. Thanks for the change!

The test is already present [here](https://github.com/epfl-lara/stainless/blob/master/frontends/benchmarks/imperative/valid/FreshCopy3.scala).

samarion:
> The test is already present here.

Oh right, I forgot you were also changing the `@pure` function behavior. Perfect!

jad-hamza:
Thank you both!

PR #1078, opened 2021-05-26
Bump Inox + better errors in watch mode + bug fix in ReturnElimination

jad-hamza:


jad-hamza:
@samarion @yannbolliger Is it OK to use latest Inox in current Stainless after your Inox merge?

samarion:
It will break caches, but should be fine otherwise (I think).

jad-hamza:
Cool, thanks!

PR #1079, opened 2021-05-26
Add weak invariants for while loops

jad-hamza:
This would be for invariants that hold except possibly after a return, as in the example:
https://github.com/epfl-lara/stainless/compare/master...jad-hamza:weak-invariant?expand=1#diff-55629b8363aa945eac95e724c9e0c143d6a3ec9acb36d5183ed377ad612d2203

Any thoughts?

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala
@@ -209,10 +209,12 @@ trait EffectsChecker { self: EffectsAnalyzer =>
         if (predEffects.nonEmpty)
           throw ImperativeEliminationException(pred, "Quantifier has effects on: " + predEffects.head.receiver.asString)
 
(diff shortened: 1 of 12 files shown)


samarion:
Couldn't you write the loop in the example as:
```scala
(while (true) {
  decreases(x)
  if (x == 0) return
  x = stainless.math.max(x / 2 - 100 / x, 0)
}).invariant(0 <= x && x <= 1000)
```

More generally, I guess the main use-case is for termination since you'll probably want the invariant to hold in order to prove post-conditions?

jad-hamza:
Yes rewriting works good too! My use-case (which can easily be rewritten too) was a bit like here, with function calls inside the body of the while, that require the mutable variables to have some property, but the property didn't hold after the return (so I have both an `invariant` which always holds + that property as a `weakInvariant`).

samarion:
Well, seems reasonable to me if you think it's useful.

PR #1084, opened 2021-05-29
Print evaluator type and values of free variable of assertion failure predicate

vkuncak:


core/src/main/scala/stainless/evaluators/Evaluator.scala
```diff
@@ -8,9 +8,15 @@ import inox.evaluators.DeterministicEvaluator
 object optCodeGen extends inox.FlagOptionDef("codegen", false)
 
 object Evaluator {
+  val RecursiveEvaluatorName: String = "default recursive evaluator"
+  val CodeGenEvaluatorName: String = "codegen evaluator"
+  var kind: String = RecursiveEvaluatorName
```
jad-hamza, line 13: Could we include the names into the evaluators instead of using the global variable `kind`? I'm afraid this requires changing Inox though to add a name field.

core/src/main/scala/stainless/evaluators/RecursiveEvaluator.scala
```diff
@@ -24,9 +27,20 @@ trait RecursiveEvaluator extends inox.evaluators.RecursiveEvaluator {
...
 
     case Assert(pred, err, body) =>
-      if (!ignoreContracts && e(pred) != BooleanLiteral(true))
-        throw RuntimeError(err.getOrElse("Assertion failed @" + expr.getPos))
-      e(body)
+      if (!ignoreContracts && e(pred) != BooleanLiteral(true)) {
+        val msg = (err match {
+          case Some(m) => m.toString + ": "
+          case None => ""
+        }) + pred.toString
```
jad-hamza: We generally use `asString` instead of `toString`, so that printing respects options such as `--print-ids`.

core/src/main/scala/stainless/evaluators/RecursiveEvaluator.scala
```diff
@@ -24,9 +27,20 @@ trait RecursiveEvaluator extends inox.evaluators.RecursiveEvaluator {
...
+        reporter.info(s"${Position.smartPos(pred.getPos)} assertion failure of ${msg}")
+        reporter.info("Relevant variables at assertion failure point:")
+        val m = rctx.mappings
+        val fvs = exprOps.variablesOf(pred)
+        val fvDump: String = fvs.map(v => v.id.toString + " -> " + m.get(v.toVal).getOrElse("?").toString)
```
jad-hamza: `toString` here too

jad-hamza (approved):
Nice! I was missing that (similar) feature from leon-web :)

core/src/main/scala/stainless/evaluators/Evaluator.scala
vkuncak: Yes, I did not want to chage Inox this time. Maybe later.

core/src/main/scala/stainless/evaluators/RecursiveEvaluator.scala
vkuncak: Replaced toString with asString in two places.

core/src/main/scala/stainless/evaluators/Evaluator.scala
samarion: Instead of storing this in a global var, you could also compute the evaluator name to output inside the `EvaluatorComponent` based on `ctx.options.findOptionOrDefault(optCodeGen)` (especially since the property isn't really global).

vkuncak: I don't like checking the option since it's not telling me what the class it, but what it should have been set to. May I use reflection and print dynamic class name?

Another question: how do I make `--check-models` benefit from this?

samarion: Reflection should work, I think there's something like `getClass.getSimpleName` which isn't too mangled.

Model checking currently takes place in Inox. The easiest way to get it to benefit from the new behavior would be to always disable `optCheckModels` in Inox (like [here](https://github.com/epfl-lara/stainless/blob/d7a4d98a3f5857eeea87515332a14122c0774344/core/src/main/scala/stainless/verification/VerificationChecker.scala#L67)) and check models inside Stainless. The logic on the Inox side is [here](https://github.com/epfl-lara/inox/blob/e477160170d83e6fa011dd3b413dab94a4918a1d/src/main/scala/inox/solvers/unrolling/UnrollingSolver.scala#L275). Model checking in Stainless could actually be more powerful because we have access to the type checker, but you might need to be careful about running into loops.

Issue #1075, opened 2021-05-25
Undetected aliasing in constructor application


yannbolliger:
The following example fails, because the aliasing that happens when `Mut(a)` is created is not recognised.

```scala
import stainless.annotation._

object Test {
  final case class Mut[@mutable T](var t: T)
  final case class S(var s: Int)

  def f2() = {
    val a = S(1)
    val b = Mut(a)
    b.t.s = 100
    assert(a.s == 100)
  }
}
```

<details>
<summary>Resulting trees</summary>

After AntiAliasing
```scala
def f2: Unit = {
  var a: S = S(1)
  var b: Mut[S] = Mut[S](a)
  b = Mut[S](S(100))
  ()
  assert(a.s == 100)
  ()
}
```

After ImperativeCodeElimination
```scala
def f2: Unit = {
  val a: S = S(1)
  val b: Mut[S] = Mut[S](a)
  val b: Mut[S] = Mut[S](S(100))
  val tmp: Unit = ()
  val tmp: Unit = ()
  assert(a.s == 100)
  ()
}
```

</details>

jad-hamza:
Maybe that property isn't enough after all?
https://github.com/epfl-lara/stainless/blob/8fa04b5b931d788896585e51e836cf2a715884a4/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala#L347-L348

Here the targets of `b` are empty and then we enter the rest of the code with the rewriting `b -> b` (is that normal?).

samarion:
This check seems too weak: https://github.com/epfl-lara/stainless/blob/8fa04b5b931d788896585e51e836cf2a715884a4/core/src/main/scala/stainless/extraction/imperative/EffectsChecker.scala#L80-L81

I think for this check, we would need `getTargets` to throw an exception in the `ADT` (and other) cases like it did before. Maybe we can change `getTargets` to have a "precise" mode where it throws an exception instead of returning an empty set when it can't compute targets.

samarion:
Yes, it should be rejected, but not exactly because of the `Mut(a)` construction. @jad-hamza added some support for aliasing when the aliased variable is never used afterwards. So if we didn't have the `assert(a.s == 100)` statement in the program, it should be transformed correctly.

samarion:
[...]

PR #1088, opened 2021-06-02
Add strict version for getTargets

jad-hamza:
Fix https://github.com/epfl-lara/stainless/issues/1075


samarion:
It might be useful for readability to introduce aliases for `getTargets(_, true)` and `getTargets(_, false)`, it's not easy to follow what's going on if we have to recall what the boolean stands for each time.

I see you made some small changes to the tests, I'm not sure I understand why the `MutableTuple` and `FixedArray` changes were required?

jad-hamza:
* For `FixedArray`, we can't compute targets of `W(30, Array(10, 20, 30, 20, 42), { w.a(0) += 1; 100 })` anymore
* For `MutableTable`, there were two issues: we can't compute targets of `(pair._1, pair._2)` anymore, and `t1` has some effects which are not allowed in assertions. I think the effects the call to `t1(...)` are not detected, I tried a fix here: https://github.com/epfl-lara/stainless/pull/1088/files#diff-8e88ff59f4e5533bd8fa25ed53c11010adbe901f39faa5647962bb39fdf45548R664
* I added an `ArrayOfPairs.scala` example, which fails at the moment with the error: "Function `insert` has effect on @pure parameter `time`"
* I added a similar example (`ArrayOfPairs2.scala`), which fails with the same error. And if we comment out this check in Stainless, we get another error: "Couldn't compute effect targets in finite array Array((time, elem))"

I'm investigating :) (I'd like to have these examples work before merging)

jad-hamza:
The error for `ArrayOfPairs.scala` comes from the fact that I added this `try catch` to overapproximate the affected variable: https://github.com/epfl-lara/stainless/pull/1088/files#diff-8e88ff59f4e5533bd8fa25ed53c11010adbe901f39faa5647962bb39fdf45548R719
Should we filter the `freeVars` here?

samarion:
`FixedArray`: why can't we compute precise targets? It should be a fresh expression, no?

`MutableTuple`: is that because of the call to `checkAliasing`? Hmm, maybe we need a third mode for `getTargets` which returns all possible targets (_i.e._ `pair._1` and `pair._2` in this case) instead of crashing. That might also fix the `ArrayOfPairs` testcase.

jad-hamza:
> `FixedArray`: why can't we compute precise targets? It should be a fresh expression, no?

You're right, it was crashing because of the order of cases in `getTargets`. I moved up the case for `isExpressionFresh`, maybe we should add caching to the functions that are invoked in the early cases (`isExpressionFresh`, `isMutableType`, `variablesOf`).

> `MutableTuple`: is that because of the call to `checkAliasing`?

No, it's in `mapApplication` for the function invocation `t1((pair._1, pair._2))`: https://github.com/epfl-lara/stainless/blob/4c10f145bd4ae29164330b09f6935c9b6d1d95e4/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L505
I've rewrote the code as follows to avoid the effect in the assertion:

```scala
    val x = t1((pair._1, pair._2))
    assert(x == 100)
```

In `mapApplication`, the error happens here: https://github.com/epfl-lara/stainless/blob/4c10f145bd4ae29164330b09f6935c9b6d1d95e4/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L184
where:
* `fiEffects` just has one effect: `Effect(pair, empty)`
* in the call `e on rArg`, `e` is that effect, and `rArg` is `(pair._1, pair._2)`

Then the call to `on` crashes here on the call to `getTargets` (link to PR):
https://github.com/jad-hamza/stainless/blob/894aa5eef5c89e5493e596b101bd62b60f4830a2/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala#L194

> Hmm, maybe we need a third mode for `getTargets` which returns all possible targets (_i.e._ `pair._1` and `pair._2` in this case) instead of crashing. That might also fix the `ArrayOfPairs` testcase.

That's a good idea, can we avoid the version that crashes altogether then?

samarion:
> maybe we should add caching to the functions that are invoked in the early cases

Possibly, but caches sometimes have a significant cost as well. If you want to add some, I'd recommend using `context.timers` to check if the cache is making a real difference. Maybe in a follow-up PR?

> fiEffects just has one effect: Effect(pair, empty)

That's weird, shouldn't `fiEffects` have the effects `Effect(pair, ._1)` and `Effect(pair, ._2)` ? That's what the `t1` function is actually mutating, no?

> That's a good idea, can we avoid the version that crashes altogether then?

Hmm, yes that might be possible.

I guess the condition for the linear let-binding would be something like `getTargets(b, true) == getTargets(b, false)` (where the boolean argument determines whether we're returning all possible targets or not) ?  
Not super intuitive, but it could be correct.

jad-hamza:
> That's weird, shouldn't `fiEffects` have the effects `Effect(pair, ._1)` and `Effect(pair, ._2)` ? That's what the `t1` function is actually mutating, no?

Yes you're right, it should be `Set(Effect(pair, ._1))`, but the effect gets truncated since the `truncate` function was not updated for tuples (I'll fix).
 
> I guess the condition for the linear let-binding would be something like `getTargets(b, true) == getTargets(b, false)` (where the boolean argument determines whether we're returning all possible targets or not) ?
> Not super intuitive, but it could be correct.

Oh no I think I misunderstood what you meant. In some cases, we are able to overapproximate the targets and return them (as in the case of `(pair._1, pair._2)`, but there will still be case that are too complicate to even overapproximate (where we'll crash)?

samarion:
> Oh no I think I misunderstood what you meant. In some cases, we are able to overapproximate the targets and return them (as in the case of (pair._1, pair._2), but there will still be case that are too complicate to even overapproximate (where we'll crash)?

I don't think it's about the case being too complicated.

We need some way to determine whether the rhs of the let-binding is a precise target in order to decide which transformation we want to apply ([here](https://github.com/epfl-lara/stainless/blob/c7383d5e81f2f7da93154d2adc975b5a105b688d/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L344)). Crashing on imprecise targets was how I thought we'd handle this, but maybe using the condition I mentioned could work as well.

I think that's the only place where we really need the notion of precise targets. The other calls to `getTarget` should be fine with using either the current non-crashing `getTarget`, or the version which returns over-approximations.

jad-hamza:
So for now I used `getDirectTargets` almost everywhere except for deciding how to transform `let`'s, and in `checkAliasing`.
By the way, is it ok to remove this part https://github.com/epfl-lara/stainless/blob/master/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L187-L197 since we already have `checkAliasing`?

At the moment everything is failing because the computation `on` in `mapApplication` (https://github.com/epfl-lara/stainless/blob/4c10f145bd4ae29164330b09f6935c9b6d1d95e4/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L184) often returns the empty set because of this case: https://github.com/jad-hamza/stainless/blob/3e3c5e89949ed9b85c033321edac570166288309/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala#L355

Should `on` call `getTargets` at all? Isn't `wrap` enough?

For instance, in https://github.com/jad-hamza/stainless/blob/3e3c5e89949ed9b85c033321edac570166288309/frontends/benchmarks/imperative/valid/LetAliasing.scala since `box1.value` is a fresh expression, the call `Effect(b.value) on box1` returns the empty set.

jad-hamza:
Another question about `getTargets`: is it normal that it returns the empty set on e.g. `x` if `x` is declared somewhere as `var x: Int = 0`. We directly return the empty set for expressions of immutable types, but we don't do exactly the same in `expressionEffects` where we also consider mutable variables of immutable types (like `x`).

samarion:
> By the way, is it ok to remove this part [...] since we already have checkAliasing?

Hmm, this makes me wonder if we didn't make a mistake in our `checkAliasing` change.

I think the old check using `varsInScope` wasn't for the same thing, but rather to ensure we never had something like:
```scala
val a = Array(1,2,3)
def f(ar: Array[Int]): Unit = {
  ar(0) = 1
  a(0) = 2
}
f(a)
```

The old check wasn't quite right either, but we still need a check like this. Something like:
```scala
val vis: Set[Variable] = varsInScope(fd)
args.flatMap(variablesOf(_)).find(vis contains _)
  .foreach(v => throw FatalError("Illegal passing of aliased local variable: " + v))
```
This is only relevant for the `ApplyLetRec` case, though.

---

Regarding the check in [mapApplication](https://github.com/epfl-lara/stainless/blob/master/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L187-L197), I believe it's indeed an (insufficient) version of `checkAliasing`. IIUC, the idea was to allow passing some variable to a same function twice as long as the effects within the function only affected disjoint parts of the mutable object. However, I don't think this actually works because the reads won't correctly reflect the changes to the aliased parts.

So yes, we should be able to remove that check. Btw, I see we don't call `checkAliasing` in the case of applications. I think we need the check there as well.

---

> At the moment everything is failing because the computation on in mapApplication [...]

That seems like a reasonable behavior to me. If we pass a fresh value to a function call, then we shouldn't need to propagate the mutation after the function completes. What are you expecting instead?

> Should on call getTargets at all? Isn't wrap enough?

Well they don't have the same API? `wrap` will return an expression and I assume we need an effect / target?

> For instance, in [..] since box1.value is a fresh expression, the call Effect(b.value) on box1 returns the empty set.

What? Why is `box1.value` a fresh expression? That doesn't sound right to me at all.

> Another question about getTargets: is it normal that it returns the empty set on [...]

Yes, that's normal. `getTargets` is only supposed to be used for handling aliasing, and there's no aliasing with `var`s of immutable types. I'm not sure why `expressionEffects` also considers assignments as effects though.

jad-hamza:
> The old check wasn't quite right either, but we still need a check like this. Something like:> 
> ```scala
> val vis: Set[Variable] = varsInScope(fd)
> args.flatMap(variablesOf(_)).find(vis contains _)
>   .foreach(v => throw FatalError("Illegal passing of aliased local variable: " + v))
> ```
> 
> This is only relevant for the `ApplyLetRec` case, though.

Thanks, I've added that in `ApplyLetRec`.

> Regarding the check in [mapApplication](https://github.com/epfl-lara/stainless/blob/master/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L187-L197), I believe it's indeed an (insufficient) version of `checkAliasing`. IIUC, the idea was to allow passing some variable to a same function twice as long as the effects within the function only affected disjoint parts of the mutable object. However, I don't think this actually works because the reads won't correctly reflect the changes to the aliased parts.
> 
> So yes, we should be able to remove that check. Btw, I see we don't call `checkAliasing` in the case of applications. I think we need the check there as well.

I've removed the old check, and added `checkAliasing` for applications.

> That seems like a reasonable behavior to me. If we pass a fresh value to a function call, then we shouldn't need to propagate the mutation after the function completes. What are you expecting instead?
> Well they don't have the same API? `wrap` will return an expression and I assume we need an effect / target?
> What? Why is `box1.value` a fresh expression?

It's because `box1.value` has type `Boolean`, and we check `isMutableType` in `isExpressionFresh`:
https://github.com/epfl-lara/stainless/blob/4c10f145bd4ae29164330b09f6935c9b6d1d95e4/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala#L487

> Yes, that's normal. `getTargets` is only supposed to be used for handling aliasing, and there's no aliasing with `var`s of immutable types. I'm not sure why `expressionEffects` also considers assignments as effects though.

I think it's useful when we check that certain expressions don't have effects (such as assertions).

jad-hamza:
I've changed the order of operations in `on`, and this fixes the `LetAliasing` example.

Now we compute:
* `getTargets(box1)` where we then append the path `_.value`, while before we were doing:
* `getTargets(box1.value)`, which is empty in this PR.

jad-hamza:
This new way of computing `on` fails for more complicated example such as:
https://github.com/epfl-lara/stainless/blob/master/frontends/benchmarks/imperative/valid/ObjectHierarchyMutation10.scala

At some point we need to compute `Effect(b.a1.x) on B(a1, a2)` which returns the empty set (because we use `getDirectTargets` on `B(a1, a2)`), or crashes if we use `getAllTargets`.

I've changed `eff.on(expr)` to be `getDirectTargets(on, eff.path)`, which works better for this example.

samarion:
Looks good to me, thanks for all the changes!

Good job finding out we can replace the calls to `getTargets(wrap(_))` with `getTargets` directly.

It could be useful to add some documentation to `getDirectTargets` and `getAllTargets`, especially the later where the name doesn't exactly explain what happens (_i.e._ the fact that it crashes on `Mut(a)` instead of returning `a` could be surprising).

jad-hamza:
> Looks good to me, thanks for all the changes!
> Good job finding out we can replace the calls to `getTargets(wrap(_))` with `getTargets` directly.

Thanks for the help :) I'm still not sure it's correct, because the issue #1090 is still present.
I've adapted this example from the issue:
```scala
import stainless.annotation._

object MutateInside {
  case class Mut[@mutable T](var t: T)
  case class Thing(var field: Int)

  def resetThing(m: Mut[Thing]): Unit = {
    m.t = Thing(0)
  }

  def main(): Unit = {
    val thing = Thing(123)
    resetThing(Mut(thing))
    assert(thing.field == 123)
  }
}
```
where the assertion should be valid but is rejected.

I'll still merge this PR, and I'll try to fix #1090 in another PR.

> It could be useful to add some documentation to `getDirectTargets` and `getAllTargets`, especially the later where the name doesn't exactly explain what happens (_i.e._ the fact that it crashes on `Mut(a)` instead of returning `a` could be surprising).

Yes you're right, I've added more comments.

Issue #1090, opened 2021-06-07
Inconsistent change propagation in `AntiAliasing`


yannbolliger:
The following test case shows of some strange behaviour with change/mutation propagation in the AntiAliasing/ImperativeCodeElimination phases.

```scala
import stainless.lang._
import stainless.annotation._

object Mutable {
  final case class Mut[@mutable T](var t: T)
  final case class Thing[@mutable T](var field: T)

  def change_thing[@mutable T](thing: Mut[Thing[T]], t: T) = {
    thing.t = Thing(freshCopy(t))
  }
  def main() = {
    val thing = Thing(123)
    change_thing(Mut(thing), 456)
    assert(thing.field == 456)

    val thing2 = Mut(thing)
    change_thing(thing2, 789)
    assert(thing.field == 789)

    val thing3 = Mut(thing)
    thing3.t = Thing(0)
    assert(thing.field == 0)
  }
}
```
For the first block, the changes made to the wrapped `thing` are correctly propagated back, whereas for the next ~two~ block~s~, the changes to the `Mut(thing)` are only propagated back to the bound `thing2`/~`thing3`~.

**EDIT**: with the latest `master` version (a5292b4a3a55a3ba85c18aaa55db2f915b95724f), the changes are also correctly propagated for the third block.

<details>
<summary> stainless trees in some phases</summary>

```
[info] [ Debug  ] Symbols before Preprocessing
[info] [ Debug  ] 
[info] [ Debug  ] -------------Functions-------------
[info] [ Debug  ] def change_thing[T @mutable](thing: Mut[Thing[T @mutable]], t: T @mutable): Unit = thing.t_=(Thing[T @mutable](freshCopy(t)))
[info] [ Debug  ] 
[info] [ Debug  ] @accessor(t)
[info] [ Debug  ] @method(Mut)
[info] [ Debug  ] def t: T @mutable = this.t
[info] [ Debug  ] 
[info] [ Debug  ] def main: Unit = {
[info] [ Debug  ]   val thing: Thing[Int] = Thing[Int](123)
[info] [ Debug  ]   change_thing[Int](Mut[Thing[Int]](thing), 456)
[info] [ Debug  ]   assert(thing.field == 456)
[info] [ Debug  ]   val thing2: Mut[Thing[Int]] = Mut[Thing[Int]](thing)
[info] [ Debug  ]   change_thing[Int](thing2, 789)
[info] [ Debug  ]   assert(thing.field == 789)
[...]

PR #1098, opened 2021-06-10
Split effects into replacement, modifying, and combined effects

jad-hamza:
Fix https://github.com/epfl-lara/stainless/issues/1090 (work in progress, see discussion)


jad-hamza:
This is currently failing (among other places) in Tutorial.scala, here is a simplified version (assertion fails while it should go through):

```scala
import stainless.lang._

object Banking {

  case class BankAccount(var checking: BigInt)

  case class Transaction(
    account: BankAccount,
    var executed: Boolean
  )

  def execute(tr: Transaction): Unit = {
    tr.executed = true
    incr(tr.account)
  }

  def incr(acc: BankAccount): Unit = {
    acc.checking += 1
  }

  def incr2(from: BankAccount, to: BankAccount): Unit = {
    require(from.checking == 0)
    execute(Transaction(from, false))
    assert(from.checking == 1)
  }
}
 ```


Some debug output from `mapApplication`: on master, where the assertion correctly goes through:
```scala
----------------------------------------------------------------------------------------------------
(nfi,incr(tr.account))
(fiEffects,Set(Effect(acc$3.checking$0)))
(localEffects,List(Set((Effect(acc$3.checking$0),Set(Effect(tr$1.account$0.checking$0))))))
----------------------------------------------------------------------------------------------------
(nfi,execute(Transaction(from, false)))
(fiEffects,Set(Effect(tr$0.executed$0), Effect(tr$0.account$0.checking$0)))
(localEffects,List(Set((Effect(tr$0.executed$0),Set()), (Effect(tr$0.account$0.checking$0),Set(Effect(from$3.checking$0))))))
```

And on the PR, where the assertion fails:
```scala
----------------------------------------------------------------------------------------------------
(nfi,incr(tr.account))
(fiEffects,Set(Effect(acc$3<empty>)))
(localEffects,List(Set((Effect(acc$3<empty>),Set(Effect(tr$1.account$0))))))
----------------------------------------------------------------------------------------------------
(nfi,execute(Transaction(from, false)))
(fiEffects,Set(Effect(tr$0<empty>)))
(localEffects,List(Set((Effect(tr$0<empty>),Set()))))
```

Because of the assignment `tr.execute = true` (the assertion goes through when we remove that assignment), the effect of `execute` becomes just `Effect(tr, empty)`, and the effect `Effect(tr, _.account)` probably gets eaten when we merge the effects in https://github.com/epfl-lara/stainless/blob/dff1b643830fdbad1ee5ff8d1cc8ad7f7cc15e05/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala#L720

Then, applying the effect `Effect(tr, empty)` `on` the argument `Transaction(...)` (with `getDirectTargets`) returns empty targets.
Maybe that's where the third mode of `getTargets` that you suggested would be useful? Or we could avoid merging effects here?

samarion:
Some general thoughts:
- The `on` method needs to be implemented using `getAllTargets` in order to correctly handle imprecise effects.
- Dropping the last path element prevents us from distinguishing precise and imprecise effects. We can still implement something correct, but we'll lose in coverage (_e.g._ the `Banking` example can't be supported because we can't determine whether the `tr` effect is a precise effect which can only modify the `executed` variable, or an imprecise effect which could mutate things inside `account` as well).

Maybe we do need some way to distinguish between precise and imprecise effects in order to implement `on` correctly?

jad-hamza:
Why do we need imprecise effects? Is it because of the truncation and merges? And what exactly do we mean by precise/imprecise?
* an expression has precise effect `Effect(x, path)` if it does `x.path = ...` (it replaces `x.path`)
* an expression has imprecise effect `Effect(x, path)` if it does modifications of the form `x.path.path2 = ...`

If this is the definition, we could call them "replacing effects" (precise) and "modifying effects" (imprecise)?

Can an expression have both a precise and imprecise effect with the same variable/path combination?

jad-hamza:
I've played a bit with replacement/modifying effects in the context of that example:
```scala
import stainless.annotation._

object MutateInside2 {
  case class Mut[@mutable T](var t: T)
  case class Thing(var field: Int)

  def resetThing(m: Mut[Thing]): Unit = {
    m.t = Thing(0)
  }

  def main(): Unit = {
    val thing = Thing(123)
    resetThing(Mut(thing))
    assert(thing.field == 123)
  }
}
```

In that example we'll compute that `resetThing` has a replacement effect on `m.t`, and so when we apply the effect to `Mut(thing)` (for the function invocation `resetThing(Mut(thing))`, we'll know that the effect doesn't modify `thing`.

But how would that work for these variants?
```scala
  def resetThing2(m: Mut[Thing]): Unit = {
    m.t = Thing(0)
    m.t.field = 100
  }
```
and
```scala
  def resetThing3(m: Mut[Thing]): Unit = {
    m.t.field = 100
    m.t = Thing(0)
  }
```

Will the effects be in both cases `ReplacementEffect(m, _.t.field)` and `ReplacementEffect(m, _.t)`?
The call to `resetThing2` will not modify `thing`, while the call to `resetThing3` will set `thing.field` to `100`.

samarion:
> But how would that work for these variants?

Probably we'd have to let the `merge` logic collapse both effects into a modifying effect on `m.t`. Then, `getTargets` should fail when trying to pass in a `Mut(thing)` argument.

jad-hamza:
It still feels strange that the effects computations succeed and return the same thing for these three functions? Which ones do you think should fail and which one should succeed?

Also, why would a replacement or modifying effect for `m.t` fail when applied on `Mut(thing)`?
It's well defined what happens on `thing` for both cases, no?
(a replacement effect does nothing, while a modifying effect propagates the modification to `thing`)

> into a modifying effect on

Isn't it replacement? (meaning we replace the whole thing) Sorry the naming isn't great here.
I'm not sure they should be collapsed, because a replacement effect on `m.t` doesn't imply modifying effect on `m.t` nor vice versa (maybe replacement effect on `m.t` implies modifying effect on `m`).

Perhaps if we see (during `merge` for instance), that there is a replacement effect which is a prefix (or equal) to a modifying effect, we fail, because in that case we cannot tell what happens to the arguments of the functions just from looking at the effects and the result of the function.

samarion:
> It still feels strange that the effects computations succeed and return the same thing for these three functions? Which ones do you think should fail and which one should succeed?

I think they should all succeed, effect computation doesn't care about effect order. It's rather at the function call site that success/failure should be determined. For example, the calls `resetThing(mut)`, `resetThing2(mut)` and `resetThing3(mut)` will all be handled correctly by the transformation.

> Also, why would a replacement or modifying effect for m.t fail when applied on Mut(thing)?

Well, the issue is having both a replacement _and_ modifying effect on `m.t`. I was thinking we would slightly abuse the modifying effect and let it subsume replacement (so it would be easy to fail here), but maybe it would make sense to really keep them separate? Or add a third kind of effect which corresponds to the merge of both of them?

> I'm not sure they should be collapsed, because a replacement effect on m.t doesn't imply modifying effect on m.t nor vice versa (maybe replacement effect on m.t implies modifying effect on m).

Collapsing is useful to allow checking call sites easily. If you have a collapsed `m.t` (which had both replacement and modifying), then you should fail if the provided argument is `Mut(thing)`.

> Perhaps if we see (during merge for instance), that there is a replacement effect which is a prefix (or equal) to a modifying effect, we fail, because in that case we cannot tell what happens to the arguments of the functions just from looking at the effects and the result of the function.

See my first point, the issue can only be discovered at call site, not while computing the function effects. If we provide a `mut` variable as argument, then the order of replacements/modifications doesn't matter because it will be correctly captured in the function's body and copied back out into the `mut` variable.

jad-hamza:
You're right this makes sense, we compute some effects, and then we fail when we cannot apply them. I added a third effect called `CombinedEffect`, and I put the failing logic in `on`:
https://github.com/epfl-lara/stainless/pull/1098/files#diff-8e88ff59f4e5533bd8fa25ed53c11010adbe901f39faa5647962bb39fdf45548R362-R371

Here's the combining logic:
https://github.com/epfl-lara/stainless/pull/1098/files#diff-8e88ff59f4e5533bd8fa25ed53c11010adbe901f39faa5647962bb39fdf45548R799-R813

jad-hamza:
It's mostly working now, except for the array of pairs examples
https://github.com/epfl-lara/stainless/blob/acb76f650ec68bb279fe2bccf27aa5a932a582c1/frontends/benchmarks/imperative/valid/ArrayOfPairs2.scala
https://github.com/epfl-lara/stainless/blob/acb76f650ec68bb279fe2bccf27aa5a932a582c1/frontends/benchmarks/imperative/valid/ArrayOfPairs3.scala

Here we try to apply a modifying effect `elemRef` (with empty path because array effects are truncated) on `Array((time, elem))`. Should the targets just be empty in that case, because a modifying effect on `elemRef` cannot also modify `elem` (otherwise it'd be a combined effect)?
We could enter `getTargets` with `strict = false` when we have a modyfing effect?

samarion:
Maybe we can improve the truncation logic? Something like truncating a replacement effect on the path `Seq(ArraySelector(_))` becomes a replacement effect on an empty path, and that has no targets when applied to a `FiniteArray`? I'm not sure...

One other idea I had would be to automatically let-bind all non-variable mutable arguments before passing them to functions. That would be semantically equivalent and allow the aliasing transformation to succeed more often. WDYT?

samarion:
> One other idea I had would be to automatically let-bind all non-variable mutable arguments before passing them to functions. That would be semantically equivalent and allow the aliasing transformation to succeed more often. WDYT?

Oh wait, that wouldn't work when we're wrapping (potentially) aliased variables. You can forget this suggestion.

jad-hamza:
> Maybe we can improve the truncation logic? Something like truncating a replacement effect on the path `Seq(ArraySelector(_))` becomes a replacement effect on an empty path, and that has no targets when applied to a `FiniteArray`? I'm not sure...

Can you explain more, why is the truncation of a replacement effect still a replacement effect? Replacing the array should be different than modifying something inside?

And what happens if there are no targets when we're modifying the field of a mutable object inside an array?

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -191,9 +191,11 @@ trait AntiAliasing
             val extractResults = Block(
               for {
                 (effects, index) <- localEffects.zipWithIndex
-                (outerEffect, innerEffects) <- effects
-                effect <- innerEffects
+                (outerEffect0, innerEffects) <- effects
+                effect0 <- innerEffects
               } yield {
+                val outerEffect = outerEffect0.removeLastUnknownAccessor
+                val effect = effect0.removeLastUnknownAccessor
```
samarion, line 198: It feels a bit strange to drop the `UnknownAccessor` here. What's the rationale?

Could we maybe make sure it's handled correctly during effect computation, or maybe in `updatedTarget`?

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -352,14 +354,15 @@ trait AntiAliasing
...
-                  // and we update it
-                  val result = eff.wrap.getOrElse(
-                    throw MalformedStainlessCode(l, "Unsupported `val` in AntiAliasing (couldn't compute update after aliasing)")
+              val copyEffects = effects(b).filter(_.receiver == vd.toVariable).flatMap { eff =>
+                // we apply the effect on the bound expression (after transformation)
+                eff.on(newExpr).map { eff2 =>
+                  Assignment(eff2.receiver,
+                    updatedTarget(
+                      Target(eff2.receiver, None, eff2.path),
+                      eff.wrap.get
```
samarion, line 363: Nit: why are you changing the type of exception thrown here?

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -324,12 +334,56 @@ trait EffectsAnalyzer extends oo.CachingPhase {
...
+  sealed abstract class Effect(val kind: EffectKind) {
+    val receiver: Variable
+    val path: Path
+
+    def +(elem: Accessor): Effect = Effect(kind, receiver, path :+ elem)
+    def ++(path2: Path): Effect = Effect(kind, receiver, path ++ path2)
+    def precise(elem: Accessor) = ReplacementEffect(receiver, path :+ elem)
+
+    def changePath(newPath: Path): Effect = Effect(kind, receiver, newPath)
+    def changeKind(newKind: EffectKind): Effect = Effect(newKind, receiver, path)
```
samarion, line 365: Nit: withPath and withKind ?

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -710,41 +770,43 @@ trait EffectsAnalyzer extends oo.CachingPhase {
           val tpe = tup.bases(idx - 1)
           if (isInductive(tpe, seen)) Seq()
           else fa +: rec(tpe, xs, seen)
-        case (_, ArrayAccessor(_) +: xs) => Seq()
+        case (ArrayType(base), (aa @ ArrayAccessor(i)) +: xs) if exprOps.variablesOf(i).isEmpty =>
+          if (isInductive(base, seen)) Seq()
+          else aa +: rec(base, xs, seen)
```
samarion, line 775: This seems a bit dangerous, it could be that `i` has an effect.

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -710,41 +770,43 @@ trait EffectsAnalyzer extends oo.CachingPhase {
...
+        case (ArrayType(base), (aa @ ArrayAccessor(i)) +: xs) =>
+          if (isInductive(base, seen)) Seq()
+          else UnknownArrayAccessor +: rec(base, xs, seen)
```
samarion, line 778: What happens when we have an `UnknownArrayAccessor` within the path, do we just crash when applying the effect?

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -710,41 +770,43 @@ trait EffectsAnalyzer extends oo.CachingPhase {
...
+          freeVars.map(v => ModifyingEffect(v, Path.empty)).toSet
       }
 
-    mutated
-      .map(truncate)
-      .groupBy(_.receiver)
-      .flatMap { case (v, effects) => merge(effects.map(_.path)).map(Effect(v, _)) }.toSet
+    val truncatedEffects = mutated.map(truncate)
+
+    val combinedEffects = truncatedEffects.flatMap { e =>
```
samarion, line 798: Could be worth adding a comment above explaining all the combinations.

samarion (commented):
Looking good! I think we're getting somewhere :)

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
jad-hamza: It's approximating the effect (I could make it more general to remove suffixes). I did this because `select` was failing when applied to an `outerEffect` which contained an `UnknownArrayAccessor`. By the way can we replace `select` by simply `wrap` (which doesn't generate side conditions).

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
jad-hamza: Did you mean in `on`?

samarion: If `eff.wrap` was `None`, we used to throw a  `MalformedStainlessCode` exception, but now it will be a `NoSuchElementException`.

jad-hamza:
> Looking good! I think we're getting somewhere :)

Yes :) thanks for the ideas and reviews!

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -361,22 +427,23 @@ trait EffectsAnalyzer extends oo.CachingPhase {
    * modify `a`. So `getTargets` crashes in that case, because our `Target` API cannot capture such
    * correspondence.
    */
-  def getTargets(expr: Expr, path: Seq[Accessor], strict: Boolean)(implicit symbols: Symbols): Set[Target] = expr match {
+  def getTargets(expr: Expr, path: Seq[Accessor], strict: Boolean, kind: EffectKind)(implicit symbols: Symbols): Set[Target] = expr match {
```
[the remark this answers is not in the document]
jad-hamza: I tried to remove the `strict` argument, but then some things fail with `getDirectTargets`, where we want to have non-strict mode, but at the same time we still want to have a target for a variable `x` and an empty path, which are discarded when using `ReplacementKind`.

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
jad-hamza: I added a case here

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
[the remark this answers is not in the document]
samarion: Ah, I see.

I guess we could change the API of `getDirectTargets` a little bit and do something like:
```scala
def getDirectTargets(expr: Expr, accessor: Accessor)(implicit symbols: Symbols) =
  getTargets(Path(Seq(accessor)).wrap(expr), ReplacementKind)
```
Not sure there's much point though. It might make `getTargets` a little simpler and easier to understand, but it's as you prefer.

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -439,16 +512,18 @@ trait EffectsAnalyzer extends oo.CachingPhase {
...
         .getOrElse(Set.empty)
 
     case fi: FunctionInvocation => Set.empty
     case (_: ApplyLetRec | _: Application) => Set.empty
-    case _: LargeArray | _: ArrayUpdated
-      if (path.isEmpty || !path.head.isInstanceOf[ArrayAccessor]) && !strict => Set.empty
+    case _: LargeArray | _: ArrayUpdated if kind == ReplacementKind && path.isEmpty =>
+      Set.empty
+    case _: LargeArray | _: ArrayUpdated if kind == ReplacementKind && !path.head.isInstanceOf[ArrayAccessor] && path.head != UnknownArrayAccessor =>
+      Set.empty
```
samarion, line 523: Nit: you could merge these two into
```scala
case _: LargeArray | _: ArrayUpdated if kind == ReplacementKind && path.headOption.forall {
  case _: ArrayAccessor | UnknownArrayAccessor => false
  case _ => true
} => Set.empty
```

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
samarion: > It's approximating the effect (I could make it more general to remove suffixes). I did this because select was failing when applied to an outerEffect which contained an UnknownArrayAccessor. 

Could we do the approximation when computing the function effects? E.g. at truncation time?

> By the way can we replace select by simply wrap (which doesn't generate side conditions).

Yes, you could probably split `select` into `wrap` and some logic which generates the path condition. I think you need to keep the path condition to handle the case where your potential effect applies to some ADT constructor which _isn't_ actually in the mutated value. E.g. you have an effect on `MutableSome.v` but actually passed in a `MutableNone`.

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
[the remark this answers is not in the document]
jad-hamza: Good idea! It's nicer :)

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
jad-hamza: > Could we do the approximation when computing the function effects? E.g. at truncation time?

In one of the examples `ArrayOfPairsX.scala`, we needed to know that the function has a `ReplacementEffect(array, Seq(UnknownAccessor))` in order to know that it has no targets on a finite array. If we approximate before to `ModifyingEffect(array)`, we wouldn't be able to compute the targets on the finite array.

> E.g. you have an effect on `MutableSome.v` but actually passed in a `MutableNone`.

But then is it ok to use `wrap` here instead of `select`? https://github.com/epfl-lara/stainless/blob/10a09e621a0264608dd362fa2f960c170af01742/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L359

I wonder if we couldn't remove that (effect postponing) branch altogether:
https://github.com/epfl-lara/stainless/blob/10a09e621a0264608dd362fa2f960c170af01742/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L351
now that we have changed the `getDirectTargets` API to take into account the accessor?
So we wouldn't check that `getTargets` when doing a let binding, because `getDirectTargets` might succeed later when given the right accessor.

That would also solve this: https://github.com/epfl-lara/stainless/issues/1099

jad-hamza: Nevermind, I see we also put the "bare" targets of the let-bound expression in the rewritings, we couldn't do that if we don't have all the targets here.

samarion: > In one of the examples ArrayOfPairsX.scala, [...]

Okay, let's keep the call to `removeUnknownAccessor`. Thanks for trying out the other options!

> But then is it ok to use wrap here instead of select?

Good question. I wonder if we shouldn't be using something like `mapApplication` there. The logic should be the same as for function calls, no?

jad-hamza: Yes, I've updated that with `select` and it worked without further modifications!

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -489,15 +554,9 @@ trait EffectsAnalyzer extends oo.CachingPhase {
...
-  def getTargets(expr: Expr, strict: Boolean)(implicit symbols: Symbols): Set[Target] = {
-    getTargets(expr, Seq.empty, strict)
-  }
-
-  // See the description above `getTargets` for more information
-  def getDirectTargets(expr: Expr, path: Seq[Accessor])(implicit symbols: Symbols) = getTargets(expr, path, false)
-  def getAllTargets(expr: Expr, path: Seq[Accessor])(implicit symbols: Symbols) = getTargets(expr, path, true)
-  def getDirectTargets(expr: Expr)(implicit symbols: Symbols) = getTargets(expr, Seq.empty, false)
-  def getAllTargets(expr: Expr)(implicit symbols: Symbols) = getTargets(expr, Seq.empty, true)
+  def getAllTargets(expr: Expr, kind: EffectKind, path: Seq[Accessor] = Seq.empty)(implicit symbols: Symbols) = getTargets(expr, kind, path)
```
samarion, line 557: Nit: I think you can make `getAllTargets` always use `ModifyingKind` now.

If I'm not mistaken, there's only one place where you call `getAllTargets` with a (possibly) different kind, and you should be able to replace that call with `getTargets` directly.

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -710,41 +767,58 @@ trait EffectsAnalyzer extends oo.CachingPhase {
           val tpe = tup.bases(idx - 1)
           if (isInductive(tpe, seen)) Seq()
           else fa +: rec(tpe, xs, seen)
-        case (_, ArrayAccessor(_) +: xs) => Seq()
+        case (ArrayType(base), (aa @ ArrayAccessor(i)) +: xs) if exprOps.variablesOf(i).isEmpty && expressionEffects(i, result).isEmpty =>
+          if (isInductive(base, seen)) Seq()
+          else aa +: rec(base, xs, seen)
+        case (ArrayType(base), ArrayAccessor(i) +: xs) =>
+          if (isInductive(base, seen)) Seq()
+          else UnknownArrayAccessor +: rec(base, xs, seen)
```
samarion, line 775: Nit: mostly personal preference, but I think it would be more readable if you merged the two cases:
```scala
case (ArrayType(base), (aa @ ArrayAccessor(i)) +: xs) =>
  if (isInductive(base, seen)) Seq()
  else {
    val accessor = if (exprOps.variablesOf(i).isEmpty && expressionEffects(i, result).isEmpty) aa
                   else UnknownArrayAccessor
    accessor +: rec(base, xs, seen)
  }
```

Also, is it necessary to truncate the path when `base` is inductive? Isn't it fine to keep the array accessor and simply let the `rec` case be truncated?

jad-hamza: Good point, fixed! (also for tuples)

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
jad-hamza: Nice catch, fixed!

jad-hamza:
Merging now, thanks a lot for the help! I'll open another PR if there are things to fix

Issue #1117, opened 2021-07-13
GenC: Translate @pure annotation to C


deiruch:
When using `@pure` to annotate a pure function, this should be reflected in the generated C code. Adding the annotation triggers additional optimizations in GCC, which are beneficial for speed & executable size. Here's an example of GenC output for two functions, one of which is annotated with `@pure`.

The `stainless.h` header file should include the following preprocessor macro:
```C
/* --------------------------- preprocessor macros ----- */

#define STAINLESS_FUNC_PURE
#if defined(__cplusplus)
#undef STAINLESS_FUNC_PURE
#define STAINLESS_FUNC_PURE _Pragma("FUNC_IS_PURE;")
#elif __GNUC__>=3
#undef STAINLESS_FUNC_PURE
#define STAINLESS_FUNC_PURE __attribute__((__pure__))
#elif defined(__has_attribute)
#if __has_attribute(pure)
#undef STAINLESS_FUNC_PURE
#define STAINLESS_FUNC_PURE __attribute__((__pure__))
#endif
#endif
```

The function declarations in `stainless.c` and `stainless.h` should look like this (note the `STAINLESS_FUNC_PURE`):
```C
/* ---------------------- function declarations ----- */

uint32_t aNonPureFunction_0(void);
STAINLESS_FUNC_PURE uint32_t thisFunctionWasAnnotatedWithPure_0(void);
```

The actual implementation in `stainless.c` should also include the annotation, as shown here:
```C
/* ----------------------- function definitions ----- */

uint32_t aNonPureFunction_0(void) {
    return insert-a-non-const-expression-here;
}

STAINLESS_FUNC_PURE uint32_t thisFunctionWasAnnotatedWithPure_0(void) {
    return 7;
}
```

A couple of comments:
- The suggested source includes a preprocessor macro definition of `STAINLESS_FUNC_PURE`
- `STAINLESS_FUNC_PURE` should be supported by clang, GCC and most compatible compilers. It will evaluate to nothing for incompatible compilers (MSVC).
- GenC should add the `STAINLESS_FUNC_PURE` attribute to auto-generated functions that represent constants also
- Ideally, GenC would infer purity of functions automatically and add the attribute for those

jad-hamza:
I made PR #1119 for this, but it doesn't work at the moment because Stainless considers `println` to be pure. So the annotation is added to functions that just do some printing, and `gcc` simplifies them to printing nothing.

[...]

Issue #696, opened 2019-09-13
Ghost elimination on multiple parameter groups


jad-hamza:
```scala
object Test {
  def f(x1: Int)(x2: Int, x3: Int)(x4: Int, x5: Int, x6: Int) = 0
  f(1)(2, 3)(4, 5, 6)
}
```
The expression `f(1)(2, 3)(4, 5, 6)` goes through this case:

https://github.com/epfl-lara/stainless/blob/e6ad1efecb10a916d6b76672c37265e2451ee99a/frontends/scalac/src/main/scala/stainless/frontends/scalac/GhostAccessRewriter.scala#L65-70

as `Apply(f(1)(2, 3), List(4, 5, 6))` and is transformed as `f(1)(2, 3)(4)`.

The reason is that the lookup `f.symbol.info` returns `(x1: Int)(x2: Int, x3: Int)(x4: Int, x5: Int, x6: Int)` and `f.symbol.info.params` returns the first parameter group `(x1: Int)`. So the arguments `args = List(4, 5, 6)` are zipped with  `x1` instead of `List(x4,x5,x6)` in https://github.com/epfl-lara/stainless/blob/e6ad1efecb10a916d6b76672c37265e2451ee99a/frontends/scalac/src/main/scala/stainless/frontends/scalac/GhostAccessRewriter.scala#L76

Another issue is that `fun` goes through `super.transform` instead of `transform` itself.

It looks like the code was written with the assumption that `fun` is a named function, but that may have been broken by a recent Scala release? Is there a way to get the type of `fun` directly, then we could use that instead of `f.symbol.info`?

PR #1119, opened 2021-07-14
Add pure attribute to pure functions

jad-hamza:
Fix https://github.com/epfl-lara/stainless/issues/1117


jad-hamza:
Also fix https://github.com/epfl-lara/stainless/issues/696

PR #1123, opened 2021-07-27
Fix computation of targets in let expressions

jad-hamza:
The example `LetTargets` is failing in the current master branch, because the computation of targets in let expressions duplicates the `path` (once for the bound expression `e`, and once for the body `b`). We get targets which are not well-formed.

### DIFF
--- a/core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
+++ b/core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
@@ -43,13 +43,16 @@ trait FunctionInlining extends CachingPhase with IdentitySorts { self =>
         case _ => super.transform(expr)
       }
 
-      // function bodies that can be inlined directly, without bindings
+      // values that can be inlined directly, without being let-bound
       private def isValue(e: Expr): Boolean = e match {
+        case _: Variable => true
         case UnitLiteral() => true
         case BooleanLiteral(_) => true
         case IntegerLiteral(_) => true
         case BVLiteral(_, _, _) => true
         case Tuple(es) => es.forall(isValue)
+        case FiniteArray(args, base) => args.forall(isValue)
+        case LargeArray(elems, default, _, _) => elems.map(_._2).forall(isValue) && isValue(default)
         case ADT(_, _, args) => args.forall(isValue)
         case _ => false
       }
@@ -68,8 +71,9 @@ trait FunctionInlining extends CachingPhase with IdentitySorts { self =>
         if (!willInline) return fi
 
         val specced = BodyWithSpecs(tfd.fullBody)
-        if (specced.specs.isEmpty && isValue(tfd.fullBody)) {
-          exprOps.freshenLocals(tfd.fullBody)
+        // simple path for inlining when all arguments are values, and the function's body doesn't contain other function invocations
+        if (specced.specs.isEmpty && args.forall(isValue) && !exprOps.containsFunctionCalls(tfd.fullBody)) {
+          exprOps.replaceFromSymbols(tfd.params.zip(args).toMap, exprOps.freshenLocals(tfd.fullBody))
         } else {
           // We need to keep the body as-is for `@synthetic` methods, such as
           // `copy` or implicit conversions for implicit classes, in order to
--- a/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
@@ -597,9 +597,8 @@ trait EffectsAnalyzer extends oo.CachingPhase {
       getTargets(b, kind, path).map(_.bind(vd, e))
 
     case Let(vd, e, b) =>
-      val eEffects = getTargets(e, kind, path)
       getTargets(b, kind, path).map(_.bind(vd, e)).flatMap { be =>
-        if (be.receiver == vd.toVariable) eEffects.map(_ append be)
+        if (be.receiver == vd.toVariable) getTargets(e, kind, be.path.path)
         else Set(be)
       }
 
@@ -721,6 +720,7 @@ trait EffectsAnalyzer extends oo.CachingPhase {
 
     def rec(expr: Expr, env: Map[Variable, Effect]): Set[Effect] = expr match {
       case Let(vd, e, b) if symbols.isMutableType(vd.tpe) =>
+
         if ((variablesOf(e) & variablesOf(b)).forall(v => !isMutableType(v.tpe))) {
           val effe = rec(e, env)
           val newEnv = (variablesOf(b) ++ freeVars).map(v => v -> ModifyingEffect(v, Path.empty)).toMap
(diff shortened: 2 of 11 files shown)


core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
```diff
@@ -43,13 +43,17 @@ trait FunctionInlining extends CachingPhase with IdentitySorts { self =>
...
-      // function bodies that can be inlined directly, without bindings
+      // values that can be inlined directly, without being let-bound
       private def isValue(e: Expr): Boolean = e match {
+        case _: Variable => true
         case UnitLiteral() => true
         case BooleanLiteral(_) => true
         case IntegerLiteral(_) => true
         case BVLiteral(_, _, _) => true
         case Tuple(es) => es.forall(isValue)
+        case ADTSelector(o, _) => isValue(o)
```
samarion, line 54: This is not a pure expression (consider `Nil().head`). Using `replaceFromSymbols` might make it disappear from the inlined expression which would change the semantics.

core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
```diff
@@ -68,8 +72,9 @@ trait FunctionInlining extends CachingPhase with IdentitySorts { self =>
         if (!willInline) return fi
 
         val specced = BodyWithSpecs(tfd.fullBody)
-        if (specced.specs.isEmpty && isValue(tfd.fullBody)) {
-          exprOps.freshenLocals(tfd.fullBody)
+        // simple path for inlining when all arguments are values, and the function's body doesn't contain other function invocations
+        if (specced.specs.isEmpty && args.forall(isValue) && !exprOps.exists { _.isInstanceOf[FunctionInvocation] }(tfd.fullBody)) {
+          exprOps.replaceFromSymbols(tfd.params.zip(args).toMap, exprOps.freshenLocals(tfd.fullBody))
```
samarion, line 77: Doesn't this mean we'll be duplicating potential checks in the function's body?

core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
```diff
@@ -68,8 +72,9 @@ trait FunctionInlining extends CachingPhase with IdentitySorts { self =>
         if (!willInline) return fi
 
         val specced = BodyWithSpecs(tfd.fullBody)
-        if (specced.specs.isEmpty && isValue(tfd.fullBody)) {
-          exprOps.freshenLocals(tfd.fullBody)
+        // simple path for inlining when all arguments are values, and the function's body doesn't contain other function invocations
+        if (specced.specs.isEmpty && args.forall(isValue) && !exprOps.exists { _.isInstanceOf[FunctionInvocation] }(tfd.fullBody)) {
```
samarion, line 76: There's a utility `exprOps.containsFunctionCalls` which implements the `FunctionInvocation` check.

samarion (commented):
I'm not sure I understand, is the function inlining change related to the target computation issue?

core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
jad-hamza: Yes it can happen. Maybe we should let the user decide when to avoid let-bindings?

jad-hamza:
> I'm not sure I understand, is the function inlining change related to the target computation issue?

No sorry, I've added a bunch of unrelated commits that you can ignore, only "Fix computation of targets in let expressions" is related to that

core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
jad-hamza: Hmm yes good point

core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
jad-hamza: Thanks

core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala
samarion: Could you just wrap the full body in `@unchecked`?

jad-hamza: This comment suggests that if we do that without let-binding, the `@unchecked` annotation could leak to not checking postconditions (I don't remember why):
https://github.com/epfl-lara/stainless/blob/2009c7687f321d4bcb9d061d1a7632a21eed99e3/core/src/main/scala/stainless/extraction/inlining/FunctionInlining.scala#L122-L124

samarion: But this special case is guarded by specs being empty (so no postcondition)?

jad-hamza: There could be specs in the caller where the body gets inlined

samarion: Yes but those shouldn't be affected by the `@unchecked`.

Although I have to agree I don't really understand the comment, so maybe it's really an issue.

jad-hamza: The comment might be obsolete, I've tried to remove it and it worked on a small example (i.e. the postcondition of the caller of an inlined function wasn't ignored), let's see what happens on the full test suite.

jad-hamza: I found an example where we should avoid the unchecked annotation except in a let binding:

```scala
object InliningUnchecked2 {
  @inline def nonNegative(x: Int): Boolean = x >= 0

  def test: Int = {
    -10
  }.ensuring(nonNegative _)
}
```

In postconditions, the unchecked annotation means we don't check the conjunct. I think we already had this discussion once, we're basically using unchecked annotation to mean slightly different things depending on placement.

samarion: Ahhh, I understand. Wow, that's a bit dangerous. It means that if we inline lets before going to type checking, we'll stop checking post-conditions containing inlined calls?

Sounds like we should split `@unchecked` into two different annotations.

jad-hamza: (sorry I edited your post by mistake)

> Ahhh, I understand. Wow, that's a bit dangerous. It means that if we inline lets before going to type checking, we'll stop checking post-conditions containing inlined calls?

Yes for instance this example (`InliningUnchecked2`) would go through.

> Sounds like we should split `@unchecked` into two different annotations.

Yes good idea I'll do the split in another PR. Can we give a (more or less) formal definition to the two annotations?
* `unchecked`: don't generate VCs when type-checking this expression (that's a bit ambiguous)
* `dropConjunct`: drop this conjunct in VCs

Thanks for the review! I'll merge this PR for now and address this issue (and potentially others) in another one.

gsps: Some extra considerations to take into account when reshuffling the `Unchecked` annotation: #1136

PR #1125, opened 2021-07-29
Split Unchecked annotation into DropVCs and DropConjunct

jad-hamza:
Follow-up to discussion here: https://github.com/epfl-lara/stainless/pull/1123#discussion_r679089745


vkuncak:
@samarion : this seems important. Does it look OK to you?

PR #1149, opened 2021-09-02
Less casts in imperative phase, more freshening of local variables, and add SplitAnd tree

jad-hamza:
* Less casts and more freshening in imperative phase
* Add SplitAnd tree (&&&) to split verification conditions

### DIFF
--- a/frontends/library/stainless/lang/package.scala
+++ b/frontends/library/stainless/lang/package.scala
@@ -30,6 +30,11 @@ package object lang {
     def ==>(that: => Boolean): Boolean = {
       if (underlying) that else true
     }
+
+    // Use this "and" operator instead of `&&` when you want verification conditions to be split
+    def &&&(that: => Boolean): Boolean = {
+      if (underlying) that else false
+    }
   }
 
   @library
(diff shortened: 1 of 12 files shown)


frontends/library/stainless/lang/package.scala
```diff
@@ -30,6 +30,11 @@ package object lang {
     def ==>(that: => Boolean): Boolean = {
       if (underlying) that else true
     }
+
+    // Use this "and" operator instead of `&&` when you want verification conditions to be split
+    def &&&(that: => Boolean): Boolean = {
+      if (underlying) that else false
```
samarion, line 36: This doesn't seem consistent with verification / termination checking. IIUC, termination and verification assume both sides of the condition are evaluated.

jad-hamza: I'd like it to behave like normal `&&` (for instance in the `TypeChecker` I add `lhs = true` to the context when type-checking the rhs. Any other change I should do?

samarion: You would also need to change the codegen evaluator to handle the new tree. It would probably be safer to use an annotation instead of a dedicated AST node to make sure it doesn't interfere with the rest of Stainless.

jad-hamza: Alright thanks, I've added an annotation

PR #1160, opened 2021-09-10
Fix issue where local function calls in postcondition are wrongly transformed in imperative phase

### LINKED ISSUE #1159: ApplyLetRec arguments are not properly mapped in postconditions
The `inside` function is accepted by Stainless.

```scala
import stainless.lang._

object Test {

  def foo(): Unit = {
    var i = 0
    def isZero = i == 0

    def inside: Unit = {
      require(i <= 10)
      i += 1
    }.ensuring(_ =>
      isZero
    )

    inside
  }

}
```

The call `isZero` is transformed to `isZero(i)` where `i` refers to the initial value of `i` instead of the new value.

jad-hamza:
Fix #1159

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/ImperativeCodeElimination.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/ImperativeCodeElimination.scala
@@ -273,17 +273,16 @@ trait ImperativeCodeElimination
                 val newSpecs = specs.map {
                   case Postcondition(post @ Lambda(Seq(res), postBody)) =>
                     val newRes = ValDef(res.id.freshen, newReturnType)
-
-                    val newBody = replaceSingle(
+                    val subst =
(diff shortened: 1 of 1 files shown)


jad-hamza:
@samarion does this make sense? The problem was that in postconditions, `ImperativeCodeElimination` adds arguments to `ApplyLetRec` for which the proper mapping wasn't done.

samarion:
This looks very iffy to me, I don't think applying the same substitution twice would be the right fix.

Could it be that the call to `toFunction` needs a different `State` argument? Maybe closer to the one used for `wrappedBody`? I'm unfortunately not very clear with the `ImperativeCodeElimination` transformation...

vkuncak:
@mario-bucev : is this related to what was just merged?

mario-bucev:
Yes, though the fix was not the same

PR #1207, opened 2021-11-19
Add cCode.noMangling annotation and split defines into header and C files

jad-hamza:


vkuncak:
@jad-hamza : can we merge this?

jad-hamza:
Yes I think so (if it works for the project)

vkuncak:
See #1233 instead

vkuncak:
Apparently merged through https://github.com/epfl-lara/stainless/pull/1233

PR #1208, opened 2021-11-19
Reenable actors tests

mario-bucev:
Note that since `scripted` is only run on nightly, we cannot see whether the tests actually succeed.


jad-hamza:
If it works, great :) Thanks

PR #1210, opened 2021-12-08
PC for local classes capturing variables

### LINKED ISSUE #1209: Assertion failed: "Cannot add already used variable to path" in context of local classes
The following code:
```scala
object InnerClassesInv {

  def test(x: Int): Unit = {
    val y = x
    val z = y

    case class Local()

    val dummy = 0 // To avoid an unrelated NPE in CodeExtraction
  }

}
```
triggers the following assertion failure:
```
java.lang.AssertionError: assertion failed: Cannot add already used variable y$21: Int to path:
val y$21: Int = x$102
val z$10: Int = y$21
true
    at scala.Predef$.assert(Predef.scala:279)
    at inox.ast.Paths$Path.withBinding(Paths.scala:112)
    at stainless.extraction.innerclasses.InnerClasses$TransformerContext$Context.withBinding(InnerClasses.scala:164)
    at stainless.extraction.innerclasses.InnerClasses$TransformerContext$Context.withBinding(InnerClasses.scala:122)
    ...
```

`Local` captures `x`, `y` and `z` to generate the following invariant:
```scala
@invariant
@method(Local)
def inv: Boolean = {
  val y: Int = this.x
  val z: Int = y
  true
}
```
However, `y` and `z` are not freshened: when we proceed to process `inv` for lifting its potential local class, we try to add these `y` and `z` to the current PC which already contains them.

mario-bucev:
Fixes #1209

### DIFF
--- a/core/src/main/scala/stainless/extraction/innerclasses/InnerClasses.scala
+++ b/core/src/main/scala/stainless/extraction/innerclasses/InnerClasses.scala
@@ -74,7 +74,7 @@ class InnerClasses(override val s: Trees, override val t: methods.Trees)
 
     /** Represent a substitution for a lifted local class */
     case class ClassSubst(
-      cd: ClassDef,               // Lifted classd
+      cd: ClassDef,               // Lifted class
       methods: Seq[FunDef],       // Lifted methods
       typeMembers: Seq[TypeDef],  // Lifted type members
       // Closed over type params: the first element of the tuple corresponds to the captured free type parameter
@@ -359,15 +359,15 @@ class InnerClasses(override val s: Trees, override val t: methods.Trees)
     }
 
     /** Current path condition expressed as a class invariant, if not trivial */
-    def pathConditionToInvariant(pathCondition: Expr, lcd: LocalClassDef): Option[LocalMethodDef] = {
-      pathCondition match {
+    def pathConditionToInvariant(path: Path, lcd: LocalClassDef): Option[LocalMethodDef] = {
+      pathToClause(path, lcd) match {
         case BooleanLiteral(true) => None
-        case _ => Some(LocalMethodDef(
+        case pc => Some(LocalMethodDef(
           ast.SymbolIdentifier("inv"),
           Seq.empty,
           Seq.empty,
           BooleanType().setPos(lcd),
-          pathCondition.setPos(lcd),
+          pc.setPos(lcd),
           Seq(IsInvariant, IsMethodOf(lcd.id))
         ).setPos(lcd))
       }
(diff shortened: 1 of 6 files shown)


core/src/main/scala/stainless/extraction/innerclasses/InnerClasses.scala
```diff
@@ -421,23 +468,40 @@ trait InnerClasses
...
+          // If the pathCondition has a let binding of a variable that the local class captures,
+          // we transform this let binding into an equality.
+          // For instance, if the pathCondition contains `let y = 42 in body`, we transform that expression
+          // into `this.y == 42 && rec(body)`.
+          // See test cases InnerClassesInvariants for examples.
+          // Note: the check to ensure that vd is not a var is redundant because checkValidLiftedClass
+          // will reject the definition anyway.
+          case Let(vd, value, body) if (freeVarsMap contains vd.toVariable) && !(vd.flags contains IsVar) =>
+            Some(And(Equals(freeVarsMap(vd.toVariable), value), body))
+
```
mario-bucev, line 481: I'm not sure whether this is sound or not.
If we allow it, we can express properties such as https://github.com/epfl-lara/stainless/blob/60d0b87e233cdc1f69d6786ecd5b363341185e6e/frontends/benchmarks/verification/valid/InnerClassesInvariants.scala#L18-L30
On the other hand, we can get into trouble if the value is mutable (e.g. through a wrapper). 
It seems that such cases get caught up by the (old) imperative phase as illegal aliasing (like https://github.com/epfl-lara/stainless/blob/60d0b87e233cdc1f69d6786ecd5b363341185e6e/frontends/benchmarks/extraction/invalid/InnerClassesInvariants2.scala).

samarion: I think we should be able to rely on an Inox `Path` to take care of outer bindings (similarly to [FunctionClosure](https://github.com/epfl-lara/stainless/blob/master/core/src/main/scala/stainless/extraction/innerfuns/FunctionClosure.scala).  
Replacing bindings with equality is not ideal because of types for which equality is unexpected (_e.g._ first-class functions).

mario-bucev: Good catch! I'll see what I can do.

mario-bucev: I have pushed a version trying to mimic `FunctionClosure`, but that version sadly breaks some benchmarks involving dynamic dispatch.
For instance, if we have:
```scala
object Test {
  abstract class Foo {
    def bar: Int
  }

  def test(x: Int): Foo = {
    require(x >= 0 && x < 100)

    def makeFoo(m: Int): Foo = {
      require(m == 0)
      new Foo {
        def bar: Int = m + x
      }
    }

    makeFoo(0)
  }
}
```
Then we have the following definitions of interest:
```scala
abstract class Foo
case class $anon(m: Int, x: Int) extends Foo

@derived(test)
def bar(thiss: Foo): Int = {
  require(thiss.isInstanceOf[$anon])
  // Similar to what FunctionClosure does
  require(((thiss.x >= 0)): @dropConjunct  @DropVCs  &&& (((thiss.x < 100)): @dropConjunct  @DropVCs  &&& ((thiss.m == 0)): @dropConjunct  @DropVCs ))
  thiss.m + thiss.x
}

@derived(bar)
def bar(thiss: Foo): Int = if (thiss.isInstanceOf[$anon]) {
  // Generates a VC that requires the preconditions of the anonyoumous class bar method 
  // to be satisfied, including for the (assumed) PC.
  bar(({
    assert(thiss.isInstanceOf[$anon], "Cast error")
    thiss
  }): @DropVCs )
} else {
  bar(({
    assert(thiss.isInstanceOf[FooExt], "Cast error")
    thiss
  }): @DropVCs )
}
```
I was thinking of replacing the second `Require` in:
https://github.com/epfl-lara/stainless/blob/83af7b4fb2f8ab42bd23f4ff6bce257a92f06286/core/src/main/scala/stainless/extraction/innerclasses/InnerClasses.scala#L465-L472
with `Assume`, but that sounds a bit too easy, and if it wasn't done in `FunctionClosure`, there is surely a reason for that :p

samarion: Oh I think you may be following `FunctionClosure` a bit too literally here. I just meant we could use a `Path` to produce a better ADT invariant in the presence of let-bindings, _e.g._ by using [Path.and(previous invariant)](https://github.com/epfl-lara/inox/blob/e74baa7fdb9a941f4e4ce8e880b96a32a96c8b59/src/main/scala/inox/ast/Paths.scala#L282). This would allow you to close over lets without resorting to equality.

mario-bucev: Apologize, but I must be missing something really obvious! 
I cannot find a way of specifying an ADT invariant that can reconcile a class field selection with a let-binding.
For instance, if we have:

```scala
def test(x: BigInt): Unit = {
  require(x > 10)
  val y = x + 10
  // The PC is `x > 10 && let y = x + 10 in true`
  case class Local() {
    def hello: Unit = {
      assert(x > 10)
      assert(y > 20)
    }
  }
}
```

If we capture `y`, it seems we get stuck.
```scala
case class HoistedLocal(x: BigInt, y: BigInt) {
  def inv: Boolean = {
    this.x > 10 // Ok; x is not let-bound (would also be fine if we were to require x to be equal to some value)
    // && this.y == x + 10 // Using equality, not what we want
    // && let y = x + 10 in true // Does not say anything about this.y
  }
  def hello: Unit = {
    assert(this.x > 10)
    assert(this.y > 20) // Cannot verify, as nothing is said about this.y
  }
}
```

On the other hand, the `FunctionClosure` trick is quite elegant I must say:
```scala
case class HoistedLocal(x: BigInt) { // y not captured
  // No `inv` needed, we essentially inline the invariants at the beginning of each method.
  def hello: Unit = {
    require(@dropConjunct @DropVCs (this.x > 10))
    val y = @DropVCs (this.x + 10)
    assert(this.x > 10)
    assert(y > 20)
  }
}
```
But alas, we lose the PC needed to verify the `require` in the dispatcher (?) for `hello`.

samarion: Ah, I see what you mean. I wasn't thinking of completely dropping the ADT invariant but that's an interesting idea! Too bad it doesn't work for dynamic dispatch :(

I was rather thinking of keeping the ADT invariant but injecting the let-binding into the invariant. The post-transformation `HoistedLocal` should look something like:
```scala
case class HoistedLocal(x: BigInt) { // y not captured
  def inv: Boolean = {
    this.x > 10 &&
    { val y = x + 10; true } // this block can actually be dropped, it's just to make things clear
  }

  def hello: Unit = {
    val y = this.x + 10
    assert(this.x > 10)
    assert(y > 20)
  }
}
```

You probably need to extract the let-bindings from the path when constructing the transformed method body (with `path.bindings`) because the conditions will already be part of the ADT invariant.

Note that you'll have duplicated let-bindings (in the invariant and in the method pre-condition) but that isn't a problem as long as the variables are freshened and they are annotated with `@DropVCs`.

mario-bucev: Keeping the ADT invariant is actually a good idea; I like your suggestion!
I'll push an updated version once I've locally run the integration test.

mario-bucev:
This PR should be ready to be re-reviewed :)

Issue #1218, opened 2021-12-22
Illegal aliasing with arrays update not caught


mario-bucev:
The following code should be rejected, but isn't:

```scala
object TestAlias {
  case class B(var x: Int)

  def test(arr: Array[B]): Unit = {
    require(arr.length >= 2)
    val arr2 = arr.updated(1, B(42))
    f(arr(0), arr2(0)) // Illegal aliasing not caught 
  }

  def f(b1: B, b2: B): Unit = {
    val x1 = b1.x
    b2.x = 50
    assert(b1.x == x1)
  }
}
```

PR #1219, opened 2021-12-22
Fix #1051 (Aliasing in constructors)

### LINKED ISSUE #1051: Illegal aliasing in ADT constructors is not caught
from @vkuncak

This assertion goes through because we assume that mutable elements are not aliased, but we don't check for aliasing when constructing `A(b, b)`

```scala
object TestAlias {
  case class A(val left: B, val right: B)
  case class B(var x: Int)

  def t: Unit = {
    val b = B(42)
    f(A(b,b))
  }

  def f(a: A): Unit = {
    val x1 = a.left.x
    a.right.x = 50
    assert(a.left.x == x1)
  }
}
```

mario-bucev:


core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -274,27 +274,118 @@ trait AntiAliasing
...
+              }
+            case ArraySelect(arr, i) =>
+              accessorsOf(arr).map {
+                case (v, accs) => (v, accs :+ ArrayAccessor(i))
+              }
+            // If we have something like v.asInstanceOf[T].a.b.asInstanceOf[S].c, we are still interested in retrieving the accessed a.b.c
+            case AsInstanceOf(e, _) => accessorsOf(e)
+            case Annotated(e, _) => accessorsOf(e)
+            case _ => None
+          }
```
samarion, line 299: Would using `Try(getAllTargets(e))` also work instead of defining yet another target finder?

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -274,27 +274,118 @@ trait AntiAliasing
...
+                //   Ideally, we would like to say that the targets of arr.updated(i, expr)
+                //   are the targets of expr plus "the targets of arr at all indices except i".
+                //   For that, we could introduce a new Accessor enum that behaves like a quantified ArrayAccessor...
+                case ArrayUpdated(arr, i, value) =>
+                  traverse(value)
+                case IsInstanceOf(_, _) =>
+                  // a.b.c.isInstanceOf[T] is not an actual target, so we stop the traversal.
+                  ()
+                case _ => super.traverse(e)
+              }
```
samarion, line 340: This traverser feels a little dangerous. Your TODO is a valid point which has soundness implications, so I'd rather err on the side of safety.

I agree with the example in your `collectTargets` doc string, so I would suggest explicitly handling the cases were we are _sure_ that no potentially aliasing targets are missed. Something like:
```scala
def traverse(e: Expr): Unit = e match {
  case ADT(id, tps, args) =>
    args.foreach(traverse)
  case ClassConstructor(_, args) =>
    args.foreach(traverse)
  case _ =>
    Try(getAllTargets(e)).toOption.getOrElse(
      exprOps.variablesOf(e).filter(v => isMutableType(v.tpe)).map(Target(_, None, Path.empty)))
}
```

Actually, if you want you can consider adding a new effect kind to the [getTargets](https://github.com/epfl-lara/stainless/blob/087a85e4437e2e156689d7a161c4c437a0052891/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala#L482) method (_e.g._ `AliasingEffect`) and adapt the `ADT` and `ClassConstructor` cases to recurse into their sub-expressions instead of returning the empty set. I'm all for consolidating similar logic in a single place.

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -543,6 +634,10 @@ trait AntiAliasing
               Application(transform(callee, env), args.map(transform(_, env))).copiedFrom(app)
             }
 
+          case cc @ ClassConstructor(_, args) =>
+            checkAliasing(cc, args)
+            super.transform(cc, env)
+
```
samarion, line 640: I'm not 100% sure, but I don't think we need this case. I believe it's legal to write `A(b, b)` when not trying to pass it to a function. Do you have a benchmark which requires this check?

I believe most/all of the testcases listed in the issue are fixed by your change to `checkAliasing` which ensures there's no aliasing within an argument (good catch btw!)

samarion (commented):
While we're at it, I believe we're missing a `checkAliasing` call in the `Let` case, in the second branch of the if-expression where targets couldn't be computed precisely but there is no overlap between the mutable free variables of the let expression and body.

This case is handled similarly to a function call so we need to check that there's no aliasing within the `newExpr` (as if it were a function call argument).

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: Ah, good point! It subsumes this `accessorOf`

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: I have removed that case to see what happens, and the following case does not get caught:
```scala
// Another good test case :)
object TestAlias {
  case class A(val left: B, val right: B) {
    def changeLeft(newX: Int): Unit = {
      val rX = right.x
      left.x = newX
      assert(rX == right.x)
    }
  }
  case class B(var x: Int)

  def test: Unit = {
    val b = B(42)
    val a = A(b, b)
    a.changeLeft(24)
    changeLeft(a, 24)
  }

  def changeLeft(a: A, newX: Int): Unit = {
    val rX = a.right.x
    a.left.x = newX
    assert(rX == a.right.x)
  }
}
```

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: That's true. This case was to allow patterns such as:
https://github.com/epfl-lara/stainless/blob/190d0b024b5af906743e54e2b1012b7cf9d9f6dc/frontends/benchmarks/verification/valid/i1051c.scala
but a soundness hole is more worrying than permitting to write the above expression (that one should be able to rewrite it to `val c2 = a.container.arr(i); a.container.arr(i) = B(C(123), c2)` anyway)

samarion: I'm not sure I understand, I don't see any argument aliasing in the example you linked?

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
samarion: Ah, that's interesting!

I wonder if it's because the call to `getAllTargets` returns a set of targets (as opposed to a list) so we don't realize that the `b` target occurs twice in different aliasing positions. If you try adapting the `getTargets` method to return a `Seq` instead of a `Set`, does the testcase still pass?

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: If I'm not mistaken, "aliasing" is introduced by the array update rewrite:
```scala
a = A(
  a.b,
  Container(
    a.container.arr.updated(i, B(C(123), a.container.arr(i).c2)),
//  ^^^^^^^^^^^^^^^                      ^^^^^^^^^^^^^^^^^^        
    a.container.other
  )
)
```
This should not count as aliasing but the traversal of `ArrayUpdated` will collect both the targets `a.container.arr` and `a.container.arr(i)` if we do not add a case in `collector#traverse`

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: It still passes (in the sense it does not get rejected)!

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
samarion: Hmm, I don't think we're supposed to be checking aliasing after the anti-aliasing transformation, it doesn't make any guarantees on outputs being aliasing-free. Might be worth debugging the code path a bit.

I went and checked `getTargets` to compare, but these lines also look fishy to me:
https://github.com/epfl-lara/stainless/blob/087a85e4437e2e156689d7a161c4c437a0052891/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala#L573-L574

Could you try removing them to see what breaks? :)

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
samarion: Okay, then that's not it :) I guess `getAllTargets` is throwing an exception because of this condition:
https://github.com/epfl-lara/stainless/blob/087a85e4437e2e156689d7a161c4c437a0052891/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala#L509-L510

So we're still getting set semantics from `exprOps.variablesOf`.

This seems to show that `checkAliasing` still isn't quite strict enough despite your new check. Maybe we need to make the `collector` traversal a bit more aggressive than what I proposed above, like:
```scala
def traverse(e: Expr): Unit = e match {
  case ADT(id, tps, args) =>
    args.foreach(traverse)
  case ClassConstructor(_, args) =>
    args.foreach(traverse)
  case _ =>
    try {
      results ++= getAllTargets(e)
    } catch {
      case _: MalformedStainlessCode => super.traverse(e)
    }
}
```
Here we're making sure to never use set semantics for imprecise target accumulation. Maybe it will help :)

mario-bucev: It still goes through!

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: https://github.com/epfl-lara/stainless/issues/1218 seems to be taken care of and everything seems to work fine except for arrays update as in the above case (I'll see what I can do to move aliasing check before the transformation).

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
samarion: Argh! Well I don't have any other ideas off the top of my head... maybe we really do need to check aliasing at construction time. In that case, you'll also need to add a similar check to `ADT` and `Tuple` expressions.

mario-bucev: Haha I forgot about those two, thank you for pointing that out :)

samarion: Oh, and `FiniteArray` too. I hope we're not missing anything else :)

vkuncak:
I think this discussion suggests that we really need to document some basic assumptions of our effect analysis. The approach in the thesis of Regis was rather straightforward, but this clearly evolved beyond that in an attempt to be more flexible. But we need to write down the invariants that this analysis aims to enforce, especially across function call boundaries.

mario-bucev:
I have pushed a version that I hope covers all remarks. I've also tried to do something with the `ArrayUpdated` case that I think is sound but over-restrictive (as the failing tests witness...): 
https://github.com/epfl-lara/stainless/blob/31ac041fb61ca3b446bc0399693dd08ccea45795/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L318-L369

Allowing more correct programs would require gathering more information about the targets collected in `others`:
https://github.com/epfl-lara/stainless/blob/31ac041fb61ca3b446bc0399693dd08ccea45795/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L345-L351

Would pursuing this route make sense?

vkuncak:
If you addressed the remarks, it should be merged. We may revisit this topic another time, but for now this seems like a great improvement already. One thing that would be grateful if you could document how this behaves from user's point of view. As far as I can see, nothing is documented in https://github.com/epfl-lara/stainless/blob/master/core/src/sphinx/imperative.rst

vkuncak:
[info] -  21: imperative/valid/ArrayOfPairs  *** FAILED *** (15 milliseconds)

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -543,6 +660,30 @@ trait AntiAliasing
...
+            checkAliasing(tpl, exprs, env)
+            super.transform(tpl, env)
+
+          case fa@FiniteArray(elems, _) =>
+            checkAliasing(fa, elems, env)
+            super.transform(fa, env)
+
+          case fs@FiniteSet(elems, _) =>
+            checkAliasing(fs, elems, env)
+            super.transform(fs, env)
```
samarion, line 681: IIRC `FiniteSet` is ok to include aliasing in `elems` because they can't be recovered from inside the set.

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -272,29 +269,149 @@ trait AntiAliasing
...
+          var result = Seq.empty[Target]
+          object collector extends SelfTreeTraverser {
+            // traverse(e) but without modifying `result`
+            def traverseNoAddedResults(e: Expr): Seq[Target] = {
+              val oldLength = result.length
+              traverse(e)
+              val (old, added) = result.splitAt(oldLength)
+              result = old
+              added
+            }
```
samarion, line 300: Couldn't you just call `collectTargetsForAliasing` recursively instead? That seems a bit cleaner.

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -272,29 +269,149 @@ trait AntiAliasing
...
+                // First, we compute the targets of `arr` and `upd` (we do not need to do so for `i` because it is an immutable type (Int))
+                // Then, we make sure that the targets of `upd` only contain paths with the prefix arr(i).
+                // If not, we conservatively reject the program.
+                // We check afterwards for aliasing within the targets of `upd`.
+                // Finally, we add the targets of `arr` to the results (but not the ones of `upd`, which may contain `arr`).
+
+                val arrTargets = traverseNoAddedResults(arr)
+                val updTargets = traverseNoAddedResults(upd)
+
+                val arrSelect = exprOps.stripAllAnnotations(ArraySelect(arr, i))
```
samarion, line 344: Do you think it would be possible to use `getAllTargets` here instead? Then your partition logic below could just check whether the `ArraySelect(arr, i)` paths are prefixes of the `updTargets` path. Am I missing something?

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -272,29 +269,149 @@ trait AntiAliasing
...
+        //    case class A(val left: B, val right: B)
+        //    case class B(var c1: C, var c2: C)
+        //    case class C(var x: Int)
+        //    val res: (Unit, A) = ...
+        //    A(B(res._2.left.c1, a.left.c2), a.right)
+        // collectTargetsForAliasing on the last expression would give us:
+        //    Target(res, None, ._2.left.c1),
+        //    Target(a, None, .left.c2)
+        //    Target(a, None, .right)
+        def collectTargetsForAliasing(e: Expr, env: Env): Seq[Target] = {
```
samarion, line 290: Is the `env` parameter ever used? Same remark for `checkAliasingWithinExpr` and `checkAliasing`.

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -272,29 +269,149 @@ trait AntiAliasing
...
+                args.foreach(traverse)
+
+              case Tuple(exprs) =>
+                exprs.foreach(traverse)
+
+              case FiniteArray(elems, _) =>
+                elems.foreach(traverse)
+
+              case FiniteSet(elems, _) =>
+                elems.foreach(traverse)
```
samarion, line 316: This has immutable type so it should have empty targets in the default case (I think :)).

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
samarion: Not super important but I think we could revert `getTargets` to set semantics since we ruled that out as the cause.

vkuncak:
Actually we have a way to convert set to list. It's controversial but best
not to assume it does not exist.

samarion:
That conversion relies on a choose which will (arguably) produce fresh values...

But if you want to assume values inside sets aren't guaranteed to be immutable, then the logic should be extended to maps and bags as well.

samarion:
Also, I'm pretty sure there are other places in the imperative transformation where we assume values added to sets become immutable.

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: Good suggestion!

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: Oops, I forgot to remove it from previous trials

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: If I'm not mistaken, for an expression such as `arr.updated(i, CaseClass(x))`, we would need a fallback mechanism (similar to `traverseDefaultCase`) because `getAllTargets(CaseClass(x))` would throw an exception instead of getting the targets of `x`?

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: Thanks, I'll remove that case as well as the other mentioned above :)

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
samarion: Yes, you probably want to wrap the `getAllTargets` into a `Try()` constructor just in case.

I'm not sure about your example though, I meant we would be calling `getAllTargets(arr(i))` in that case (which won't crash). And if the call throws an exception, then the partitioning logic below should just fall into the `others` case.

mario-bucev: It's true, we would however reject field updates within an array such as:
```scala
object Test {
  case class A(var x: B, var y: B)
  case class B(var v: Int)

  def test(arr: Array[A]): Unit = {
    require(arr.length >= 1)
    // i.e. arr = arr.updated(0, A(B(123), arr(0).y))
    arr(0).x = B(123)
  }
}
``` 
due to `getAllTargets(upd)` throwing (we could maybe define a local function that has the same behavior as `getAllTargets` except that it traverses the `ClassConstructor` case, a bit like what `collectTargetsForAliasing` does without the `ArrayUpdated` case).

samarion: I'm still confused. When are we calling `getAllTargets(upd)`?

I was thinking the logic would look something like this:
```scala
val selectTargets = Try(getAllTargets(ArraySelect(arr, i))).toOption.getOrElse(Set.empty)
val (atUpdatedLoc, others) = updTargets.partition {
  case t @ Target(v, None, path) => selectTargets.exists(t prefixOf _)
  case _ => false
}
```

mario-bucev: Oh, so only the `ArraySelect` would get through `getAllTargets` while still using the `traverseNoAddedResult` for `updTargets` (that should be replaced with a recursive call to `collectTargetsForAliasing`, as suggested)?
https://github.com/epfl-lara/stainless/blob/31ac041fb61ca3b446bc0399693dd08ccea45795/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L342
(Thanks for bearing with me haha)

samarion: Exactly!

mario-bucev: Ahh, now everything makes sense, thanks :)

vkuncak:
Can you re-target main branch, @mario-bucev ?

vkuncak:
@mario-bucev will split the PR into one for constructors to be merged, and one for `array.updated` issue that is to be addressed eventually.

mario-bucev:
Exactly, right now I'm running the integration tests locally before pushing

mario-bucev:
""Ready"" for rerere-re-review :see_no_evil:

PR #1224, opened 2022-01-17
Updating Inox and removing Solver reference from VCResult objects

mario-bucev:
The second change allows the `Solver` object to be garbage-collected while `checkVCs` is running.


vkuncak:
LGTM

PR #1235, opened 2022-02-17
GenC: introducing a binding to 'guard' against references created by the Referentiator

mario-bucev:
Some small tweaks to GenC so that expressions such as:
```scala
import stainless._
import stainless.lang._
import stainless.annotation._

object Refs {
  case class Ref[T](var x: T)

  case class Container[T](ref: T)

  @cCode.`export`
  def test1(v: Int): Unit = {
    placeholder(Ref(v + 123))
  }

  def placeholder(x: Ref[Int]): Unit = ()
}
```
get (roughly) translated to:
```c
void test1(int32_t v) {
    int32_t tmp = v + 123;
    placeholder(&tmp);
}
```
and not:
```c
void test1(int32_t v) {
    placeholder(&(v + 123));
}
```


core/src/main/scala/stainless/ast/SymbolIdentifier.scala
```diff
@@ -25,7 +25,9 @@ object Symbol {
 }
 
 class SymbolIdentifier private[stainless](id: Identifier, val symbol: Symbol)
-  extends Identifier(id.name, id.globalId, id.id, alwaysShowUniqueID = false)
+  extends Identifier(id.name, id.globalId, id.id, alwaysShowUniqueID = false) {
+  override def freshen: SymbolIdentifier = new SymbolIdentifier(id.freshen, symbol)
+}
```
mario-bucev, line 30: This is not related to GenC, but I've noticed that symbol identifiers get lost in case of freshening.

PR #1239, opened 2022-03-08
Experimental test cases generation

mario-bucev:
Add two new phases that attempts in creating test cases from reported counter-examples, namely `--testgen` (outputs Scala code) and `--genc-testgen` (outputs C code)


vkuncak:
Aside from identifier changes, this seems fine to merge. It's not the only functionality where there would be much more to do; the point is that it is not too complex and does not interfere with the rest.

mario-bucev:
I've changed the identifier of `TestsGeneration` :)

PR #1241, opened 2022-03-10
Do not duplicate ScalaZ3 jar (for standalone builds)

mario-bucev:
We apparently duplicated the content of ScalaZ3 jar within the Stainless jar.


vkuncak:
I see you are explicitly excluding it, but any idea why it was even pulled in and why only on Linux?

mario-bucev:
I think it was due to the assembly process merging the ScalaZ3 jar when including Inox in the resulting jar. As for why only Linux, the other platforms do not have a ScalaZ3 jar, so there is nothing to include...

Issue #1122, opened 2021-07-23
Effect on if then else expression


jad-hamza:
Issue reported by @samarion in https://github.com/epfl-lara/stainless/pull/1116

> It seems to me that we have some more general issues with conditional effects. What would happen if we try to pass something like `if (cond) a else b` to a function invocation which has an effect on that argument? Don't we just basically drop the cond? That seems wrong to me.

Here is an example:

```scala
object EffectOnITE {

  case class A(var x: BigInt)
  def reset(a: A): Unit = { a.x = 0 }

  def f(a1: A, a2: A): Unit = {
    reset(if (true) a1 else a2)
    assert(a1.x == 0)
    assert(a2.x == 0) // should not go through
  }

}
```

vkuncak:
Not sure what #1121 did, but this example still passes even if the assertion fails in REPL, as expected.
```scala
object EffectOnITE {
  case class A(var x: BigInt)
  def reset(a: A): Unit = { a.x = 0 }

  def f(a1: A, a2: A): Unit = {
    reset(if (true) a1 else a2)
    assert(a1.x == 0)
    assert(a2.x == 0) // should not go through
  }
  def test ={
    f(A(1), A(2))
  }
}
```

vkuncak:
Maybe related to https://github.com/epfl-lara/stainless/issues/916

mario-bucev:
It also ~~effects~~ affects match expression:
```scala
object EffectOnMatchExpr {

  case class A(var x: BigInt)
  def reset(a: A): Unit = { a.x = 0 }

  def f(a1: A, a2: A): Unit = {
    reset { true match {
      case true => a1
      case _ => a2
    }}
    assert(a1.x == 0)
    assert(a2.x == 0) // should not go through
  }

}
```

Issue #1099, opened 2021-06-10
AntiAliasing transformation combining aliasing and effect postponing


jad-hamza:
In this example, `AntiAliasing` goes through the `val c = C(a2)` using the branch that checks that variables in the bound expressions are distinct from the variables after the `val`:
https://github.com/epfl-lara/stainless/blob/dff1b643830fdbad1ee5ff8d1cc8ad7f7cc15e05/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L351
But here the check isn't strict enough because `a2` aliases `a1` and `a1` is used.

```scala
object SimpleAliasing2 {
  case class A(var x: Int)
  case class C(a: A)
  def f(a1: A) = {
    require(a1.x == 0)
    val a2 = a1
    val c = C(a2)
    c.a.x += 1
    assert(a1.x == 0) // assertion goes through while it shouldn't
  }
}
```

Issue #916, opened 2021-02-24
Badly translated nested conditions in the AntiAliasing phase


brunnerant:
Nested conditions and assignments are incorrectly tracked by the `AntiAliasing` phase, resulting in an incorrect translation. The following example illustrates the bug:
```Scala
object Bug { 
  case class A(var a: BigInt)

  def bug(b1: Boolean, b2: Boolean) = {
      val a = A(1)
      val b = A(2)
      val c = A(3)
      val x = if (b1) (if (b2) a else b) else c
      x.a = 4
  }
}
```
This snippet is translated to (after `AntiAliasing`):
```Scala
def bug((b1: Boolean) @pure, (b2: Boolean) @pure): Unit = {
  var a: A = A(1)
  var b: A = A(2)
  var c: A = A(3)
  val x: A = if (b1) {
    if (b2) {
      a
    } else {
      b
    }
  } else {
    c
  }
  a = @unchecked (if (b1) {
    A(4)
  } else {
    a
  }).asInstanceOf[A]
  c = @unchecked (if (¬b1) {
    A(4)
  } else {
    c
  }).asInstanceOf[A]
  b = @unchecked (if (b1) {
    A(4)
  } else {
    b
  }).asInstanceOf[A]
  ()
}
```
As can be seen, the branching on the boolean `b2` is not part of the translated code, which is obviously wrong.

vkuncak:
I am now getting 
```
java.lang.ClassCastException: inox.ast.Expressions$IfExpr cannot be cast to inox.ast.Expressions$Variable
	at stainless.extraction.imperative.TreeDeconstructor.$anonfun$deconstruct$6(Trees.scala:337)
	at inox.ast.Deconstructors$Operator$.$anonfun$unapply$1(Deconstructors.scala:569)
	at stainless.extraction.imperative.AntiAliasing$transformer$1$.transform(AntiAliasing.scala:533)
	at stainless.extraction.imperative.AntiAliasing$transformer$1$.$anonfun$transform$26(AntiAliasing.scala:454)```

PR #1245, opened 2022-03-14
Addressing some issues in AntiAliasing

mario-bucev:
Hopefully fixes the underlying problems of #916, #1099, and #1122 without introducing new bugs...


core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -356,6 +364,11 @@ class AntiAliasing(override val s: Trees)(override val t: s.type)(using override
           case v: Variable if env.rewritings.contains(v.toVal) =>
             freshenLocals(env.rewritings(v.toVal))
 
+          case Assignment(v, expr) =>
+            // We explicitly override this case to not have the recursion go through `v` and potentially taking the above case
+            // (otherwise, we could get something like `env.rewritings(v.toVal) = expr` which may not be well-defined).
+            Assignment(v, transform(expr, env))
```
samarion, line 370: If `v` belongs to `env.rewritings`, doesn't that mean that we've added a `LetVar` to the rewritings map? That would be incorrect in itself, so if this case is cropping up we might have an underlying issue.

mario-bucev: It appears that the only `v`s for which this case is problematic are those that are of a mutable type and `val`-defined (such as the `val x = if (b1) a1 else a2` in [`invalid/i916a1`](https://github.com/epfl-lara/stainless/blob/e91c5d6fb8c86ab7631130ccd86c6daca8cb2b74/frontends/benchmarks/imperative/invalid/i916a1.scala)).
Such rewriting in `env` is added by: 
https://github.com/epfl-lara/stainless/blob/e91c5d6fb8c86ab7631130ccd86c6daca8cb2b74/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L417-L434 
and seems to be intended.

samarion: But when are we seeing an `Assignment(x, ...)`? Scala vals can't be assigned, are we applying a transformation twice?

mario-bucev: `val x` is transformed into a `var x` and it is re-assigned each time we modify one of its component.
If we e.g. see `x.a = 123`, it is transformed into an assignment: `x = A(123)`

samarion: Ah, I see. Thanks for bearing with me :)

I don't like the fact that we're applying `transform` to assignments produced during the phase since it makes it difficult to distinguish issues in the phase from issues in the input program. I think it would be worth cleaning up these cases to avoid applying `transform` to the generated `Assignment` expressions. For example,

https://github.com/mario-bucev/stainless/blob/667ef3da0422c9306a96afc760e7cd6bbaf17c32/core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala#L501-L502

should probably be something like below instead

```scala
val applied = updatedTarget(target, v)
Assignment(target.receiver, transform(applied, env)).copiedFrom(up)
```

PR #1246, opened 2022-03-15
Fix #1159

### LINKED ISSUE #1159: ApplyLetRec arguments are not properly mapped in postconditions

mario-bucev:
Following the suggestion https://github.com/epfl-lara/stainless/pull/1160#issuecomment-918938254

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/ImperativeCodeElimination.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/ImperativeCodeElimination.scala
@@ -242,46 +242,75 @@ class ImperativeCodeElimination(override val s: Trees)(override val t: s.type)
 
               if (modifiedVars.isEmpty) fdWithoutSideEffects else {
                 val freshVars: Seq[Variable] = modifiedVars.map(_.freshen)
-
                 val newParams: Seq[ValDef] = inner.params ++ freshVars.map(_.toVal)
-                val freshVarDecls: Seq[Variable] = freshVars.map(_.freshen)
 
-                val rewritingMap: Map[Variable, Variable] = modifiedVars.zip(freshVarDecls).toMap
-                val freshBody = postMap {
-                  case Assignment(v, e) => rewritingMap.get(v).map(nv => Assignment(nv, e))
-                  case v: Variable => rewritingMap.get(v)
-                  case _ => None
-                } (bd)
+                def recHelper(body: Expr,
+                              // This map is used by the postcondition rewriting to replace the result variable
+                              // of the ensuring clause to its freshened couter-part
+                              extraVarReplace: Map[Variable, Expr],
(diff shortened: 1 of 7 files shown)


core/src/main/scala/stainless/extraction/imperative/ImperativeCodeElimination.scala
```diff
@@ -242,46 +242,72 @@ class ImperativeCodeElimination(override val s: Trees)(override val t: s.type)
...
+                             (bodyWrapper: (Expr, Seq[Variable]) => Expr): (Expr, Expr => Expr, Seq[Variable]) = {
+                  assert(extraVarReplace.keySet.intersect(modifiedVars.toSet).isEmpty)
+                  val freshVarDecls: Seq[Variable] = modifiedVars.map(_.freshen)
+                  val rewritingMap: Map[Variable, Variable] = modifiedVars.zip(freshVarDecls).toMap
+
+                  val freshBody = postMap {
+                    case Assignment(v, e) => rewritingMap.get(v).map(nv => Assignment(nv, e))
+                    case v: Variable => rewritingMap.get(v).orElse(extraVarReplace.get(v))
+                    case Old(v: Variable) if freshVarDecls.contains(v) =>
+                      Some(freshVars(freshVarDecls.indexOf(v)))
```
samarion, line 260: Ideally, this should be more restrictive and only apply within the postcondition to allow the ImperativeCleanup phase to correctly catch all incorrect uses of `old`. Could you maybe make this part of the `extraVarReplace` (_e.g._ by making it into a `Map[Expr, Expr]`)?

samarion (approved):

core/src/main/scala/stainless/extraction/imperative/ImperativeCodeElimination.scala
mario-bucev: Good point!

mario-bucev: (I'll push an updated version)

mario-bucev: Actually, the `Old(v)` expression will already have its variable rewritten to its freshened variant (contained in `freshVarDecls`). I pushed a version that is simply passed a flag to tell us whether we should rewrite `old` expressions (less elegant, that's for sure...).

vkuncak:
@samarion again it's marked "approved". Did you mean to just comment without clicking "approve"?

PR #1247, opened 2022-03-15
VC checking in parallel

mario-bucev:
The VCs were actually not checked in parallel due to the use of `Future.successful` (eager) instead of `Future.apply` (maybe on purpose?).
It also enables parallelism back for macOS (as it seems the problem was caused by https://github.com/Z3Prover/z3/issues/1080, which is now addressed)

### DIFF
--- a/core/src/main/scala/stainless/verification/VerificationChecker.scala
+++ b/core/src/main/scala/stainless/verification/VerificationChecker.scala
@@ -6,6 +6,7 @@ package verification
 import inox.Options
 import inox.solvers._
 
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger}
 import scala.util.{ Success, Failure }
 import scala.concurrent.Future
 import scala.collection.mutable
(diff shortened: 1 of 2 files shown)


core/src/main/scala/stainless/verification/VerificationChecker.scala
```diff
@@ -92,43 +93,51 @@ trait VerificationChecker { self =>
...
-            }
-          }
+    def processVC(vc: VC): Option[(VC, VCResult)] = {
+      if (stop) None else {
+        val simplifiedCondition = simplifyExpr(
+          simplifyLets(removeAssertions(vc.condition))
+        )(using PurityOptions.assumeChecked)
+
+        // For some reasons, the synthesized copy method lacks default parameters...
+        val simplifiedVC = (vc.copy()(condition = simplifiedCondition, fid = vc.fid, kind = vc.kind, satisfiability = vc.satisfiability): VC).setPos(vc)
```
samarion, line 109: Is the copy method default arguments issue still there in the latest Scala versions?

core/src/main/scala/stainless/verification/VerificationChecker.scala
```diff
@@ -350,7 +359,8 @@ trait VerificationChecker { self =>
 
 object VerificationChecker {
   // number of verified VCs (incremented when a VC is verified)
-  var verified: Int = 0
+  // it may be accessed concurrently, and is hence defined as an atomic integer.
+  val verified: AtomicInteger = new AtomicInteger(0)
   // total number of VCs (we add to that counter when entering `checkVCs`)
   // this is cumulative across different subprograms (for `SplitCallBack`)
   var total: Int = 0
```
samarion, line 366: `total` and `startedVerification` are probably also concurrent because the `SplitCallback` runs verification of sub-programs in parallel.

core/src/main/scala/stainless/verification/VerificationChecker.scala
```diff
@@ -92,43 +93,51 @@ trait VerificationChecker { self =>
...
+
+        Some(vc -> res)
       }
+    }
+
+    val results = Future.traverse(vcs) { vc =>
+      // Note that `successful(e)` is eager and gets immediately evaluated whereas Future(e) is a scheduled task.
+      // If parallelism is not explicitly enabled, we fallback to eager evaluation
+      // (even on a thread pool of a single thread, it seems that the VCs won't necessarily be sequentially processed).
+      if (nParallel.exists(_ > 1)) Future(processVC(vc))
```
samarion, line 139: Why do we fallback to no parallelism when it's not explicitly enabled? It seems strange to have a different behavior here to the default.

samarion (approved):
Great catch! I wonder how we never realized checking wasn't parallelized...

Do you see a performance improvement on any benchmarks with this change?

core/src/main/scala/stainless/verification/VerificationChecker.scala
mario-bucev: Sadly yes

core/src/main/scala/stainless/verification/VerificationChecker.scala
mario-bucev: Ah that's right! I'll make them atomic as well

core/src/main/scala/stainless/verification/VerificationChecker.scala
mario-bucev: I have remarked that the default configuration creates a thread pool with the number of logical cores; in my experience, I got better results (less spurious timeouts) by explicitly setting the `-Dparallel` to the number of physical cores, and only when using a single solver (the portfolio solver seems to run all of its underlying solvers in parallel, causing competition for CPU)

mario-bucev:
Yes, especially on benchmarks where we can get away by using a single solver :)

vkuncak:
@samarion : it's marked as approved, but do you want to have a quick look at the final changes?

Issue #1068, opened 2021-05-21
Effects in arguments of functions


jad-hamza:
This example should go through but doesn't:

```scala
object ArgumentsEffects {

  def f(x1: Int, x2: Int, x3: Int): Unit = {
    require(x1 == 1)
    require(x2 == 1)
    require(x3 == 3)
  }

  def test: Unit = {
    var x = 0
    f({x += 1; x}, x, {x+= 2; x })
  }

}
```

```scala
// Before ImperativeCodeElimination
[ Debug  ] def test$0: Unit = {
[ Debug  ]   var (x$220: Int) @var = 0
[ Debug  ]   f$75({
[ Debug  ]     x$220 = (x$220 + 1)
[ Debug  ]     x$220
[ Debug  ]   }, x$220, {
[ Debug  ]     x$220 = (x$220 + 2)
[ Debug  ]     x$220
[ Debug  ]   })
[ Debug  ] }

// After ImperativeCodeElimination
[ Debug  ] def test$0: Unit = {
[ Debug  ]   val (x$221: Int) @var = 0
[ Debug  ]   val (x$223: Int) @var = x$221 + 1
[ Debug  ]   val tmp$15: Unit = ()
[ Debug  ]   val (x$222: Int) @var = x$223 + 2
[ Debug  ]   val tmp$14: Unit = ()
[ Debug  ]   f$75(x$223, x$222, x$222)
[ Debug  ] }
```

vkuncak:
Note that this works.
```scala
  def test: Unit = {
    var x = 0
    val xx1 = {x += 1; x}
    val xx2 = x
    val xx3 = {x+= 2; x }
    f(xx1, xx2, xx3)
  }
```
I wonder if we should not just first introduce variables for block arguments and then transform?

samarion:
Yes, great point. I was making a similar reflection in the review of https://github.com/epfl-lara/stainless/pull/1252 because this issue is more general than function arguments.

PR #1252, opened 2022-03-21
Round 2 of AntiAliasing changes

mario-bucev:
(note: without mutation, we couldn't create a cycle with https://github.com/epfl-lara/bolts/pull/44 (without resorting to lazy evaluation) :p)


mario-bucev:
Converting this to a draft due to some inefficiency introduced by #1245.
For instance, given:
```scala
import stainless._
import stainless.lang._

object i1251b {

  case class Box(var length: Int)

  def mutate(b: Box): Unit = { b.length = 123; }

  def h2(b1: Box, b2: Box, cond: Boolean): Unit = {
    val x = b1.length
    val c = if (cond) b1 else b2
    val d = c
    mutate(d)
    assert(b1.length == x)
  }
}
```

we generate:
```scala
val x: Int = b1.length
var c: Box = (if (cond) {
  b1
} else {
  b2
})
var d: Box = (if (cond) {
  b1
} else if (!cond) {
  b2
} else {
  c
})
var res: (Unit, Box) = mutate(@DropVCs (if (cond) {
  b1
} else if (!cond && !cond) {
  b2
} else if (!cond && !!cond) {
  if (cond) {
    b1
  } else if (!cond) {
    b2
  } else {
    c
  }
} else if (cond) {
  b1
} else if (!cond && !cond) {
  b2
} else if (!cond && !!cond) {
  c
} else {
  d
}))
(if (cond) {
  b1 = @DropVCs Box(@DropVCs res._2.length).asInstanceOf[Box]
} else {
  ()
})
(if (!cond && (!cond && !cond)) {
  b2 = @DropVCs Box(@DropVCs res._2.length).asInstanceOf[Box]
} else {
  ()
})
(if (!cond && (!(!cond && !cond) && (!cond && !!cond))) {
  c = @DropVCs Box(@DropVCs res._2.length).asInstanceOf[Box]
} else {
  ()
})
(if (!cond && (!(!cond && !cond) && !(!cond && !!cond))) {
  d = @DropVCs Box(@DropVCs res._2.length).asInstanceOf[Box]
} else {
  ()
})
res._1
assert(b1.length == x)
()
```
due to duplicating every occurrence of `c` and `d` with `if (cond) b1 else b2`. 
Even if we were a bit more careful in substituting, we would still end up with expression duplication.
So maybe we should rethink `env.rewritings`?

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -490,9 +496,11 @@ class AntiAliasing(override val s: Trees)(override val t: s.type)(using override
             if (targets.exists(target => !env.bindings.contains(target.receiver.toVal)))
               throw MalformedStainlessCode(up, "Unsupported form of array update")
 
+            val recV = transform(v, env)
+
             Block(targets map { target =>
-              val applied = updatedTarget(target, v)
-              transform(Assignment(target.receiver, applied).copiedFrom(up), env)
+              val applied = updatedTarget(target, recV)
+              Assignment(target.receiver, applied).copiedFrom(up)
```
samarion, line 503: I think this transformation (and a few others) might be incorrectly dropping stuff from the input array expression. If there was a `Block` expression in `a`, then I believe it would be silently dropped in the output here.

I'm not sure what the cleanest approach to fix this is, though. I guess some options are:
1. adding a method which extracts the non-target part of an expression and join it into the output,
2. we could always include the expression in the output and hope simplification cleans it up later on by having the result be something like
    ```scala
    Block(transform(a, env), Assignement(...))
    ```
3. we could have a pre-transformation which makes sure to bring all blocks/targets into a normal form which is compatible with target assignment. This option would also make sure that mutating calls inside targets are handled correctly.

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -608,24 +620,38 @@ class AntiAliasing(override val s: Trees)(override val t: s.type)(using override
...
+            val rewrElems = rewriteAndCheckArgs(fa, elems)
+            FiniteArray(rewrElems.map(transform(_, env)), tpe).copiedFrom(fa)
+
+          case au@ArrayUpdated(arr, i, arg) =>
+            val Seq(rewrArr, rewrI, rewrArg) = rewriteAndCheckArgs(au, Seq(arr, i, arg))
+            ArrayUpdated(transform(rewrArr, env), transform(rewrI, env), transform(rewrArg, env)).copiedFrom(au)
+
+          case mu@MapUpdated(map, k, v) =>
+            val Seq(rewrMap, rewrK, rewrV) = rewriteAndCheckArgs(mu, Seq(map, k, v))
+            MapUpdated(transform(rewrMap, env), transform(rewrK, env), transform(rewrV, env)).copiedFrom(mu)
```
samarion, line 645: I think these are only relevant when the array and map base types are mutable, no?

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -572,9 +573,16 @@ trait EffectsAnalyzer extends oo.CachingPhase {
...
     case _: LargeArray | _: ArrayUpdated if kind == ReplacementKind && !path.head.isInstanceOf[ArrayAccessor] && path.head != UnknownArrayAccessor =>
       Seq.empty
-    // TODO: These two cases are incorrect, but removing them breaks existing codebase...
-    case _: MutableMapUpdated => Seq.empty
-    case _: ArrayUpdated => Seq.empty
+    case au: ArrayUpdated => au.getType match {
+      case ArrayType(base) if !symbols.isMutableType(base) => Seq.empty
+      case _ =>
+        throw MalformedStainlessCode(expr, s"Couldn't compute effect targets of array copy update ${au.asString}")
+    }
```
samarion, line 580: This (and the corresponding `MapUpdated` logic) seems to strict. This will also prevent users from using the `snapshot(a).updated(i, v)` construct which we were discussing.

samarion (commented):
About duplicating expressions, we're already doing this a lot in the imperative phase (e.g. during effect copying after function calls), so it's not necessarily a deal breaker.

You could try introducing a some additional simplifications in the `ImperativeCleanup` phase to deal with these. Maybe using `simplifyLets` and/or `simplifyExpr` will give good results.

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -384,12 +390,12 @@ class AntiAliasing(override val s: Trees)(override val t: s.type)(using override
             val updates1 =
               targets1 map { target =>
                 val applied = updatedTarget(target, ArraySelect(array2, index2).setPos(swap))
-                transform(Assignment(target.receiver, applied).setPos(swap), env)
+                Assignment(target.receiver, applied).setPos(swap)
```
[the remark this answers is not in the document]
mario-bucev: Good point, I've totally missed that out! I'll also add tests for these cases.

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: Option 3 seems to also benefit https://github.com/epfl-lara/stainless/issues/1068

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
mario-bucev: That's true, for instance, the [`SortedArray` benchmark needs to be updated](https://github.com/epfl-lara/bolts/pull/44) to account for that restriction (I'm realizing that we can also return `Seq.empty` if the effect kind is replacement).
For `ModfiyingKind` though, I do not think there is a way to express the concerned targets.

mario-bucev:
It remains to comment on parts of the code, I'll notify you once (if ever...) I'm done :)

mario-bucev:
@samarion Ready for re-review! (when & if you have the time and desire to do so, of course :) )

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -564,49 +578,62 @@ trait EffectsAnalyzer extends oo.CachingPhase {
...
 
-    case Let(vd, e, b) if !symbols.isMutableType(vd.tpe) =>
-      getTargets(b, kind, path).map(_.bind(vd, e))
-
     case Let(vd, e, b) =>
-      getTargets(b, kind, path).map(_.bind(vd, e)).flatMap { be =>
-        if (be.receiver == vd.toVariable) getTargets(e, kind, be.path.path)
-        else Seq(be)
+      val targs0 = getTargets(b, kind, path)
+      // TODO: Explain why
```
samarion, line 627: Please do :)

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -717,33 +756,33 @@ trait EffectsAnalyzer extends oo.CachingPhase {
...
-    def inEnv(effect: Effect, env: Map[Variable, Effect]): Option[Effect] =
-      env.get(effect.receiver).map(e => Effect(effect.kind, e.receiver, e.path ++ effect.path))
+    def inEnv(effect: Effect, env: Map[Variable, Set[Effect]]): Set[Effect] =
+      env.getOrElse(effect.receiver, Set.empty).map(e => Effect(effect.kind, e.receiver, e.path ++ effect.path))
 
-    def effect(expr: Expr, env: Map[Variable, Effect]): Seq[Effect] =
+    def effect(expr: Expr, env: Map[Variable, Set[Effect]]): Set[Effect] =
       getAllTargets(expr) flatMap { (target: Target) =>
-        inEnv(target.toEffect(ModifyingKind), env)
+        inEnv(target.toEffect(ModifyingKind), env).toSet
```
samarion, line 764: Nit: you don't need the call to `toSet` here, right?

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
```diff
@@ -782,7 +821,7 @@ trait EffectsAnalyzer extends oo.CachingPhase {
           .filter(p => effects contains p._2)
           .flatMap(_._1)
 
-      case Assignment(v, value) => rec(value, env) ++ env.get(v)
+      case Assignment(v, value) => rec(value, env) ++ env.get(v).toSeq.flatten
```
samarion, line 824: Nit: why `toSeq`?

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -212,58 +212,79 @@ class AntiAliasing(override val s: Trees)(override val t: s.type)(using override
...
+          val assignment = if (isMutableType(target.receiver.getType)) {
+            val newValue = updatedTarget(target, wrap(result))
+            val adt = target.receiver.getType
             val castedValue =
               if (adt.isInstanceOf[ADTType] && adt.asInstanceOf[ADTType].getSort.constructors.size == 1)
                 newValue
               else
-                Annotated(AsInstanceOf(newValue, effect.receiver.getType), Seq(DropVCs)).copiedFrom(newValue)
-            Assignment(effect.receiver, castedValue).setPos(pos)
+                wrap(AsInstanceOf(newValue, target.receiver.getType))
```
samarion, line 247: Should we have `AsInstanceOf(...).copiedFrom(newValue)` to propagate positions?

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -353,89 +366,200 @@ class AntiAliasing(override val s: Trees)(override val t: s.type)(using override
...
+        //   (as a.b.c is not a *proper* prefix of itself)
+        def aliasingValDefs(updTarget: Target, env: Env): Map[ValDef, Set[Target]] = {
+          env.targets.map {
+            case (vd, candidatePrefixes) =>
+              val properPrefixes = candidatePrefixes.filter(_.maybeProperPrefixOf(updTarget))
+              vd -> properPrefixes
+          }.filter(_._2.nonEmpty)
+        }
+
+        def aliasingValDefs2(updTargets: Set[Target], env: Env): Map[ValDef, Set[Target]] = {
```
samarion, line 434: Does overloading not work here? Adding the 2 suffix is more confusing IMO

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -353,89 +366,200 @@ class AntiAliasing(override val s: Trees)(override val t: s.type)(using override
...
+                  // If cond1 && ... && condn provably cover all possible cases, then we know that `newExpr` is
+                  // a precise alias to n existing variables, so there are no new targets to consider (case 2).
+                  // Otherwise, it may be the case that `newExpr` refer to fresh expression and alias existing variables (case 3),
+                  // such as in the following example:
+                  //   val y = if (cond1) x else Ref(123)
+                  // Here, the targets of `y` would be the targets of x, with condition cond1.
+                  // But it also conditionally introduce a new target, namely Ref(123) when !cond1.
+                  // In such case, we introduce `y` as a new target, with condition !cond1.
+                  assert(targs.forall(_.condition.isDefined))
+                  val negatedConds = not(orJoin(targs.map(_.condition.get).toSeq))
```
samarion, line 546: Does this really work in practice? IIRC, the `not` and `orJoin` simplifications are really trivial so you might not even catch the case where you have `cond` and `!cond` as conditions. Maybe you could introduce some dedicated logic since you have a strong expectation of what the conditions will look like.

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -445,80 +569,49 @@ class AntiAliasing(override val s: Trees)(override val t: s.type)(using override
...
+            val recI = transform(i, env)
+            val recV = transform(v, env)
+            val targets = getDirectTargetsDealiased(arr, ArrayAccessor(recI), env)
+              .getOrElse(throw MalformedStainlessCode(up, "Unsupported form of array updated"))
 
-          case up @ ArrayUpdate(a, i, v) =>
-            val ra = freshenLocals(replaceFromSymbols(env.rewritings, a))
-            val targets = getDirectTargets(ra, ArrayAccessor(i))
+            // Sanity check (the same as for Swap)
+            assert(Some(targets) == getDirectTargetsDealiased(transform(arr, env), ArrayAccessor(recI), env))
```
samarion, line 588: Isn't there a risk that target computation will fail on the transformed expression?

samarion (commented):
I've read through most of the change and left a few comments. I quite like the move away from rewritings with explicit target tracking instead.

I didn't check the normalization logic yet, but I'm surprised by some of the benchmarks which were disabled due to it (for example, the MutateInsideN benchmarks). Which problems are the block and let hoisting steps solving?

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
mario-bucev: Oops, thank you for catching that!

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: Ah yes, good point!

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: I had some issues with overloading IRC, but I can try adding a `DummyImplicit` to resolve the issue

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
```diff
@@ -353,89 +366,200 @@ class AntiAliasing(override val s: Trees)(override val t: s.type)(using override
...
+        }
+
+        // Group targets that have the same receiver and path under one target (by or'ing their condition)
+        def groupedTargets(targets: Set[Target]): Set[Target] = {
+          targets.groupBy(t => (t.receiver, t.path))
+            .map { case ((recv, path), tgs) =>
+              val conds = tgs.flatMap(_.condition)
+              val combinedCond = {
+                if (conds.isEmpty) None
+                else {
```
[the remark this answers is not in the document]
mario-bucev: Fair point!

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: You are right, it doesn't work at all in practice (it ends up generating quite a lot of extra update statements, though it's not incorrect). I'll add something that at least catches things such as `cond` and `!cond`.

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
mario-bucev: `getDirectTargetsDealiased` returns a `None` in case of failure (which would fire the assertion...)

core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
mario-bucev: I have reverted the `getTargets` returning a `Seq` to return a `Set` (as it was before), but I forgot to review the parts in `EffectsAnalyzer` that depend on this `Seq`/`Set` return type (like for the commented snippet below!)

mario-bucev:
Thank you for reviewing! For the normalization, the two main motivations are to sequence effects of function calls (and addresses issue such as #1068) and introduce `val` bindings for indices and keys.
As for hoisting let within let, I think this is a "by-product" of the normalization and not necessarily sought after (though the comment suggests it is... I'll change that).

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
samarion: Yes, I meant: won't this generate spurious failures? It's not clear to me that we expect `getDirectTargetsDealiased` to always work on transformed trees.

samarion:
Regarding block hoisting, could we maybe avoid flattening them? Blocks are somewhat useful in allowing local effects which is what the `MutateInsideN` benchmarks are testing.

core/src/main/scala/stainless/extraction/imperative/AntiAliasing.scala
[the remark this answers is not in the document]
mario-bucev: Actually, `orJoin` returns `BooleanLiteral(false)` on an empty sequence (but `andJoin` does return BooleanLiteral(true)`)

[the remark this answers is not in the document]
samarion: Oh right, good point.

Hmm, but then shouldn't we consider an empty condition as implying `BooleanLiteral(true)`? It seems fishy to me that we have special handling of the empty case. Maybe `conds` should rather be defined as
```scala
val conds = tgs.map(_.condition.getOrElse(BooleanLiteral(true))
```
(and then we can just use `orJoin` and avoid the special case for empty conditions).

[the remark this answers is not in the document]
mario-bucev: That's a more sensical thing to do indeed

Issue #1259, opened 2022-05-06
Avoid capturing unnecessary constrained variables in `FunctionClosure` phase


mario-bucev:
The following `inner` function unnecessarily captures `x` when lifted:
```scala
object Test {
  def outer(x: Int): Unit = {
    require(x >= 0)
    def inner(y: Int): Unit = ()
  }
}
```
is roughly transformed to:
```scala
def outer(x: Int): Unit = {
  require(x >= 0)
}

def inner(x: Int, y: Int): Unit = {
  require(x >= 0)
}
```
(due to `x` being constrained with the `require`).

A naive approach checking for occurrences is not enough though, in the following example, we should not only capture `a` but also `x` (due to constraining `a`): 
```scala
object Test {
  def outer(x: Int, a: Int): Unit = {
    require(x >= 0 && a >= x)
    def inner(y: Int): Unit = {
      val b = a // Captures a, but the hoisting of inner must also capture x for the PC a >= x (and x >= 0 as well)
    }
  }
}
```

samarion:
We might drop cases where we have `require(false)` with this change. This could be relevant for potential transformations which _e.g._ introduce case reasoning. I don't think we actually have any currently, just something to think about.

What is your use case for such a change?

mario-bucev:
Sorry, I forgot to justify!
In the last Friday's meeting, we discussed the equivalence checker, and Dragana pointed out that it may fail on some programs containing simple `while` loops. She reduced the issue down to the hoisting of the transformed `while`-functions unnecessarily capturing some variables.
If two `while`-functions are equivalent but happen to capture different, unrelated variables, the EC rejects the program due to these functions not having the same signature.
We have discussed possible solutions, and have come to open this ticket :)

(I am unsure about whether the EC actually rejects or timeouts, but it does not output an expected result, btw @drganam feel free to correct me if I'm wrong on something)

samarion:
I see, then it makes sense to me.

I'm not sure I understand your comment about the naive approach. Isn't it clear that `a` is free in the body of `inner` and should therefore be captured?

mario-bucev:
I've updated the comment, it was not clear, sorry! What I meant is that we should also capture variables appearing in the PCs for variables appearing in the body of inner functions (which is obvious, but can be overlooked).

PR #1265, opened 2022-05-19
Avoid unnecessary capture of PC variables when hoisting functions

mario-bucev:
Addresses #1259
Edit: marked as a draft as I've found some bugs (that seem to be related to not taking into account function specifications)
Edit 2: the bug I've encountered has nothing to do with this PR... so I'm going to re-ready it (I'll try to fix the big in a separate PR).

### DIFF
--- a/core/src/main/scala/stainless/extraction/innerfuns/FunctionClosure.scala
+++ b/core/src/main/scala/stainless/extraction/innerfuns/FunctionClosure.scala
@@ -115,7 +115,7 @@ class FunctionClosure(override val s: Trees, override  val t: ast.Trees)
         }
       }
 
-      val nestedWithPaths = (for((fds, path) <- nestedWithPathsFull; fd <- fds) yield (fd, path)).toMap
+      val nestedWithPaths: Map[LocalFunDef, Path] = (for((fds, path) <- nestedWithPathsFull; fd <- fds) yield (fd, path)).toMap
       val nestedFuns = nestedWithPaths.keys.toSeq
       val nestedFunsIds = nestedFuns.map(_.id).toSet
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/stainless/extraction/innerfuns/FunctionClosure.scala
```diff
@@ -128,31 +128,86 @@ class FunctionClosure(override val s: Trees, override  val t: ast.Trees)
...
+              case Path.CloseBound(vd, e) if fdsFreeVars.contains(vd.toVariable) =>
+                exprOps.variablesOf(e)
+              case Path.CloseBound(vd, e) if exprOps.variablesOf(e).intersect(fdsFreeVars).nonEmpty =>
+                // `vd` may occur in a constraint somewhere, which can also transitively constraint the FVs
+                Set(vd.toVariable)
+              case Path.Condition(cond) =>
+                val varsCond = exprOps.variablesOf(cond)
+                if (varsCond.intersect(fdsFreeVars).nonEmpty) varsCond
+                else Set.empty[Variable]
+              case _ => Set.empty[Variable]
```
samarion, line 174: Nit: what about dependent types? Shouldn't we also include their free variables?

This might also affect the `CloseBound` cases.

core/src/main/scala/stainless/extraction/innerfuns/FunctionClosure.scala
```diff
@@ -128,31 +128,86 @@ class FunctionClosure(override val s: Trees, override  val t: ast.Trees)
...
+            // -inner2: `y` appears free, so it will be captured as well. `y` also appears
+            // in a PC condition x <= y. We could drop this condition, but we chose here to capture `x`
+            // as well in order to have 0 <= x && x <= y available in the hoisted version of `inner2`.
+            // (a more sophisticated approach would be to eliminate `x` entirely and have 0 <= y instead).
+            val path = nestedWithPaths(fd)
+            val picked = path.elements.toSet.flatMap {
+              case Path.CloseBound(vd, e) if fdsFreeVars.contains(vd.toVariable) =>
+                exprOps.variablesOf(e)
+              case Path.CloseBound(vd, e) if exprOps.variablesOf(e).intersect(fdsFreeVars).nonEmpty =>
+                // `vd` may occur in a constraint somewhere, which can also transitively constraint the FVs
```
samarion, line 168: Nit: the second 'constraint' should be 'constrain'

samarion (approved):

mario-bucev:
Thanks for the review! I'll address the remarks, in the meantime, I mark the PR as a draft to avoid any accidental merge...

core/src/main/scala/stainless/extraction/innerfuns/FunctionClosure.scala
mario-bucev: Doesn't `variablesOf` also take into consideration the FVs occurring in types? Having a look at it, it calls `typeOps.variablesOf`, but maybe I've missed something.

samarion: Yes, it does. But you need to actually call it on `vd.tpe` to get the free variables, _e.g._
```scala
case Path.OpenBound(vd) => exprOps.variablesOf(vd.tpe)
```

mario-bucev: I see, thanks for the clarification!

mario-bucev: Hopefully the pushed changes cover the dependent types. The `fdsFreeVars` now includes the FVs of types (if any). 
For `CloseBound` (1st case), we do not need to take the FVs of `vd.tpe` as it's already in `fdsFreeVars`.
For the 2nd case, the FVs of `vd.tpe` will be picked up by the next iteration of `step`.
Otherwise, the `vd` is not relevant for this inner function and is dropped.

core/src/main/scala/stainless/extraction/innerfuns/FunctionClosure.scala
```diff
@@ -163,9 +165,11 @@ class FunctionClosure(override val s: Trees, override  val t: ast.Trees)
             val path = nestedWithPaths(fd)
             val picked = path.elements.toSet.flatMap {
               case Path.CloseBound(vd, e) if fdsFreeVars.contains(vd.toVariable) =>
+                // Take all FVs of `e` so that we can "reconstruct" `vd` from them in `closeFd`
+                // (note that this `vd` is already picked because it appears in `fdsFreeVars`).
                 exprOps.variablesOf(e)
               case Path.CloseBound(vd, e) if exprOps.variablesOf(e).intersect(fdsFreeVars).nonEmpty =>
-                // `vd` may occur in a constraint somewhere, which can also transitively constraint the FVs
+                // `vd` may occur in a constrain somewhere, which can also transitively constraint the FVs
```
samarion, line 172: Oups, I meant the other 'constraint' :) the sentence should be "`vd` may occur in a constraint somewhere, which can also transitively constrain the FVs"

mario-bucev: Haha my bad!

PR #1294, opened 2022-09-16
Improve equivalence checking

drganam:
Improvements to equivalence checking phase, including: function call matching, conditional equivalence (--norm option), and user-provided test generators (mkTest annotation)

### DIFF
--- a/core/src/main/scala/stainless/extraction/trace/Trace.scala
+++ b/core/src/main/scala/stainless/extraction/trace/Trace.scala
@@ -19,9 +19,30 @@ class Trace(override val s: Trees, override val t: termination.Trees)
   private[this] class Identity(override val s: self.s.type, override val t: self.t.type) extends transformers.ConcreteTreeTransformer(s, t)
   private[this] val identity = new Identity(self.s, self.t)
 
+  private def evaluate(syms: s.Symbols, expr: Expr) = {
+    type ProgramType = inox.Program{val trees: self.s.type; val symbols: syms.type}
+    val prog: ProgramType = inox.Program(self.s)(syms)
+    val sem = new inox.Semantics {
+      val trees: self.s.type = self.s
+      val symbols: syms.type = syms
+      val program: prog.type = prog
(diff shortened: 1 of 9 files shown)


vkuncak:
@mario-bucev @samarion : Any chance you could take a quick look? Equivalence checking is already in master; this PR improves it. Most changes are localized in that phase and I we do not necessarily need a detailed review of that one; I suggest you focus on any possible interference (my quick check indicates this is minimal).

core/src/main/scala/stainless/extraction/trace/Trace.scala
```diff
@@ -37,58 +58,270 @@ class Trace(override val s: Trees, override val t: termination.Trees)
...
+        val f2Calls = getFunCalls(fd2).filter(!_.flags.exists(_.name == "library"))
+
+        val pairs = f1Calls zip f1Calls.map(m => f2Calls.find(f => m != f && checkArgs(m, f) && f.id.name == m.id.name).orElse(f2Calls.find(f => m != f && checkArgs(m, f))))
+
+        val validpairs = pairs.filter(elem => elem._2 != None)
+
+        validpairs.map(elem => (elem._1, elem._2) match {
+          case (m, Some(f)) =>
+            (equivalenceCheck(m, f, true), m, f)
+        })
```
mario-bucev, line 222: Minor suggestion
```suggestion
        pairs.collect {
          case (m, Some(f)) =>
            (equivalenceCheck(m, f, true), m, f)
        }
```

mario-bucev (commented):
Looks good to me :)

core/src/main/scala/stainless/extraction/trace/Trace.scala
drganam: Updated and squashed, thanks!

mario-bucev:
The CI is not happy, let us spank it a bit

drganam:
Now it's back to happy

mario-bucev:
@vkuncak Can we merge?

PR #1311, opened 2022-10-12
Equivalence checking documentation + examples

drganam:


vkuncak:
I will proceed to accept this with tests being there as showcase. Soon we should active some subset of them to run as part of test suite.

PR #1315, opened 2022-10-25
Experimental integration of OL- and OCBSL- based simplifiers

mario-bucev:
Note: the default simplifier remain the current one

### DIFF
--- a/core/src/main/scala/stainless/verification/VerificationChecker.scala
+++ b/core/src/main/scala/stainless/verification/VerificationChecker.scala
@@ -5,18 +5,49 @@ package verification
 
 import inox.Options
 import inox.solvers._
+import stainless.transformers.LatticesSimplifier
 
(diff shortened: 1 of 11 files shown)


core/src/main/scala/stainless/verification/VerificationChecker.scala
```diff
@@ -5,19 +5,50 @@ package verification
...
+  case Vanilla
+  case OCBSL
+  case OL
+  case Bland
+}
+object SimplifierKind {
+  def tryFromString(s: String): Option[SimplifierKind] = s.toLowerCase match {
+    case "vanilla" => Some(Vanilla)
+    case "ocbsl" => Some(OCBSL)
+    case "ol" | "olympique lyonnais" => Some(OL)
```
vkuncak, line 41: Always good to have insightful long option names.

PR #1324, opened 2022-11-07
Fix #1321

### LINKED ISSUE #1321: Error in example with covariant list
This crashes with both smt-z3 and smt-cvc4.
```scala
object ScalaList:

  sealed abstract class List[+A]
  case object Nil extends List[Nothing]
  case class Cons[+A](head: A, tail: List[A]) extends List[A]

  extension[A,B](xs: List[A]) def map(f: A => B): List[B] =
    xs match 
      case Nil => Nil
      case Cons(a, as) => Cons(f(a), as.map(f))

  trait MyInt:
    def x: BigInt

  final case class PosInt(v: BigInt) extends MyInt:
    require(0 <= v)
    def x = v
  end PosInt

  def applyTwice[A](xs: List[A], f: A => A): List[A] =
    xs.map(f).map(f)

  def getPos(n: BigInt): List[PosInt] =
    require(0 <= n)
    if n <= 0 then Nil
    else Cons(PosInt(n), getPos(n-1))

  def onlyPos(x: PosInt): PosInt =
    PosInt(x.v + 1)

  def work(n: BigInt) =
    require(0 <= n) 
    applyTwice[PosInt](getPos(n), onlyPos)

end ScalaList
```

mario-bucev:
#1321 root cause is due to having the lambda parameter `x` be typed differently.
In the body of the lambda, `x` has a plain type (non-refined). The lambda function type, however, has parameter type refined. The fix is to replace all occurrences of `x: MySimpleType` with `x: MyRefinedType`.

### DIFF
--- a/core/src/main/scala/stainless/verification/TypeChecker.scala
+++ b/core/src/main/scala/stainless/verification/TypeChecker.scala
@@ -1069,8 +1069,13 @@ class TypeChecker(val program: StainlessProgram, val context: inox.Context, val 
         checkType(tc.bindWithValues(from, es.init), es.last, to)
 
       case (Lambda(params1, body), FunctionType(params2, to)) =>
-        val vds = params1.zip(params2).map { case (vd, tp) => vd.copy(tpe = tp) }
-        checkType(tc.bind(vds), body, to)
+        val (vds, retyped) = params1.zip(params2).foldLeft((Seq.empty[ValDef], Map.empty[Variable, Type])) {
+          case ((accVds, accRetyped), (vd, tp)) =>
+            if (vd.tpe == tp) (accVds :+ vd, accRetyped)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/stainless/verification/TypeChecker.scala
```diff
@@ -1069,8 +1069,13 @@ class TypeChecker(val program: StainlessProgram, val context: inox.Context, val
         checkType(tc.bindWithValues(from, es.init), es.last, to)
 
       case (Lambda(params1, body), FunctionType(params2, to)) =>
-        val vds = params1.zip(params2).map { case (vd, tp) => vd.copy(tpe = tp) }
-        checkType(tc.bind(vds), body, to)
+        val (vds, retyped) = params1.zip(params2).foldLeft((Seq.empty[ValDef], Map.empty[Variable, Type])) {
+          case ((accVcs, accRetyped), (vd, tp)) =>
```
vkuncak, line 1073: accVDs

PR #1339, opened 2022-11-17
Update SBT Stainless plugin

### LINKED ISSUE #1312: Using sbt with stainless
@mario-bucev has more details on this, but following:

https://www.scala-sbt.org/1.x/docs/sbt-new-and-Templates.html

we should have a working epfl-lara/stainless.g8 repository for using stainless and sbt.

mario-bucev:
Closes #1312.
The plugin can now be used "offline", and we should soon create a new release with the SBT plugin attached.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 stainless.conf
 .stainless-external-tests/
 .stainless-package-standalone/
+.stainless-package-sbt-plugin/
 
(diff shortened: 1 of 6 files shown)


vkuncak (approved):
Looks good. I suppose we should make a release soon so the plugin artefact is there.

PR #1378, opened 2023-02-21
Equivalence checking as a component

mario-bucev:
Equivalence checking is now a component, activated with `--equivchk`. As before, it requires `--models` and `--comparefuns` to be fed with a list of model function(s) and candidate function(s) respectively, and optionally accept a `--norm=<fn name>` for normalization function. Note that `@traceInduct` functions need not to run with the `--equivchk` option.
New options:
- `--equivchk-n=<int>` (default: 3): specify the `N` hyperparameter
- `--equivchk-init-score=<int>` (default: 200): specify the initial score for models
- `--equivchk-max-perm=<int>` (default: 16): maximum number of matching for auxiliary functions
- `--silent-verification=<bool>` (default: `false`): do not print any message when a VC fails to verify (whether due to invalidity or timeout).

This PR: 
- :warning: **factors `unificationConstraint` from `oo.TypeOps` to `ast.TypeOps`. `oo.TypeOps` overrides `ast.TypeOps` for `oo` constructs.**
- splits `Trace` into `EquivalenceChecker` and `TraceInductElimination`.
- adds a test suite for equivalence checking, so the nightly CI will run for longer


mario-bucev (commented):
This is not stack-safe:
https://github.com/epfl-lara/stainless/pull/1378/files#diff-200037b858c6d7d9bcc7631625e33cdeba01cc605d80098960961b87d6806e8dR151-R189

vkuncak (approved):

PR #1387, opened 2023-03-09
Equivalence checking: apply max limit on permutation while pruning

mario-bucev:
Before this PR, the strategy was to generate the maximum number of allowed matching (default of 16) and then perform pruning on it. Some picked matching may get pruned, leaving us with less than the limit which is not something we would like. 
This PR changes the behaviour to pick matching (passing the pruning test) as long as we don't go over the limit.


mario-bucev:
`funnyarith3` for example is a good test case (which sometimes fails depending on how symbols are sorted, affecting the order of the matching as well).

PR #1402, opened 2023-04-25
Add covariant List and Option to the library

mario-bucev:
As requested by @vkuncak.
Removes the `CovariantList` benchmark since it is now in the library.


frontends/common/src/it/scala/stainless/LibrarySuite.scala
```diff
@@ -23,7 +23,7 @@ abstract class AbstractLibrarySuite(opts: Seq[inox.OptionValue[_]]) extends AnyF
   }
 
   protected def keep(tr: ast.Trees)(fd: tr.FunDef): Boolean = fd match {
-    case fd if fd.flags.exists(_.name == "unchecked") => false
+    case fd if fd.flags.exists(f => f.name == "unchecked" || f.name == "synthetic") => false
```
mario-bucev, line 26: Synthetic functions are filtered out before being passed to the type checker when using the usual filter. Since the `LibrarySuite` is a custom filter, removing synthetic option was also intended?

vkuncak: Maybe @samarion would have a clue.

mario-bucev: For context, the type checker says that `ObjectPrimitiveSize` does not have visibility on `Object` (since it is mutually recursive with it - probably due to the encoding of `::`).

samarion: No idea, sorry :) Maybe @jad-hamza ?

PR #1421, opened 2023-05-30
Add mapTr benchmark

drganam:
Adding an example from the FP course


mario-bucev:
Ah yes, the Scala 2 frontend is still here, I've forgot about it sorry!
The benchmark should be moved to `benchmarks/dotty-specific/valid`

PR #1434, opened 2023-08-21
Update faq.rst

drganam:
Java is **not** a good language for specifications?


mario-bucev:
That's a bold claim haha, maybe we should ask the original author (I think @vkuncak ?)

Issue #1463, opened 2023-10-31
Incorrect "refinements checks for subtyping" generated in some cases


mario-bucev:
The following snippet is said to be invalid:
```scala
import stainless.lang._

object RefnChecks {
  def fun(x: BigInt): Option[Some[BigInt]] = {
    (
      if (x == 0) Some[Some[BigInt]](Some(x))
      else None[Some[BigInt]]()
    ) match {
      case _ => None()
    }
  }
}
```
In particular for the VC:
```
[Warning ]  - Result for 'refinements checks for subtyping' VC for fun @6:7:
[Warning ] v$3306.isInstanceOf[Some]
[Warning ] RefnChecks.scala:6:7:  => INVALID
                 if (x == 0) Some[Some[BigInt]](Some(x))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
[Warning ] Found counter-example:
[Warning ]   v$3306: Option[BigInt] -> None[BigInt]()
```
The `AdtSpecialization` introduces these refinements (this transformation seems to be correct):
```scala
def fun(x: BigInt): Option[{ v$3305: Option[BigInt] | v$3305 is Some }] = {
  val targetBound$3303: Boolean = x == 0
  // v$3306 from the VC is from here
  val targetBound$3304: Option[{ v$3306: Option[BigInt] | v$3306 is Some }] = if (targetBound$3303) {
    Some[{ v$3307: Option[BigInt] | v$3307 is Some }](Some[BigInt](x))
  } else {
    None$325[{ v$3308: Option[BigInt] | v$3308 is Some }]()
  }
  targetBound$3304 match {
    case _ =>
      None$325[{ v$3310: Option[BigInt] | v$3310 is Some }]()
  }
}
```

PR #1466, opened 2023-11-01
Applying some type widening in `ReturnElimination` to avoid triggering `AdtSpecialization`

mario-bucev:
For instance, the provided tests in `RefnChecksWithReturn.scala` would unexpectedly generate "refinement checks for subtyping" which may result in being invalid due to #1463.

### DIFF
--- a/core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
@@ -411,8 +411,8 @@ class ReturnElimination(override val s: Trees, override val t: Trees)
             case Seq(e) => transform(e, currentType)
 
             case e +: rest if exprHasReturn(e) =>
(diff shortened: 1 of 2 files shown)


core/src/main/scala/stainless/extraction/imperative/ReturnElimination.scala
```diff
@@ -495,6 +497,19 @@ class ReturnElimination(override val s: Trees, override val t: Trees)
     (res, fnSum)
   }
 
+  // Recursively widen class types to their top ancestor, and strips sigma, pi and refinement types.
+  // This used to type the control flow type parameters in order to avoid triggering AdtSpecialization
```
vkuncak, line 501: This **is** used ?

mario-bucev: whoops, thanks for catching that!

PR #1467, opened 2023-11-02
Fix `while` loops being mistakenly considered as ghost

mario-bucev:
This PR fixes `while` loops being treated as ghost when they referred to ghost constructs in their `invariant`.


vkuncak:
Well, that speeds  up some code!

PR #1471, opened 2023-11-06
Fix filtering for `while` loops and inner functions

mario-bucev:
This changes how `--functions` filters functions in presence of `while` loops and inner functions.
* Loops: when a function is specified in `--functions`, all its `while` functions get verified as well
* Inner function: if a function containing inner functions is passed to `--functions`, all its inner functions get verified as well. It is also now possible to verify an inner function without verifying its parent function.
 
Note: most of the diffs are from merging the test config files for the equivalence checking component due to a test suite refactor.


mario-bucev:
Requires https://github.com/epfl-lara/stainless/pull/1472 to be merged first

vkuncak (approved):

PR #1472, opened 2023-11-06
Change annotations propagation to consider transitive owners

mario-bucev:
Before this PR, *all* annotations from the direct owner of a symbol were applied to that symbol. For instance, a `@library` object containing a function `f` would (correctly) get the `@library` flag, but not any of its inner functions (which is an incorrect behavior).
Furthermore, a function annotated with `@opaque` (incorrectly?) propagates this annotation to its direct inner functions.

This PR changes the following:
* Only the following annotations are propagated to the "children" of a symbol:
`stainless.annotation.ignore`
`stainless.annotation.library`
`stainless.annotation.extern`
`stainless.annotation.dropVCs`
`stainless.annotation.pure`
`stainless.annotation.wrapping`
`stainless.annotation.keep`
`stainless.annotation.keepFor`
`stainless.annotation.cCode.drop`
* These annotations are transitively propagated

### DIFF
--- a/frontends/dotty/src/main/scala/stainless/frontends/dotc/ASTExtractors.scala
+++ b/frontends/dotty/src/main/scala/stainless/frontends/dotc/ASTExtractors.scala
@@ -40,6 +40,21 @@ trait ASTExtractors {
     defn.AnyValType,
   )
 
+  // Annotations that are propagated to symbols owned by an owner containing these.
+  // Note: we do not necessarily want @opaque/@inlineOnce function to have their inner functions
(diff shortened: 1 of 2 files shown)


vkuncak:
I do not understand where transitivity is implemented. Not in the diff as far as I can tell, so it was there already but not consistently so?

mario-bucev:
Before this PR, we only checked for the annotation of the direct owner (`sym.owner`), now we consider all transitive owners using `sym.ownersIterator.drop(1)` (where we drop the first element because `ownersIterator` returns the `sym` itself as first element).

vkuncak (approved):

PR #1476, opened 2023-11-08
Propagate ghost annotation from class to methods

mario-bucev:
PR #1472 changed the way propagation of annotation works, but I did not propagate `@ghost` annotation from `@ghost` class to methods. This PR fixes that.
Note that `@ghost` does not propagate to fields (see [here](https://github.com/epfl-lara/stainless/blob/106629548058cd72c31e965f05a415dc834d6abf/frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala#L500) where we explicitly ignore owners).


vkuncak:
What do you know, we have ghost classes? Presumably those are classes all of whose instances must be ghost?

mario-bucev:
Yes it seems so. The nightly failed due to `sbt scripted` failing for [this test](https://github.com/epfl-lara/stainless/blob/106629548058cd72c31e965f05a415dc834d6abf/sbt-plugin/src/sbt-test/sbt-plugin/ghost/actor-tests/src/main/scala/Actors.scala#L43-L47) (I forgot to test `sbt scripted` locally...).
I've pushed an invalid example to "document" this as well.

PR #1477, opened 2023-11-08
Fix positions for inlined functions

mario-bucev:
In some cases, VCs positions could be incorrect when using inline functions.
This is particularly the case for synthetic `copy` methods which are always inlined.
For instance, in the following snippet, the VC generated for `mc.copy(x = y)` would incorrectly point at the definition of `MyClass`. This PR fixes this issue and the correct position is now reported.
```scala
case class MyClass(x: BigInt, y: BigInt)

def buildClass(mc: MyClass, x: BigInt): MyClass = {
  mc.copy(x = y)
}.ensuring(_.x >= 0)
```


vkuncak:
Are you saying that the position for a function f that is called from g and inlined will also be reported in g? Does this apply to user-defined functions as well?

vkuncak (approved):

Issue #1306, opened 2022-10-10
"Cannot add already used variable" for local function definitions


b-studios:
The following code
```
  def root(a: BigInt): BigInt = {
    require(a >= 0)
    var i: BigInt = 0
    def nextSquare = ((i + 1) * (i + 1))
    (while (nextSquare <= a) {
      decreases(a - nextSquare) // termination
      i = i + 1
    }) invariant { i >= 0 && ((i * i) <= a) }
    i
  } ensuring { root => (root * root) <= a && a < ((root + 1) * (root + 1)) }
```

leads to
```
[error] (verified / Compile / compileIncremental) java.lang.AssertionError: assertion failed: Cannot add already used variable i$42: BigInt to path:
[error] a$1 >= 0 && {
[error]   val i$28: BigInt = 0
[error]   i$32 >= 0 && (i$32 * i$32 <= a$1 && {
[error]     val t$70: (BigInt, BigInt) = sum$1(i$42)
[error]     true
[error]   })
[error] }
```
Please note that inlining `nextSquare` manually resolves the issue.

Might be related to #1209

PR #1478, opened 2023-11-08
Disable test `i1306b`

mario-bucev:
`i1306b` is a nice test to ensure we do not regress on #1306 but it is sadly quite flaky.
I have disabled it (not deleted it) and have added a variant `i1306c` that does not make use of non-linear arithmetic (which seems to be the cause of the issue).

### DIFF
--- a/frontends/benchmarks/imperative/valid/i1306c.scala
+++ b/frontends/benchmarks/imperative/valid/i1306c.scala
@@ -1,0 +1,43 @@
+import stainless._
+import stainless.lang._
+import stainless.annotation._
+
+object i1306c {
+  // Like coal power plants, lot of energy lost for little gain...
+  def laboriousIdentity1(a: BigInt): BigInt = {
+    require(a >= 0)
+    var i: BigInt = 0
+    def nextIncrement = i + 1
+    def accessI(x: BigInt, y: BigInt) = i + x + y
+    def modifyI(x: BigInt, y: BigInt) = { i += x + y }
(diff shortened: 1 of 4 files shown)


frontends/benchmarks/imperative/valid/i1306c.scala
```diff
@@ -0,0 +1,23 @@
...
+import stainless.annotation._
+
+object i1306c {
+  // Like coal power plants, lot of energy lost for little gain...
+  def laboriousIdentity(a: BigInt): BigInt = {
+    require(a >= 0)
+    var i: BigInt = 0
+    def nextIncrement = i + 1
+    def accessI(x: BigInt, y: BigInt) = i + x + y
+    def modifyI(x: BigInt, y: BigInt) = { i += x + y }
```
vkuncak, line 12: @mario-bucev Function `modify` is introduced but not used, is this on purpose, to test that its presence does not cause any issues?

vkuncak (approved):

frontends/benchmarks/imperative/valid/i1306c.scala
mario-bucev: Good catch, since this test is derived from `i1306b`, I do not remember exactly why I've added this function... I'll call it with nullifying arguments.

mario-bucev: I've also added a variant of the function that increments `i` with `modifyI`.

PR #1488, opened 2023-11-23
Equivalence checker: add 'unknown safety' category

mario-bucev:
"unknown safety" refers to inconclusive VCs that are unrelated to equivalence checking.


vkuncak:
@drganam could you check, please?

PR #1502, opened 2024-02-16
Remove duplicated default value doc

shilangyu:
Before this change the information about the default value was duplicated in the docs of `--help`:

```txt
[  Info  ] --strict-arithmetic[=true|false]                 Check arithmetic operations for unintended behavior and overflows (default: true) (default: true)
```

The default value notice is added later, outside of the description string.


mario-bucev:
Thanks for the catch! For some reason the CI is not running and I don't have the GH rights to merge a PR without it running :man_facepalming:  I'll see if someone can do it.

samarion:
There's an allowlist of users for which the larabot CI will run (somewhere int he larabot config). And it seems we've disabled force-push from GH completely, even for admins :)

@vkuncak I'll re-enable force push for administrators so we can merge this PR.

PR #1523, opened 2024-05-08
Remove remaining of Scala 2 in some scripts

mario-bucev:


bin/package-sbt-plugin.sh
```diff
@@ -8,6 +8,7 @@ fi
 
 SCALA_VERSION="3.3.3"
 SCALA_LIB_VERSION="3.3.3"
+PUBLISHED_SBT_PLUGIN_DIR="$HOME/.ivy2/local/ch.epfl.lara/sbt-stainless/scala_2.12/sbt_1.0/$STAINLESS_VERSION"
```
samuelchassot, line 11: is is really `scala_2.12`?

mario-bucev: Yes, SBT plugins are compiled under 2.12 (at least for the SBT versions we have)

samuelchassot (approved):

PR #1562, opened 2024-08-22
Releases notes for 0.9.8.8 and update docs

mario-bucev:
For the 0.9.8.8 release


romac:
I was wondering why the newer releases do not increase either the minor or patch versions anymore but instead use even more fine-grained sub-patch versions? Is there a reason for this?

mario-bucev:
@vkuncak

PR #1580, opened 2024-09-13
Fix Scala version to 3.5.0 in packaging script + add logging

dhalilov:
The packaging script fails as it cannot find the `./frontends/stainless-dotty-standalone/target/scala-3.3.3/stainless-dotty-standalone-xxx.jar` file, since it is located in the `scala-3.5.0` directory instead. This PR fixes that by matching the Scala version in the script with the Scala version used to build Stainless.


dhalilov:
I've made logs visible to the user in the compilation phase (the longest one) to reassure him that the script is not stuck.

PR #1603, opened 2024-11-19
[Experiment] Disable the RefinementLifting phase

mbovel:


mario-bucev:
Aaah I see you also stepped on a rake, glad I was not alone 😄!

samuelchassot:
> Aaah I see you also stepped on a rake, glad I was not alone 😄!

I gave him the notes I took when we tried in August :D maybe someone will finally manage 🤞

vkuncak:
Where do we stand here? I was under impression that this should work in principle. 
Could we diagnose where things go wrong and post some examples here?

PR #1618, opened 2024-12-10
Store all errors caught in the frontend

mbovel:
While trying to debug my modifications to the frontend, I found that some exceptions are swallowed. This makes it very hard to debug the frontend.

I found that this is due to the way errors are handled in `ThreadedFrontend`.

cc @mario-bucev and @samuelchassot: this is the problem I had this afternoon.

### Background

- b0338b2992710adf18c46cf63a22ab837ceacde8 made sure that catching an error in the frontend would call `Callback.failed`.
- bf38baebf645c4c504a14e38fa40860cb961e09e removed the intermediate catch that caught the errors thrown during the frontend `onRun` method.
- `Callback.failed` is now a no-op.

### Current semantic and questions

- `Frontend.onEnd` runs even if exceptions are thrown during `Frontend.onRun`. Is that excepted? I guess this might be to allow `Frontend.onEnd` to recover from some errors?
- If an exception is thrown in `Frontend.onEnd`, it swallows any exception thrown in `Frontend.onRun`. Here is a minimal reproduction:

  ```scala
  @main def main(): Nothing =
    try
      throw new Exception("Forever lost in the void")
    finally
      throw new Exception("Thrown from finally")
  ```

   Is this expected? I think the first exception should at least be logged in debug mode. Should it also be rethrown?
- The first exception which is not `UnsupportedCodeException` thrown during `Frontend.onEnd`, or during `Frontend.onRun` is rethrown without being reported as an error, it is only logged at the debug level.
- `UnsupportedCodeException` which are thrown before the first non-`UnsupportedCodeException` exception are reported as errors. The others are ignored. This seems weird. See `def rethrow`.

### Proposed changes

This PR enables logging all exceptions thrown in the frontend, at debug level.

It doesn't make any other changes. In particular, the semantic described above is preserved.

Depending on the answers to my questions above, further changes might be needed.


mario-bucev:
All the exceptions are caught in the exception handler added before the thread is started:
https://github.com/epfl-lara/stainless/blob/88e4de560a90940e3d1cd81ba3ed826ca9cd543a/core/src/main/scala/stainless/frontend/ThreadedFrontend.scala#L45-L49

These are then handled in a helper method `rethrow`, called in `onStop` and `onJoin`:
https://github.com/epfl-lara/stainless/blob/88e4de560a90940e3d1cd81ba3ed826ca9cd543a/core/src/main/scala/stainless/frontend/ThreadedFrontend.scala#L68-L75

If there is an exception other than `UnsupportedCodeException` (which is usually thrown for features Stainless does not support), all the remaining exceptions are swallowed (due to the `exceptions.clear()`), and that exception is rethrown.
I think we should do something similar to the `UnsupportedCodeException` (with `ctx.reporter.error`). That said, the other kind of exceptions are usually due to a bug e.g. `ClassCastException`.

mario-bucev:
Regarding `Frontend.onEnd`, this method in a no-op in the only concrete implementation of `ThreadedFrontend` (an anonymous class in `DottyCompiler`).

https://github.com/epfl-lara/stainless/blob/4f2e2403e5c21b1a675f211d45c35449ff9acff1/frontends/dotty/src/main/scala/stainless/frontends/dotc/DottyCompiler.scala#L142-L143

https://github.com/epfl-lara/stainless/blob/4f2e2403e5c21b1a675f211d45c35449ff9acff1/frontends/dotty/src/main/scala/stainless/frontends/dotc/DottyCompiler.scala#L175

vkuncak:
@mbovel can you rebase?

Issue #1275, opened 2022-05-30
Tail recursion elimination for genc


vkuncak:
Investigate if it is feasible to peroform tail recursion elimination in genc itself, for non-mutual recursion.

Currently, this is not needed when using genc with GCC output, which performs tail recursion elimination; doing it inside genc would allow the use in those compilers whose authors apparently find tail recursion to be beyond their intellectual abilities.

PR #1626, opened 2025-01-05
Tail recursion elimination for GenC (#1275)

zhekai-jiang:
This is to address #1275 . We added a new phase to perform tail recursion elimination.

We added test cases in the [GenCSuite](https://github.com/epfl-lara/stainless/pull/1626/files#diff-2091c70888d42120d35c352d0b060558b0c5ad02baa02de0ecd77b0fd0ded464).

As discussed during the presentation, we may want to take a closer look at ghost elimination and see whether it is doing the job correctly.


vkuncak:
Verification of `TailRecComplexArgs.scala` failed on the CI.

vkuncak:
also of `TailRecUnitIf.scala`
https://github.com/epfl-lara/stainless/actions/runs/12630587164/job/35190698906

zhekai-jiang:
Could we try to run the tests again? (Apparently it was because we didn't add the `decreases` measures for our new test cases, so Stainless was not able to verify everything, but this issue was not related to the generated C code.)

zhekai-jiang:
By the way, I removed the test case [TailRecReturnInLoop](https://github.com/epfl-lara/stainless/pull/1626/commits/3b01e4d2de03545ba8d53b827148e658a9d8cc72#diff-7fe0c0b871147f9f3217c7ea99a4399e0f88e8a5f34faa440902b3408509f292) because apparently this kind of thing is not allowed by Stainless, because the loop will be rewritten into a function, and if we make a recursive call in the loop, it will make the inner function mutually recursive with the enclosing function and Stainless throws an error (which I think is from https://github.com/epfl-lara/stainless/blob/88e4de560a90940e3d1cd81ba3ed826ca9cd543a/core/src/main/scala/stainless/verification/TypeChecker.scala#L1236-L1251)

samuelchassot:
I rerun the CI 👍 



> By the way, I removed the test case [TailRecReturnInLoop](https://github.com/epfl-lara/stainless/pull/1626/commits/3b01e4d2de03545ba8d53b827148e658a9d8cc72#diff-7fe0c0b871147f9f3217c7ea99a4399e0f88e8a5f34faa440902b3408509f292) because apparently this kind of thing is not allowed by Stainless, because the loop will be rewritten into a function, and if we make a recursive call in the loop, it will make the inner function mutually recursive with the enclosing function and Stainless throws an error (which I think is from
> 
> https://github.com/epfl-lara/stainless/blob/88e4de560a90940e3d1cd81ba3ed826ca9cd543a/core/src/main/scala/stainless/verification/TypeChecker.scala#L1236-L1251
> 
> )

About that: can't we keep the benchmark with the `while` loop then?

zhekai-jiang:
I see 3 failed tests, but they are not related to our part. Anyone has any idea why...?

samuelchassot:
@vkuncak They're the same failing tests as on the other PR. 
That's even more puzzling

samuelchassot:
@sankalpgambhir could it be related to some Princess versions? They're not changing anything in the build file though.

samuelchassot:
We investigated and these tests failing have nothing to do with your PR. It is happening because of the latest release of the Princess Solver. We will ignore those tests in a soon-to-be-merged PR.
You'll be able to update your branch and the CI should then pass :)

PR #1629, opened 2025-01-10
Ignore recently failing Princess tests

sankalpgambhir:
Ignores three verification tests `invalid/Generics2`, `invalid/Overrides`, and `invalid/MyTuple3` that fail on recent Princess builds.

I'll separately investigate why the tests fail (throwing exceptions in Princess).


samuelchassot:
whaaat?!
The lean mergesort test passed yesterday on this run:
https://github.com/epfl-lara/stainless/actions/runs/12693122868/job/35380008438

sankalpgambhir:
Re-ran CI 🙈

PR #1637, opened 2025-01-22
Add AUR package reference and instructions

sankalpgambhir:
Adds the following section to the README, describing the AUR package, copied to be rendered here:

==========================================================

### Arch User Repository

A package for Stainless is available on the Arch User Repository (AUR) for ArchLinux as [`stainless-git`](https://aur.archlinux.org/packages/stainless-git), which follows the latest commit on the `main` branch. You can use your favorite AUR helper (we've tried `yay`, see [AUR Helpers](https://wiki.archlinux.org/title/AUR_helpers)),  or follow the [detailed instructions as on the ArchWiki](https://wiki.archlinux.org/title/Arch_User_Repository#Installing_and_upgrading_packages) to install the package.

For quick reference, with `yay` (or other AUR helpers accordingly):

```shell
yay -S stainless-git
```

and manually:

```shell
git clone https://aur.archlinux.org/stainless-git.git
cd stainless-git
makepkg -si
```

The option `-s` acquires dependencies (`java`, `git`, `sbt`) using `pacman`, `-i` installs Stainless system-wide.  Omit `-i` to avoid installing system-wide, and only perform a local build in the directory.

The solver packages `z3`, `cvc4`, and `cvc5`<sup>AUR</sup> can be added as optional dependencies, and having at least one is recommended for general use.

Issues with the package build should ideally be reported on the [AUR package page](https://aur.archlinux.org/packages/stainless-git) itself. 

==========================================================

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ For quick reference, with `yay` (or other AUR helpers accordingly):
 yay -S stainless-git
 ```
 
-and manually, assuming `sbt`, `git`, and a `java` runtime are available:
+and manually:
 
 ```shell
 git clone https://aur.archlinux.org/stainless-git.git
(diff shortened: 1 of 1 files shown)


sankalpgambhir:
(CI did not fail, was cancelled because it was a waste, having added another commit)

sankalpgambhir:
Should the line about "having at least one [solver] is recommended" be excluded? Native Z3 should be packaged anyway.

PR #1659, opened 2025-07-07
Add watch-json CLI option

quartztz:
The current joint behavior of the `--watch` and `--json` options is such that, on file change, only the terminal view is updated, and the json report is produced once the job is killed. 

In a larger project of developing a Stainless VSCode extension, this PR adds `--watch-json`, an option to be run in tandem with `--watch` that enables report outputs on file change. The addition is minimal (2 files changed), and has been tested to work.

To test this PR, run `sbt universal:stage` in the project root, then use the executable of your choosing (I tested with `stainless-dotty`) as follows: 

```console
$ frontends/dotty/target/universal/stage/bin/stainless-dotty --watch-json --watch path/to/your/file.scala
```

then edit the file, save its contents, and verify that the contents of the generated `report.json` file match the state of the file.


vkuncak:
@quartztz this sounds like a cool useful change! We have a small tutorial doc update coming up, but could you rebase the PR right after that and we will merge it?

sankalpgambhir:
The doc update was merged last week a bit after Viktor's comment!

You can rebase onto it and see that things work.

PR #1666, opened 2025-09-18
Fix crashing on receiving no input files

sankalpgambhir:
Instead, a warning is emitted and the input is treated as producing zero VCs.

The internal options that were previously printed are instead moved to a debug message.


sankalpgambhir:
(please squash)

PR #1669, opened 2025-11-13
Add eqcache option

drganam:
New command line option to optionally pass a cache of equivalent functions. This information is used to bypass the checks for equivalence sublemmas. For now, the cache can only store one set of equivalent functions. In the future, we might want to generalize to accept multiple sets of equivalent functions.


samuelchassot:
TODO
- [ ] generalize the option to accept a list of list of equivalent functions

PR #1678, opened 2025-12-09
Ignore `ClassTag` parameters

mbovel:
This PR implements filtering out parameters in function definitions and argument in function calls that have a `ClassTag` type when extracting code; we just ignore them as they are not relevant for verification.

Done with @samuelchassot.


mbovel:
We do something slightly different for functions and classes. For functions, we just filter the individual arguments, such that:

```scala
def f(x: Int)(c: ClassTag)
```

become:

```scala
def f(x: Int)()
```

But for classes, Stainless would reject classes with several arguments list, so entirely remove lists that are empty after filtering, such that:

```scala
class C(x: Int)(c: ClassTag)
```

becomes:

```scala
class C(x: Int)
```

mbovel:
Currently, invoking Stainless on individual test files works, but we have failure when running the test suite 🤔 


https://github.com/epfl-lara/stainless/actions/runs/20103721835/job/57681480478?pr=1678#step:11:246

```
Cause: inox.package$FatalError: Cannot recover from missing dependencies
[info]   at inox.package$FatalError$.apply(package.scala:24)
[info]   at inox.Reporter.onFatal(Reporter.scala:47)
[info]   at inox.Reporter.fatalError(Reporter.scala:56)
[info]   at inox.Reporter.fatalError(Reporter.scala:113)
[info]   at stainless.frontend.Recovery$.recover(Recovery.scala:111)
[info]   at stainless.frontend.Preprocessing.transform(Preprocessing.scala:18)
[info]   at stainless.InputUtils.stainless$InputUtils$$anon$2$$_$endExtractions$$anonfun$1(InputUtils.scala:86)
[info]   at stainless.extraction.utils.DebugSymbols.debugWithoutSummary$$anonfun$1(NamedPipeline.scala:112)
[info]   at stainless.extraction.utils.DebugSymbols.debug(NamedPipeline.scala:64)
[info]   at stainless.extraction.utils.DebugSymbols.debug$(NamedPipeline.scala:29)
[info]   at stainless.InputUtils$$anon$1.debug(InputUtils.scala:43)
[info]   at stainless.extraction.utils.DebugSymbols.debugWithoutSummary(NamedPipeline.scala:112)
[info]   at stainless.extraction.utils.DebugSymbols.debugWithoutSummary$(NamedPipeline.scala:29)
[info]   at stainless.InputUtils$$anon$1.debugWithoutSummary(InputUtils.scala:43)
[info]   at stainless.InputUtils$$anon$2.endExtractions(InputUtils.scala:86)
[info]   at stainless.frontend.ThreadedFrontend$$anon$1.run(ThreadedFrontend.scala:36)
```

samuelchassot:
TODO:
- [ ] Remove the ignored parameters from patterns

frontends/benchmarks/extraction/valid/ClassTagPatternMatch.scala
```diff
@@ -0,0 +1,23 @@
+import scala.reflect.ClassTag
+import javax.xml.crypto.Data
```
vkuncak: Why are we importing `javax.xml.crypto.Data` ?

mbovel:
So it seems the problem here are default getters of copy methods. Maybe we could generalize to handle all default getters that have ignored types?

Here is a test case that doesn't pass yet:

```scala
import scala.reflect.ClassTag

/** Tests that Stainless accepts `ClassTag` parameters. It should strip them
  * from the extracted code, as they are not needed for verification.
  */
object ClassTagDefaultParams {
  def f(c: ClassTag[Int] = ClassTag.Int): Int = 42
}
```

There is an `canExtractSynthetic` method, I am wondering if we could change it to something like:

```scala
  def canExtractSynthetic(sym: Symbol) = {
    (sym `is` Implicit) ||
    (isDefaultGetter(sym) && !isIgnoredParameterType(sym.info.resultType))  ||
    isCopyMethod(sym)
  }
```

But that alone doesn't work.

samuelchassot:
I'll merge the current PR and move this to an issue, we'll work on it next week but at least I can release Stainless with the new feature, and I don't think default ClassTags are often seen, so it shouldn't be an issue in the meantime

PR #1691, opened 2025-12-18
Fix bugs in classtag elimination found when adapting ZipLex

samuelchassot:


samuelchassot:
I'm getting another error at the type checking step (inox typechecking so when VCs are generated). I'm debugging to see whether it is related before merging this one.

samuelchassot:
Bug is described in #1692, is probably related, but can be fixed later IMO

PR #1694, opened 2025-12-22
Fix a bug in extraction for method calls with multiple list of arguments

### LINKED ISSUE #1692: Trait's default implementations crash when using implicit `ClassTag`
This example crashes at the type checking level (inox typechecking):

```scala
import stainless.collection.List
import scala.reflect.ClassTag

object BugReproduce:
  trait Interface:
    def pred[C: ClassTag](l: List[C]): Boolean = true
    def f[C: ClassTag](l: List[C]): List[C]
  end Interface

  case object Impl extends Interface:
    override def f[C: ClassTag](l: List[C]): List[C] = {
      require(pred(l))
      l
    }
  end Impl
end BugReproduce
```

with the following error (with some debug prints)
```
Calling freshBindWithValues from infertype for function pred[C]({   assert(!forall((res: Inte...

[Internal] Function `freshBindWithValues` expects sequences with the same size. 
[Internal] DEBUG: vds.size = 2, es.size = 1
[Internal] Got vds = [thiss: Interface, l: List[C]]
[Internal] es = [assert(!forall((res: Interface) => !false), "Choose satisfiability")
[Internal] choose$2[C](thiss, l)]
[Internal] Please inform the authors of Inox about this message
[ Error  ] Stainless terminated with an error.
[ Error  ] Debug output is available in the file `stainless-stack-trace.txt`. If the crash is caused by Stainless, you may report your issue on https://github.com/epfl-lara/stainless/issues
[ Error  ] You may use --debug=stack to have the stack trace displayed in the output.
```

without class tags, this works fine:

```scala
import stainless.collection.List
import scala.reflect.ClassTag

object BugReproduce:
  trait Interface:
    def pred[C](l: List[C]): Boolean = true
    def f[C](l: List[C]): List[C]
  end Interface

  case object Impl extends Interface:
    override def f[C](l: List[C]): List[C] = {
      require(pred(l))
      l
    }
  end Impl
end BugReproduce
```

The workaround is to extract the methods that won't be overridden in a companion object.

samuelchassot:
Fixes a problem in the extraction of method calls with multiple lists of arguments. Previously, the receiver (i.e., probably `this`) was lost as it was extracted as a function call without receiver.

Fixes #1692 

Was discovered while working on class tags, as they add implicitly a new list of parameters to methods.

Example:

```scala
trait Interface:
  def pred[C: ClassTag](l: List[C]): Boolean = true
  def f[C: ClassTag](l: List[C]): List[C]
end Interface

case object Impl extends Interface:
  override def f[C: ClassTag](l: List[C]): List[C] = {
    require(pred(l))
    l
  }
end Impl
```

Here, the call to `pred` in the `require` of `f` was extracted as: 

```scala
@final
@method(Impl)
def f[C](l: List[C]): List[C] = {
  require(pred[C](l))
  l
}
```

instead of 

```scala
@final
@method(Impl)
def f[C](l: List[C]): List[C] = {
  require(this.pred[C](l))
  l
}
```

See #1692 for more details.

### DIFF
--- a/frontends/benchmarks/extraction/valid/ClassTagMethodCalls.scala
+++ b/frontends/benchmarks/extraction/valid/ClassTagMethodCalls.scala
@@ -1,0 +1,19 @@
+import stainless.collection.List
+import scala.reflect.ClassTag
+
+object ClassTagMethodCalls:
+  trait Interface:
+    def pred[C: ClassTag](l: List[C]): Boolean = true
+    def f[C: ClassTag](l: List[C]): List[C]
+  end Interface
+
+  case object Impl extends Interface:
+    override def f[C: ClassTag](l: List[C]): List[C] = {
+      require(pred(l))
+      l
+    }
+  end Impl
+
+  @main def classTagMethodCallsMain(): Unit =
+    val res = Impl.f(List(1, 2, 3))
+end ClassTagMethodCalls
--- a/frontends/benchmarks/extraction/valid/MethodCallsWithMultipleArgsLists.scala
+++ b/frontends/benchmarks/extraction/valid/MethodCallsWithMultipleArgsLists.scala
@@ -1,0 +1,18 @@
+import stainless.collection.List
+
+object MethodCallsWithMultipleArgsLists:
+  trait Interface:
+    def pred[C](l: List[C])(x: Int)(y: Int): Boolean = true
+    def f[C](l: List[C]): List[C]
+  end Interface
+
+  case object Impl extends Interface:
+    override def f[C](l: List[C]): List[C] = {
+      require(pred(l)(0)(0))
+      l
+    }
+  end Impl
+
+  @main def methodCallsWithMultipleArgsListsMain(): Unit =
+    val res = Impl.f(List(1, 2, 3))
+end MethodCallsWithMultipleArgsLists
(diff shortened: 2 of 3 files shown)


mbovel (commented):
This seems okay-ish as a quick fix.

However, I find it misleading that `ExCall` also matches `ExThisCall`. These extractors are mainly used here:

https://github.com/epfl-lara/stainless/blob/2b3d9c0d42cd425b73d3fe6269381c9a7e10c937/frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala#L1867-L1878

where there is clearly a case for `this.m(...)` and a distinct case for `m(...)`.

If I understand correctly, after this PR, `this`-calls with multiple argument lists will not be matched by `ExThisCall` but instead by `ExCall`, while `this`-calls with a single argument list will still be matched by `ExThisCall`.

That creates a semantic overlap where `ExCall` subsumes part of `ExThisCall`’s intended domain. This makes the extractor behavior arity-dependent, which I find confusing and harder to reason about.

I would find it more coherent to either:

1. Extend `ExThisCall` to support multiple argument lists, or
2. Let `ExCall` match `this`-calls as it currently does, but then completely remove `ExThisCall`, which should no longer be necessary.

---

On a separate note, `ExThisCall` also seems limited to paths of depth one: it matches `this.m` but not `this.c.m`, since it only recognizes a single top-level `TermRef` and does not recurse. We might want to test this and potentially address it separately.

PR #1723, opened 2026-03-30
Fix versions for sbt plugin

samuelchassot:
@mbovel @kasiaMarek 
These versions also need to be updated when we update the Stainless version, otherwise the sbt plugin cannot be packaged  👍


mbovel (approved):

samuelchassot:
@mbovel what about we do this?

PR #1726, opened 2026-04-02
Refinement pattern matching

kasiaMarek:
Should be reviewed and merged after https://github.com/epfl-lara/stainless/pull/1725.

### DIFF
--- a/core/src/main/scala/stainless/evaluators/RecursiveEvaluator.scala
+++ b/core/src/main/scala/stainless/evaluators/RecursiveEvaluator.scala
@@ -145,7 +145,8 @@ abstract class RecursiveEvaluator(override val program: Program,
           case _ => None
         }
 
-      case (RefinementPattern(vd, underlying, pred), scrut) =>
+      case (RefinementPattern(underlying, Lambda(vds, pred)), scrut) =>
+        val Seq(vd) = vds: @unchecked
         matchesPattern(underlying, scrut).flatMap { mapping =>
           if (e(pred)(using rctx.withNewVar(vd, scrut), gctx) == BooleanLiteral(true)) {
             Some(mapping + (vd -> scrut))
--- a/core/src/main/scala/stainless/evaluators/RecursiveEvaluator.scala
+++ b/core/src/main/scala/stainless/evaluators/RecursiveEvaluator.scala
@@ -145,6 +145,15 @@ abstract class RecursiveEvaluator(override val program: Program,
           case _ => None
         }
 
+      case (RefinementPattern(vd, underlying, pred), scrut) =>
+        matchesPattern(underlying, scrut).flatMap { mapping =>
+          if (e(pred)(using rctx.withNewVar(vd, scrut), gctx) == BooleanLiteral(true)) {
+            Some(mapping + (vd -> scrut))
+          } else {
+            None
+          }
+        }
+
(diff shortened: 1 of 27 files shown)


vkuncak:
@mbovel can you take a look? Both for checking and so you get more of a feel what's inside Stainless

core/src/main/scala/stainless/extraction/oo/TransformerWithType.scala
```diff
@@ -49,6 +49,12 @@ trait TransformerWithType extends TreeTransformer {
       val rsubs = subs map (transform(_, tpe))
       t.AlternativePattern(ob map transform, rsubs).copiedFrom(pat)
 
+    case s.RefinementPattern(underlying, pred) =>
+      t.RefinementPattern(
+        transform(underlying, tpe),
+        transform(pred).asInstanceOf[t.Lambda]
```
kasiaMarek: Check and throw if not Lambda

core/src/main/scala/stainless/evaluators/RecursiveEvaluator.scala
```diff
@@ -145,6 +145,16 @@ abstract class RecursiveEvaluator(override val program: Program,
           case _ => None
         }
 
+      case (RefinementPattern(underlying, Lambda(vds, pred)), scrut) =>
```
kasiaMarek, line 148: Add a test?

core/src/main/scala/stainless/extraction/oo/TypeEncoding.scala
```diff
@@ -890,12 +890,23 @@ class TypeEncoding(override val s: Trees, override val t: Trees)
...
+        case _ => body
+      }
+
     override def transform(pat: s.Pattern, tpe: s.Type): t.Pattern = pat match {
       case s.WildcardPattern(None) => t.WildcardPattern(None).copiedFrom(pat)
       case s.WildcardPattern(Some(vd)) if isObject(vd.tpe) && isObject(tpe) =>
         t.WildcardPattern(Some(transform(vd))).copiedFrom(pat)
       case s.WildcardPattern(Some(vd)) =>
-        instanceOfPattern(t.WildcardPattern(Some(transform(vd))).copiedFrom(pat), tpe, vd.getType)
+        withRefine(vd) {
```
kasiaMarek, line 907: We only preserve top-level refinements. Should we somehow keep the nested ones? Maybe not, we think the only way the type param would be refined is if the scrutinee was refined.

frontends/benchmarks/refinement/valid/RefinementMatch.scala
```diff
@@ -0,0 +1,17 @@
...
+    case p : Pos => true
+    case _ : {i: Int with i == 0} => true
+
+case class Box(val value: Int)
+
+def checkNonNegative2(b: Boolean): Boolean =
+  val box = Box(if b then 42 else 0)
+  box match
+    case Box(p : Pos) => true
+    case b @ Box(_ : {i: Int with i == 0}) => true
```
kasiaMarek:
```suggestion
    case b @ Box(i: Int with i == 0) => true
```

mbovel (approved):
We reviewed the code together with @kasiaMarek this morning. Looks good to me overall! Minor comments are above and will be investigated before merging.

PR #1729, opened 2026-04-14
Make refinement type predicates ghost environments

### LINKED ISSUE #1728: Refinement type predicates should be ghost contexts
For specification purposes, refinement types predicates should allow to access ghost methods and fields.

E.g.,
```scala
    @pure 
    @ghost
    def valid: Boolean
    
    @pure
    def contains(key: {k: Long with valid}): {res: Boolean with valid && (res == abstractMap.contains(key))}
```

should work. Currently, it fails:

```scala
[ Error  ] ./src/main/scala/com/mutablemaps/map/MutableMapsInterface.scala:63:37: Cannot access a ghost symbol outside of a ghost context. [ this.valid in method $anonfun ]
               def contains(key: {k: Long with valid}): {res: Boolean with valid && (res == abstractMap.contains(key))}
                                               ^^^^^
```

I will modify the ghost checker.

Refinement type predicates appearing in `isInstanceOf` and in Patterns must NOT be ghost, as they are executed at runtime. This will likely require a new state in the ghost checker.

samuelchassot:
fixes #1728


vkuncak:
@samuelchassot your test cases all seem to import silent checks. Should we also test that the other import options do what we want?

samuelchassot:
> @samuelchassot your test cases all seem to import silent checks. Should we also test that the other import options do what we want?

Currently, `isInstanceOf` is usable only on classes within Stainless, meaning that the runtime checks option is not usable for now. This is an example:

```scala
import language.experimental.qualifiedTypes.runtimeChecks
import stainless.annotation.ghost
import stainless.lang._
import stainless.lang.StaticChecks._

case class A(@ghost val x: BigInt) {

  def f(v: BigInt): Boolean = {
    require(v >= x)
    // The following refinement type should not be allowed to access ghost symbols, as its predicate is executed at runtime.
    val h: {i: BigInt with i >= x} = v
    true
  }
}
```

and the current output:

```
[info] [  Info  ] Compiling with standard Scala 3.8.4-RC1-bin-20260327-ca312c4-NIGHTLY compiler front end...
[info] [ Error  ] /Users/samuel/EPFL/stainless/frontends/benchmarks/extraction/invalid/RefinementGhost03.scala:11:38: isInstanceOf can only be used with class types
[info]                val h: {i: BigInt with i >= x} = v
[info]                                                 ^
[info] [  Info  ] Finished compiling
```

So I suggest we do not add it now, and we look into it later when the `isInstanceOf` is supported.

mbovel (approved):
Looks good to me!

samuelchassot:
@kasiaMarek Oh amazing, I was about to fix that this morning!
The error was so lame hahah Thanks a lot for the fix :)

mbovel (approved):

PR #1730, opened 2026-04-24
feat: support `isInstaceOf` for base types

kasiaMarek:


vkuncak:
Simpler. more general, and does not slow down CI, so LGTM

mbovel:
Youhou, I didn't expect it to be that short, cool!

PR #1731, opened 2026-05-12
fix an issue with function closure with refinement types

kasiaMarek:
when lifting a nested function pass all local values as refined parameters


samuelchassot:
I found one of the bug, for the file `frontends/benchmarks/imperative/valid/FindIndex.scala` which can be minimized to
```scala
import stainless.annotation._
import stainless.lang._
import stainless.io._

object FindIndex {
  def containsFrom[T](a: Array[T], t: T, j: Int): Boolean = {
    require(0 <= j && j < a.length)
    false
  }

  @pure @extern // admitted lemma
  def containsFromLemma[T](a: Array[T], t: T, j: Int): Unit = {
    require(0 <= j && j < a.length && containsFrom(a, t, j) && a(j) != t)

 }.ensuring(_ =>
    containsFrom(a, t, j + 1)
  )

  @inline
  def contains[T](a: Array[T], t: T): Boolean = a.length > 0 && containsFrom(a, t, 0)

  def findIndex[T](a: Array[T], t: T): Int = {
    require(contains(a, t))
    var i: Int = 0
    (while (true) {
      if (a(i) == t) then return i
      i += 1
    })
    assert(false)
    0
  }
}

```

The problem comes from `FunctionInlining`, which inlines `contains` in the body of the function for the `while` loop, but not in the refinement type in the paremeter.

So in the wellFormdness check, when it checks for unbound variables (in `inox/src/main/scala/inox/ast/Definitions.scala:230 -> 235`), the `prop$2` variable is unbound because the types do not match:
tree:
```scala
def findIndexWhile[T](a: Array[T], t: T, prop: { param: Unit | @DropVCs @dropConjunct contains[T](a, t) }, i: Int): (ControlFlow[Int, Unit], Int) = {
  val i: Int = i
  val ix: Int = i
  val targetBound: Boolean = a(ix) == t
  val t: ControlFlow[Int, Unit] = if (targetBound) {
    Return[Int, Unit](i)
  } else {
    Proceed[Int, Unit](())
  }
  val res: ControlFlow[Int, Unit] = t
  val cf: ControlFlow[Int, Unit] = res
  val t: (ControlFlow[Int, Unit], Int) = cf match {
    case Return(retValue) =>
      (Return[Int, Unit](retValue), i)
    case Proceed(proceedValue) =>
      val i: Int = i + 1
      (Proceed[Int, Unit](()), i)
  }
  val res: ControlFlow[Int, Unit] = t._1
  val i: Int = t._2
  res match {
    case Return(retValue) =>
      (Return[Int, Unit](retValue), i)
    case Proceed(proceedValue) =>
      if (true) {
        findIndexWhile[T](a, t, prop, i)
      } else {
        (Proceed[Int, Unit](()), i)
      }
  }
}.ensuring {
  (cfWhile: (ControlFlow[Int, Unit], Int)) => {
    val i: Int = cfWhile._2
    cfWhile._1 match {
      case Return(retValue) =>
        true
      case Proceed(proceedValue) =>
        !true
    }
  }
}
```

and printing the unbound variable:

```scala
[info] Unbound variables: uniqueName= prop$2, type= { param: Unit | @DropVCs @dropConjunct {
[info]   val a: Array[T] = a
[info]   val t: T = t
[info]   @DropVCs (a.length > 0 && containsFrom[T](a, t, 0))
[info] } }
```

I am working on a fix.

EDIT:
the fix is to inline functions in function parameters too:
```scala
if ((fd.flags contains Synthetic) && (fd.flags contains Inline)) (None, ())
    else (Some(identity.transform(fd.copy(
      fullBody = new Inliner().transform(fd.fullBody),
      params = fd.params.map(vd => vd.copy(tpe = new Inliner().transform(vd.tpe))), // <-- NEW
      flags = fd.flags filterNot (f => f == Inline || f == InlineOnce)
    ))), ())
  }
```

core/src/main/scala/stainless/extraction/innerfuns/FunctionClosure.scala
```diff
@@ -29,6 +31,11 @@ class FunctionClosure(override val s: Trees, override  val t: ast.Trees)
       tparamsMap: Map[TypeParameter, TypeParameter]
     )
 
+    def isProp(vd: ValDef): Boolean = vd.tpe match {
```
samuelchassot: "prop" here is what we discussed for preconditions without parameters? Could we maybe have something clearer?

core/src/main/scala/stainless/extraction/innerfuns/FunctionClosure.scala
```diff
@@ -47,38 +54,27 @@ class FunctionClosure(override val s: Trees, override  val t: ast.Trees)
...
         }
 
       val freeMap = freshVals.toMap
-      val freshParams = freshVals.filterNot(p => pc.bindings exists (_._1.id == p._1.id)).map(_._2)
-
-      // We annotate outer path conditions with `DropConjunct` so that they are not checked when
-      // calling the inner function (as we know they already hold at this point).
-      // And we annotated bound expressions and path conditions with `DropVCs` so that they
-      // don't generate verification conditions (e.g. index within bounds), as these would be
-      // already checked in the outer function.
```
samuelchassot, line 73: Are we still annotating with DropConjuct to assume they're correct? if so let's keep the comment somewhere.

samuelchassot:
I will look at the trees, as it is hard to visualize the transformation by looking a the code.

samuelchassot:
Apart from the comments it looks good to me. The trees look good, it seems like it is doing the right thing. I will start a full nightly CI on this branch so that we are sure it works on the whole bolts repo before merging.

samuelchassot (changes requested):

core/src/main/scala/stainless/extraction/innerfuns/FunctionClosure.scala
kasiaMarek: I can introduce a special term and type trees for prop. But I'm not sure when it would be best to desugar them into `()` and `Unit` as adding another type to Inox seems redundant. I also don't want to add a new phase for something this trivial.

samuelchassot: Oh no, sorry, I meant the naming here in the code. That was not obvious to me what it meant before looking for the usage.

kasiaMarek: ~Here is a user-facing prop: https://github.com/epfl-lara/stainless/commit/d42e80c7adbbef8441f93981a684e8b4a7d47986 .~
~I can also extract it into a special tree in Stainless, but I'm not sure if it's worth it.~
Nvm, different discussion.

kasiaMarek: Does `isConditionAsRefinement` work?

samuelchassot: yep :)

samuelchassot:
I passed nightly with the fix from #1746 👍

PR #1733, opened 2026-05-19
Create a sanitize pass that removes QualifierSkolemIndex annotations 

kasiaMarek:


mbovel (approved):
That looks great to me!

I think running this `TreeMap` before extraction will be more robust and more maintainable.

As @kasiaMarek mentioned, the drawback is the additional tree traversal.
My guess is that it will not significantly hurt the performance.

PR #1737, opened 2026-06-01
Enable silent flag for refinement types by default

kasiaMarek:
Adds silent flag for the refinement types by default.
Adds the `IntListStructure` test.


frontends/benchmarks/refinement/valid/ListStructure.scala
samuelchassot, line 1: I also added it in a previous PR. Is it a different benchmark? 
https://github.com/epfl-lara/stainless/pull/1735

samuelchassot (approved):
Apart from the potentially duplicate benchmark, LGTM, we can merge.

samuelchassot:
I just checked, and it is (slightly?) different, so we can keep it.
Any objection? @kasiaMarek

kasiaMarek:
> I just checked, and it is (slightly?) different, so we can keep it.
> Any objection? @kasiaMarek

Sorry, I didn't notice you added it already. It's almost the same, let me remove it.

samuelchassot:
@kasiaMarek Let's actually keep it, as yours has refinements :)

PR #1747, opened 2026-06-09
Refinement mutation

kasiaMarek:
- [ ] add a check that variables are not referenced from refinements on variables


samuelchassot:
@kasiaMarek can you update to main?

samuelchassot:
@kasiaMarek @mbovel 
I am working on an array of `Int` with a `def valid: Boolean`, and I want to use refinement types to avoid wrapping it to enforce the invariant, as we wanted :) See https://github.com/epfl-lara/bolts/pull/161 for more details.

 While testing locally, I could not prove the updated in the caller, because of missing transformation to Sigma type, so I used this branch merged with main (to have access to Int <-> BigInt), but it still does not work.
 
This is the minimal example I could find to reproduce the bug:

```scala
@mutable
  case class Box(var value: Int){
    def valid: Boolean = value >= 0
    def update(v: Int): Unit = {
      require(valid)
      value = v
    }.ensuring(_ => valid)
  }
  def test3(b: Box with b.valid): Unit = {
    require(b.value == 1)
    assert(b.value == 1)
    b.update(10)
    assert(b.value == 10) // FAILS TO PROVE
  }
```

and indeed, the the is wrong:

```scala
[ Debug  ] def test3(b: { param0: Box | valid(param0) }): Unit = {
[ Debug  ]   require(((b.value == 1)): @ghost )
[ Debug  ]   assert(((b.value == 1)): @ghost )
[ Debug  ]   val res: (Box, Unit) = update(b, 10)
[ Debug  ]   val tmp: Unit = res._2
[ Debug  ]   assert(((b.value == 10)): @ghost )
[ Debug  ]   ()
[ Debug  ] }
```

as the result of `update` is never used.

samuelchassot:
Same is happening with 
```scala
def test2(): Unit = {
    val b: Box with b.valid = Box(1)
    assert(b.value == 1)
    b.update(10)
    assert(b.value == 10)
  }
```
with trees
```scala
[ Debug  ] def test2: Unit = {
[ Debug  ]   val b: { param0: Box | valid(param0) } = {
[ Debug  ]     val x: Box = Box(1)
[ Debug  ]     assert(true && {
[ Debug  ]       val param0: Box = x
[ Debug  ]       valid(param0)
[ Debug  ]     }, "Cast error")
[ Debug  ]     x
[ Debug  ]   }
[ Debug  ]   assert(b.value == 1)
[ Debug  ]   val res: (Box, Unit) = update(b, 10)
[ Debug  ]   val tmp: Unit = res._2
[ Debug  ]   assert(b.value == 10)
[ Debug  ]   ()
[ Debug  ] }
```

PR #1755, opened 2026-06-23
Add BitVector <->  Integer to frontend

samuelchassot:


samuelchassot:
Then we can prove things like

```scala
def isizeEqSize[T](xs: List[T]): Unit = {
    decreases(xs)
    xs match {
      case Nil() => ()
      case Cons(_, tail) => isizeEqSize(tail)
    }
  }.ensuring(_ => xs.size <= bitVectorToBigInt(Int.MaxValue) ==> (xs.size == bitVectorToBigInt(xs.isize)))
  ```

**Comment by samuelchassot** [comment]
@sankalpgambhir this is ready :)

File: frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala [hunk 1]
```diff
@@ -1603,7 +1603,12 @@ class CodeExtraction(inoxCtx: inox.Context,
     case ExBigIntLiteral(Literal(cnst)) =>
       xt.IntegerLiteral(BigInt(cnst.stringValue))
 
-    case ExBigIntLiteral(_) => outOfSubsetError(tr, "Only literal arguments are allowed for BigInt.")
+    // `BigInt(x)` on a non-literal `Int`/`Long` denotes the integer represented by that bitvector
+    case ExBigIntLiteral(arg)
+        if arg.tpe.widenDealias.typeSymbol == defn.IntClass || arg.tpe.widenDealias.typeSymbol == defn.LongClass =>
+      xt.BVToInt(extractTree(arg)).setPos(tr.sourcePos)
+
+    case ExBigIntLiteral(_) => outOfSubsetError(tr, "Only literal, Int or Long arguments are allowed for BigInt.")
```
vkuncak, line 1611: Actually, if we wanted to work with very large constants, the only way to write them as literal is as strings. Should we perhaps allow extracting BigInt(s) where s is a literal string that, at compile time, parses into a valid BigInt ?

vkuncak (approved):

frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
vkuncak: What do you think, @samuelchassot ?

samuelchassot: Good point!
I think it already works, as demonstrated by benchmarks, see my other comment 👍

frontends/benchmarks/strictarithmetic/invalid/BigIntToLongOverflow.scala
```diff
@@ -0,0 +1,8 @@
+import stainless.lang._
+
+object BigIntToLongOverflow {
+  // 2^63 is one past Long.MaxValue: `.toLong` would wrap around, so under strict
+  // arithmetic the conversion is rejected.
+  def tooBig: Long = BigInt("9223372036854775808").toLong
+  def tooBig2: Long = BigInt("10000000000000000000000000000").toLong
```
samuelchassot, line 7: like here, with a string literal that is too large to fit in a long.

frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
samuelchassot: this example works
```scala
object TestBigInt {
  def f() = {
    val i = 43
    val bi = BigInt(i)
    val verylargebigint = BigInt("1234567890123456789012345678900000000000032040300303003030303030303030")
    assert(verylargebigint > bi)
  }
}
```

PR #1756, opened 2026-06-25
Add lemmas to library to relate big int and int operations

samuelchassot:


samuelchassot:
@sankalpgambhir this should trigger the bug

PR #1757, opened 2026-06-25
Use old in refinements as post condition

samuelchassot:


frontends/benchmarks/extraction/invalid/RefinementVariable2.scala
```diff
@@ -0,0 +1,8 @@
+package variable2
+
+case class A(var x: Int):
+  def getX(): Int = x
+  def isXPos(): Boolean = getX() > 0
+  def f(): Unit =
+    var err: Int with (err > 0) == isXPos() = x
```
vkuncak, line 7: @samuelchassot can we avoid writing  `Int with x == y = z`
I find it really confusing. Let's put parentheses in such cases, e.g. ... `(Int with x == y) = z`

frontends/dotty/src/main/scala/stainless/frontends/dotc/CodeExtraction.scala
```diff
@@ -883,6 +883,65 @@ class CodeExtraction(inoxCtx: inox.Context,
       }
     }
 
+    // Allow `old(...)` to appear in a refinement type on the return type, mirroring its
+    // use in `ensuring`. Since `old` refers to the pre-state, it cannot remain inside a
+    // type, so we lift such a refinement predicate into a postcondition and erase the
```
vkuncak, line 888: @samuelchassot did you really intend that we lift `old` and transform it into ensuring? That sounds a bit like putting back the refinement lifting that @kasiaMarek worked hard to remove? It means that the refinement is lost as a type. For example, when you pass such function to a list map, you will not get that all elements of the result have the refined types.

vkuncak (changes requested):

PR #1758, opened 2026-07-02
Refinement mutation

samuelchassot:
subsumes #1747 , so that we can contribute to the branch


samuelchassot:
Copying discussions from #1747 

@kasiaMarek @mbovel 
I am working on an array of `Int` with a `def valid: Boolean`, and I want to use refinement types to avoid wrapping it to enforce the invariant, as we wanted :) See https://github.com/epfl-lara/bolts/pull/161 for more details.

 While testing locally, I could not prove the updated in the caller, because of missing transformation to Sigma type, so I used this branch merged with main (to have access to Int <-> BigInt), but it still does not work.
 
This is the minimal example I could find to reproduce the bug:

```scala
@mutable
  case class Box(var value: Int){
    def valid: Boolean = value >= 0
    def update(v: Int): Unit = {
      require(valid)
      value = v
    }.ensuring(_ => valid)
  }
  def test3(b: Box with b.valid): Unit = {
    require(b.value == 1)
    assert(b.value == 1)
    b.update(10)
    assert(b.value == 10) // FAILS TO PROVE
  }
```

and indeed, the the is wrong:

```scala
[ Debug  ] def test3(b: { param0: Box | valid(param0) }): Unit = {
[ Debug  ]   require(((b.value == 1)): @ghost )
[ Debug  ]   assert(((b.value == 1)): @ghost )
[ Debug  ]   val res: (Box, Unit) = update(b, 10)
[ Debug  ]   val tmp: Unit = res._2
[ Debug  ]   assert(((b.value == 10)): @ghost )
[ Debug  ]   ()
[ Debug  ] }
```

as the result of `update` is never used. 


Same is happening with 
```scala
def test2(): Unit = {
    val b: Box with b.valid = Box(1)
    assert(b.value == 1)
    b.update(10)
    assert(b.value == 10)
  }
```
with trees
```scala
[ Debug  ] def test2: Unit = {
[ Debug  ]   val b: { param0: Box | valid(param0) } = {
[ Debug  ]     val x: Box = Box(1)
[ Debug  ]     assert(true && {
[ Debug  ]       val param0: Box = x
[ Debug  ]       valid(param0)
[ Debug  ]     }, "Cast error")
[ Debug  ]     x
[ Debug  ]   }
[ Debug  ]   assert(b.value == 1)
[ Debug  ]   val res: (Box, Unit) = update(b, 10)
[ Debug  ]   val tmp: Unit = res._2
[ Debug  ]   assert(b.value == 10)
[ Debug  ]   ()
[ Debug  ] }
```

samuelchassot:
here is a patch, partly suggested by claude, so please @kasiaMarek could you have a look? feel free to fix however you want :)
```
diff --git a/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala b/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
index 3a00faed..d4fa4c85 100644
--- a/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala
@@ -770,6 +770,7 @@ trait EffectsAnalyzer extends oo.CachingPhase {
     case at: ADTType   => ADTFieldAccessor(id)
     case ct: ClassType => ClassFieldAccessor(id)
     case ta: TypeApply => typeToAccessor(ta.getType, id)
+    case rt: RefinementType => typeToAccessor(rt.vd.tpe, id)
     case _ => throw FatalError(s"Cannot have accessors over type $tpe")
   }
 
diff --git a/core/src/main/scala/stainless/extraction/imperative/SymbolOps.scala b/core/src/main/scala/stainless/extraction/imperative/SymbolOps.scala
index 34442ce1..7094de72 100644
--- a/core/src/main/scala/stainless/extraction/imperative/SymbolOps.scala
+++ b/core/src/main/scala/stainless/extraction/imperative/SymbolOps.scala
@@ -93,6 +93,11 @@ trait SymbolOps extends TypeOps with oo.SymbolOps { self =>
     case arr: ArrayType => true
     case map: MutableMapType => true
     case ft: FunctionType => false
+    case pt: PiType => false
+    // The base type of a RefinementType is hidden inside its ValDef, so the
+    // generic NAryType case would not see it (RefinementType deconstructs with no sub-types).
+    case RefinementType(vd, _) => isMutableType(vd.tpe, mutableClasses, visited)
+    case SigmaType(params, to) => (params.map(_.tpe) :+ to).exists(isMutableType(_, mutableClasses, visited))
     case ct: ClassType => isMutableClassType(ct, mutableClasses, visited)
     case adt: ADTType => isMutableADTType(adt, mutableClasses, visited)
     case ta: TypeApply if ta.isAbstract => ta.getTypeDef.flags contains IsMutable
diff --git a/core/src/main/scala/stainless/extraction/methods/MutabilityAnalyzer.scala b/core/src/main/scala/stainless/extraction/methods/MutabilityAnalyzer.scala
index e804aedd..0aea3132 100644
--- a/core/src/main/scala/stainless/extraction/methods/MutabilityAnalyzer.scala
+++ b/core/src/main/scala/stainless/extraction/methods/MutabilityAnalyzer.scala
@@ -51,6 +51,11 @@ trait MutabilityAnalyzer extends oo.ExtractionPipeline { self =>
             fd.isGetter && rec(fd.returnType, seen + cid)
           }
         case _: FunctionType => false
+        case _: PiType => false
+        // The base type of a RefinementType is hidden inside its ValDef, so the
+        // generic NAryType case would not see it (RefinementType deconstructs with no sub-types).
+        case RefinementType(vd, _) => rec(vd.tpe, seen)
+        case SigmaType(params, to) => params.exists(vd => rec(vd.tpe, seen)) || rec(to, seen)
         case NAryType(tps, _) => tps.exists(rec(_, seen))
       }
 ```

**Comment by samuelchassot** [comment]
@kasiaMarek if you don't have time to work on that, just let me know and I'll take care of it :)

**Comment by kasiaMarek** [comment]
> could you have a look?

I think the patch looks good. Because `Box` had a refinement it didn't recognize it a MutableType, so normalization in AntiAlising didn't fire properly. But the change in `EffectsAnalyzer` is redundant, this is only called with `.getType`, so `RefinementType` won't be there.

**Comment by samuelchassot** [comment]
> > could you have a look?
> 
> I think the patch looks good. Because `Box` had a refinement it didn't recognize it a MutableType, so normalization in AntiAlising didn't fire properly. But the change in `EffectsAnalyzer` is redundant, this is only called with `.getType`, so `RefinementType` won't be there.

Thank you!

File: core/src/main/scala/stainless/extraction/imperative/EffectsAnalyzer.scala [hunk 1]
```diff
@@ -150,9 +151,20 @@ trait EffectsAnalyzer extends oo.CachingPhase {
       fd.params.filter(vd => receivers(vd.toVariable))
     }
 
+    /* Since the return type may depend on the `aliasedParams` (they may be referenced in a refinement type),
+     * we return a sigma type with fresh variables for `aliasedParams` and a the original return type.
```
samuelchassot:
```suggestion
     * we return a sigma type with fresh variables for `aliasedParams` and the original return type.
```

samuelchassot:
I can't approve my own PR lol, but when this passes, this can be merged :)

PR #1759, opened 2026-07-07
Add golden-file snapshots of generated C to GenCSuite

mbovel:
Compare each genc/valid benchmark's generated .c and .h against committed .expected.c/.expected.h golden files, so unintended changes to the emitted code (not just compilation/behaviour) are caught. Set STAINLESS_GENC_UPDATE_EXPECTED=1 to (re)generate the goldens; on mismatch the test prints a unified diff for CI.

The golden files use an .expected.c/.expected.h suffix (kept out of the *.c/*.h gitignore via negation) so they are tracked and get C syntax highlighting.

Two GenC bugs surfaced:
- ~TailRecUnitNoExplicitEnd generates a non-terminating while(true) loop, so its execution test is ignored (nonTerminatingBenchmarks) to avoid hanging the suite.~
  Fixed in 94fd50b.
- ~RefInCtor and ImageProcessing have non-deterministic C output (Set-derived enum/union ordering and fresh-variable numbering), so they are excluded from the golden comparison (nonDeterministicBenchmarks).~
  Fixed in a2be173.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -114,6 +114,10 @@ project/project/
 *.c
 *.h
 *.out
+# ... but keep the GenC golden snapshots (named `.expected.c` / `.expected.h` for
+# syntax highlighting) under version control.
+!*.expected.c
+!*.expected.h
 
 # LZW GenC benchmarks files
 input.txt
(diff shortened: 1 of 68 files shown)


mbovel:
Failure in Bolts Lite seems unrelated to this PR:


```
Running 'stainless-dotty --config-file=stainless.conf.nightly ' on bolts project: data-structures/uarray...
$ find data-structures/uarray -name '*.scala' -exec stainless-dotty --config-file=stainless.conf.nightly  {} +
find: ‘data-structures/uarray’: No such file or directory
'stainless-dotty ' failed on bolts project: data-structures/uarray.
```

https://github.com/epfl-lara/stainless/actions/runs/28872446760/job/85638524873#step:11:3733

Trying to re-run.

samuelchassot (approved):
I haven't checked the expected c/h files by hand, but the rest is LGTM

PR #1761, opened 2026-07-07
Add a new UInt31 type in the library, add refinement types compilation option everywhere

samuelchassot:
This type represents a `BigInt` within the bounds [0, Int.MaxValue]. The plan is to add support to GenC later to generate `int32` in C for them.

At the same time, this PR adds the refinement type experimental flag to compiler options in the different places where needed to enable them in the std library.


samuelchassot:
TODO for a later PR

- [ ] Generate C code in GenC as `int32` for `UInt31`

PR #1762, opened 2026-07-07
Cleanup Main class and build.sbt to remove leftovers of the double frontend setup

samuelchassot:


mbovel (approved):
The Main class simplification looks excellent to me.

Not sure if [move common frontend to dotty as we have one frontend now](https://github.com/epfl-lara/stainless/pull/1762/changes/c2dea923e42bdf64739a7943e054d98d0cbe683d) is so useful, but why not!

samuelchassot:
> The Main class simplification looks excellent to me.
> 
> Not sure if [move common frontend to dotty as we have one frontend now](https://github.com/epfl-lara/stainless/pull/1762/changes/c2dea923e42bdf64739a7943e054d98d0cbe683d) is so useful, but why not!

mmhh, for me this is confusing to have both common and dotty packages even though we have only one now. Do you think we should keep both packages?

mbovel:
> mmhh, for me this is confusing to have both common and dotty packages even though we have only one now. Do you think we should keep both packages?

No it's fine. These tests are dotty specific anyway 👍

samuelchassot:
Alright so I am merging!

PR #1765, opened 2026-07-14
Enable old() in refinement types for function return types

samuelchassot:


vkuncak:
@samuelchassot In this program:
```scala
import stainless.lang._
object RefinementOld:
  case class Box(var x: BigInt)
  def addX(b: Box, y: BigInt): { res: Unit with b.x == old(b).x + y } =
    b.x = b.x + 0
```
If I run
```
stainless RefinementOld.scala --debug=verification,trees --debug-phases=AntiAliasing
```
it shows
```scala
[ Debug  ] def addX(b: Box, (y: BigInt) @pure): (b: Box, { param0: Unit | (b.x == b.x + y) }) = {
[ Debug  ]   var b: Box = b
[ Debug  ]   val res: { param0: Unit | (b.x == b.x + y) } = {
[ Debug  ]     b = Box(b.x + 0).asInstanceOf[Box]
[ Debug  ]     ()
[ Debug  ]   }.asInstanceOf[{ param0: Unit | (b.x == b.x + y) }]
[ Debug  ]   (b, res)
[ Debug  ] }
[ Debug  ] 
[ Debug  ] -------------Classes-------------
[ Debug  ] case class Box((x: BigInt) @var)
```

Then, verification performs the postcondition check **twice** and returns the counterexample twice.
I think the first check is asInstanceOf and the other one is the return parameter.
We surely want to avoid this, even if the check is cached. 
Should we insert some of those "do not verify as it's generated" assumptions?

vkuncak:
Another thing, @samuelchassot  it would be good to at least port `frontends/benchmarks/imperative/valid/Tutorial.scala` (the banking part) to use refinements with mutation, so we have a bit more involved example.

vkuncak:
In fact, @samuelchassot @kasiaMarek @mbovel the problem arises even without mutation. It looks like our type checker does not fully utilize refinement types as assumptions when generating verification conditions.

For example, this code without mutation:
```scala
object AssumingRefinement:
  case class Box(x: BigInt)
  def inc(): (Unit, {res: Box with res.x > 0}) =
    val b1 : {b2: Box with b2.x > 0} = 
      Box(-1).asInstanceOf[{b3: Box with b3.x > 0}]
    ((), b1)
```
generates three VCs with counterexamples:
```
frontends/benchmarks/refinement/valid/AssumingRefined.scala:4:7:   inc  type-checking     invalid           U:smt-z3  0.0
frontends/benchmarks/refinement/valid/AssumingRefined.scala:6:7:   inc  cast correctness  invalid           U:smt-z3  0.0
frontends/benchmarks/refinement/valid/AssumingRefined.scala:6:7:   inc  cast correctness  invalid           U:smt-z3  0.1
```

This already happens on the main branch before this PR, of course, as it has nothing to with mutation, but with type checking in the presence of refinement types (and pairs, apparently).

Once we have done a check in one place, subsequently we should be assuming this type, and thus later checks should be succeeding. Indeed, this is how it works for assertions already; given
```scala
  def test(x: Int) =
    assert(x > 0)
    assert(x > 0)
```
the first assertion fails, and subsequent one succeeds. That's why intermediate assertions can help verification; earlier ones become assumptions for later ones.

samuelchassot:
For the record, here are the trees for this code:
```scala
  case class Box(var x: BigInt) {
    def bump(): {res: Boolean with old(this).x + 1 == this.x} = {
      x = x + 1
      true
    }
  }

  // `old` on a mutated parameter, in the refinement of the return type
  def addX(b: Box, y: BigInt): { res: Unit with b.x == old(b).x + y } = {
    b.x = b.x + y
  }

  // Same postcondition, written with `ensuring`
  def addXEnsuring(b: Box, y: BigInt): Unit = {
    b.x = b.x + y
  }.ensuring(_ => b.x == old(b).x + y)

  // Callers can use the postcondition, in both styles
  def caller(): Unit = {
    val b = Box(10)
    addX(b, 5)
    assert(b.x == 15)
    addXEnsuring(b, 5)
    assert(b.x == 20)
  }

  // `old` on an immutable, unmutated parameter: old(t) == t
  case class Imm(i: BigInt)
  def immOld(t: Imm): { res: BigInt with res == old(t).i } = {
    t.i
  }
  
  def callerImm(): Unit = {
    val t = Imm(10)
    val temp = immOld(t)
    assert(immOld(t) == BigInt(10))
  }
```

```scala 
def callerImm: Unit = {
  val t: Imm = Imm(10)
  val temp: { param0: BigInt | (param0 == t.i) } = immOld(t)
  assert(immOld(t) == 10)
  ()
}

def addXEnsuring(b: Box, y: BigInt): (Box, Unit) = {
  val b: Box = b
  val b: Box = Box(b.x + y).asInstanceOf[Box]
  val tmp: Unit = ()
  val res: Unit = ()
  (b, res)
}.ensuring {
  (_$1: (Box, Unit)) => {
    val targetBound: Box = b
    _$1._1.x == targetBound.x + y
  }
}

def addX(b: Box, y: BigInt): (b: Box, { param0: Unit | (b.x == b.x + y) }) = {
  val b: Box = b
  val b: Box = Box(b.x + y).asInstanceOf[Box]
  val tmp: Unit = ()
  val res: { param0: Unit | (b.x == b.x + y) } = ().asInstanceOf[{ param0: Unit | (b.x == b.x + y) }]
  (b, res)
}

def bump(thiss: Box): (thiss: Box, { param0: Boolean | (thiss.x + 1 == thiss.x) }) = {
  val thiss: Box = thiss
  val thiss: Box = Box(thiss.x + 1).asInstanceOf[Box]
  val tmp: Unit = ()
  val tmp: Unit = ()
  val res: { param0: Boolean | (thiss.x + 1 == thiss.x) } = true.asInstanceOf[{ param0: Boolean | (thiss.x + 1 == thiss.x) }]
  (thiss, res)
}

def caller: Unit = {
  val b: Box = Box(10)
  val y$1: BigInt = 5
  val res: (b: Box, Unit) = addX(b, y$1)
  val b: Box = @DropVCs @DropVCs res._1.asInstanceOf[Box]
  val tmp: Unit = ()
  val b: Box = @DropVCs @DropVCs b.asInstanceOf[Box]
  val tmp: Unit = ()
  val tmp: Unit = res._2
  assert(b.x == 15)
  val res: (Box, Unit) = addXEnsuring(b, 5)
  val b: Box = @DropVCs @DropVCs res._1.asInstanceOf[Box]
  val tmp: Unit = ()
  val b: Box = @DropVCs @DropVCs b.asInstanceOf[Box]
  val tmp: Unit = ()
  val tmp: Unit = res._2
  assert(b.x == 20)
  ()
}

def immOld(t: Imm): { param0: BigInt | (param0 == t.i) } = t.i.asInstanceOf[{ param0: BigInt | (param0 == t.i) }]
```
In particular, note how the refinement type is not added at call site for `addX` because `b` is mutable (in `caller`). It is however added at call site for the `immOld` in `callerImm`, because the parameter is immutable. In this case, it is replaced correctly by the call site equivalent of `old(b)` i.e., `param0`.

PR #1769, opened 2026-08-07
Update inox and add regression test from issue 1768

samuelchassot:


samuelchassot:
@sankalpgambhir should we merge that?

sankalpgambhir:
Yes, the fix for this was merged earlier. It's also tested in the Inox semantics tests already.

samuelchassot:
> Yes, the fix for this was merged earlier. It's also tested in the Inox semantics tests already.

Okay so let me update the inox commit again to include the princess fix https://github.com/epfl-lara/inox/pull/267

samuelchassot:
once it passes, I'll merge!

PR #1770, opened 2026-08-18
Fix extraction of opaque type aliases used outside their defining scope

mbovel:
Extraction failed with `Unsupported call to + on v$4` on primitive calls whose receiver has an opaque type seen from outside its defining scope. This was previously unobservable (such calls don't typecheck in plain Scala) but qualified types make it reachable: a dependent qualifier typed inside the scope (e.g. `other: UInt32 with v + other <= max`) is instantiated at call sites outside it.

## Changes

- `extractType`: move the opaque-alias resolution case (`translucentSuperType`) above the `isAbstractOrAliasType` cases that shadowed it; `isAbstractOrAliasType` is also true for opaque aliases, but `widenDealiasKeepRefiningAnnots` does not unfold them outside their scope, so they were extracted symbolically and primitive dispatch fell through. Remove the now-redundant unreachable copy of that case.
- `extractType`: add the analogous case for *applied* (parameterized) opaque aliases, beta-reducing via `AppliedType.translucentSuperType`.
- `extractTypeDef`: extract bounded opaque types (`TypeBoundsTree` with a non-empty `alias`, incl. under `LambdaTypeTree`) as aliases instead of abstract types, and drop the dead `translucentSuperType` computation.
- New benchmarks: `OpaqueTypesRefinements` (escaping qualifier), `OpaqueTypes2` / `OpaqueTypes3` (bounded / parameterized opaque cross-scope), and `OpaqueTypes2Invalid`, which guards the soundness fix (a `valid` test would pass vacuously if the precondition VC disappears again).

## History

- The resolve-through-`translucentSuperType` behavior was introduced in b409ab5f4 ("Dealias opaque types to their translucent super type") and extended to type definitions in 7675362bf ("Fix handling of opaque type aliases").
- Both were silently disabled by the Scala 3 frontend port (4f748fb63): the type-member case became unreachable and the `extractTypeDef` handling became dead code.

### DIFF
--- a/frontends/benchmarks/dotty-specific/invalid/OpaqueTypes2Invalid.scala
+++ b/frontends/benchmarks/dotty-specific/invalid/OpaqueTypes2Invalid.scala
@@ -1,0 +1,19 @@
+object OpaqueTypes2Invalid {
+  opaque type Positive <: BigInt = BigInt
+  object Positive {
+    def apply(b: BigInt): Positive = {
+      require(b >= 0)
+      b
+    }
+  }
+}
+
+// Guards against the precondition VC silently disappearing: when the bounded
+// opaque type is extracted as an abstract type instead of an alias, the type
+// encoding boxes `apply`'s body and swallows its `require`, making this call
+// verify vacuously. Extracting the alias keeps the precondition, so this
+// violating call must be reported as invalid.
+object OpaqueTypes2InvalidUse {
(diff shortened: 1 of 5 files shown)


samuelchassot (approved):

samuelchassot:
Whoops I misclicked and it added a merge commit. Given the history, I think squashing is fine. WDYT?

PR #1774, opened 2026-08-20
Fix a bug refinement transformation when on a function parameter in a mutable type (issue #1775)

samuelchassot:
Fixes issue #1775


samuelchassot:
@mbovel wdyt of that?

samuelchassot:
The bug comes from the transformation performed by `AntiAliasing` replacing the mutated parameter in the whole.
When the mutated parameter appears in the refinement of another parameter, like in:
```scala
@mutable
case class Box(var value: BigInt)

def updateBox(b: Box with b.value >= 0, v: BigInt with v == b.value): {res: Unit with b.value >= 0} = {
  b.value = b.value + v
}
```

where the mutated parameter is `b` appears in the refinement of `v`.

The trees before `AntiAliasing` are:

```scala
def updateBox$0(
    b$0: { param0$0: Box$0 | (param0$0.value$10 >= 0) }, 
    (v$0: { param0$1: BigInt | (param0$1 == b$0.value$10) }) @pure
): { param0$5: Unit | (b$0.value$10 >= 0) } = 
    b$0.value$10 = (b$0.value$10 + v$0)
```

Then, `AntiAliasing` will replace the mutate `b$0` by a series of immutable `b$X`. The problem arises when this transformation is applied to the type of `v$0` in the body of the function. This makes the type different between `v$0` in the body and in the `fdparams`, which makes stainless crash because `v$0` in the body is `unbound`.

Trees after `AntiAliasing` are:

```scala
def updateBox$0(
    b$0: { param0$0: Box$0 | (param0$0.value$10 >= 0) }, 
    (v$0: { param0$1: BigInt | (param0$1 == b$0.value$10) }) @pure
): (b$1: { param0$0: Box$0 | (param0$0.value$10 >= 0) }, { param0$5: Unit | (b$1.value$10 >= 0) }) = 
{
  var b$2: { param0$0: Box$0 | (param0$0.value$10 >= 0) } = b$0
  val res$0: { param0$5: Unit | (b$2.value$10 >= 0) } = {
    b$2 = Box$0(b$2.value$10 + v$0).asInstanceOf[Box$0]
    ()
  }
  (b$2, res$0)
}
```

In these trees, the type of `v$0` is not shown, so here are some debug prints added to the error message to see the problem more clearly:

```
[info] [ Debug  ] Ensuring well-formedness after phase AntiAliasing...
[info] [DEBUG] occurrence v$0 has embedded type: { param0$8: BigInt | (param0$8 == b$2.value$10) }
[info] [DEBUG]   fd.params has v$0 with type:  { param0$1: BigInt | (param0$1 == b$0.value$10) }
[info] [ Error  ] Stainless terminated with an error.
[info] [ Error  ] Debug output is available in the file `stainless-stack-trace.txt`. If the crash is caused by Stainless, you may report your issue on https://github.com/epfl-lara/stainless/issues
[info] [ Error  ] You may use --debug=stack to have the stack trace displayed in the output.
```

and we can see that the type of `v$0` in the `fd.params` is `{ param0$1: BigInt | (param0$1 == b$0.value$10) }` where the `b` reference is indeed `b$0`, but in the body of the function it became `{ param0$8: BigInt | (param0$8 == b$2.value$10) }` where `b` is now `b$2`.

samuelchassot:
In fact, this happens with local `val` too, meaning the problem is more general:

```scala
  def updateBox2(b: Box with b.value >= 0): {res: Unit with b.value >= 0} = {
    val v: BigInt with v == b.value = b.value
    val x: BigInt with x == b.value = b.value
    b.value = b.value + v
    assert(v == x)
  }
```

This is related to the decision we made earlier to not apply the transformation for mutable variables in refinement types (cf Scala workshop presentation).

The fix should therefore be more general.

PR #1777, opened 2026-08-24
Floating-point measure

agilot:
Many scientific computing functions iterate until convergence using a floating-point stopping criterion. This PR adds floating-point measures to support such functions.


samuelchassot:
LGTM! @vkuncak any opinion?

vkuncak:
@agilot good point. Indeed, it seems < is indeed well founded on floats, including the weird values like NaN, +0, Infinity (for the weird cases x - 1 < x will become false and that should be detected). Could you add some tests that are rejected due to measure not being positive or not decreasing? It would be nice to illustrate the funny cases.

agilot:
Indeed. I have now added a few tests to illustrate what is rejected due to NaN and Infinities and what is accepted due the differences between floating-points and reals.

PR #1781, opened 2026-08-28
Run verification before GenC and add an option to skip verification

samuelchassot:


mbovel (approved):
Looks good to me.

samuelchassot:
> Looks good to me.

Thanks! once it passes I'll merge!

mbovel:
Fable 5 review:

# Review: epfl-lara/stainless #1781 — Run verification before GenC and add an option to skip verification

**Overall:** the design is sound — wrap `GenCRun` around a `VerificationRun`, gate C code generation on the verification report, and add `--skip-verification` as an escape hatch. The points below are roughly in order of importance.

## Main points

### 1. The tests disable the very feature being added

Both `GenCSuite` and `TailRecGenCSuite` now pass `--skip-verification` everywhere, so the new "verify before GenC" path has zero test coverage. Suggested additions:

- A test where a program with valid VCs goes through `--genc` **without** the flag.
- A test where a failing VC causes the `"Verification failed, aborting C code generation"` fatal error.

Without these, a regression in the gating logic would go unnoticed.

### 2. Error semantics changed from synchronous throw to failed Future

The old `GenCRun.apply` threw `FatalError` synchronously from the `try/catch`; now `fatalError` is called inside `Future.map`, so it surfaces as a failed future. The `an [inox.FatalError] should be thrownBy runMainWithArgs(...)` tests only keep passing if `runMainWithArgs` awaits and rethrows — presumably it does, but worth confirming, since the failure tests are exactly the ones exercising this path.

### 3. Redundant plumbing of the `skipped` flag

The flag now lives in three places:

- Read directly from `ctx.options` in `AbstractReport` (`Report.scala`).
- Read again in `VerificationAnalysis.toReport`.
- Stored as a constructor parameter on `VerificationReport`.

Since the relevant `VerificationReport` code paths have access to a context, the constructor param could likely be dropped in favor of reading the option (or pass it exactly once). As written there are two sources of truth that can diverge.

**Related bug:** `VerificationReport.parse` (the JSON cache path) doesn't restore `skipped`, so a report round-tripped through the cache silently loses the flag and the `(ALL VCs Skipped)` row.

### 4. Wasted work when skipping

With `--skip-verification`, `GenCRun.apply` still calls `verifRun.apply`, which runs the whole extraction pipeline just to return an empty analysis and emit a near-empty report. Short-circuiting in `GenCRun` (skip the `verifRun` call entirely when the flag is set) would be simpler and faster, and would remove the need for the empty-analysis branch in `VerificationComponent.execute` for this use case.

## Smaller points

- `FreshIdentifier("(ALL VCs Skipped)")` as a display row is a bit of a hack — the summary line already says "VCs skipped", so arguably emit no rows at all. If kept, consider `Level.Warning` rather than `Normal`; skipping verification is at least as noteworthy as admitting VCs, which gets red.
- In the skip branch of `VerificationComponent.execute`, `sources = functions0.toSet` (unfiltered), while every other branch filters `DropVCs` first — harmless but inconsistent.
- `--genc` verifies only the GenC-selected `ids` and their dependencies, which may be a narrower set than plain `--verification` would check on the same file. Probably intended, but worth a sentence in the docs/help text.
- The double extraction (once inside `verifRun`, once via `GenCRun.pipelineBegin`) roughly doubles frontend work on the happy path. Fine for now, but a comment acknowledging it would help the next reader.

## Verdict

Nothing blocking except the missing test coverage — as it stands, CI proves the old behavior still works with the flag but never exercises the headline change.

mbovel:
Simplifying as suggested in 3. looks like a good idea to me.

I think 1. is not a problem, but maybe we should still have a test without verification disabled?

Not sure about 2. and 4., might be worth checking.

PR #1782, opened 2026-08-28
Fix bug when adding an opaque type in the library

samuelchassot:


samuelchassot:
@mbovel can you have a look, I'm not sure to understand what's happening. I'll look deeper but your inputs would be valuable 👍

mbovel (changes requested):
This PR adds several special cases that seem overly specialized to me.

`v.symbol.info.classSymbol.isPackageObject` is incidental, but I don't think it would handle the problem generally.

Catching `UnsupportedCodeException` also seems dubious to me.

I am trying to make something simpler/more general, will post here with the result soon.

mbovel:
Closing in favor of #1793.

PR #1784, opened 2026-09-01
Add an array backed by List (with all operations using both BigInt and Int)

samuelchassot:
Adds a new datastructure to the library: an array specified by a list, with all operations duplicated to both accept `Int` and `BigInt` indices.
Also updates Inox to increase the grace timeout in potfolio.

This new structure is verified in a different library suite because it requires cvc5, z3, and Princess. Adding all of them to the current LibrarySuite slows everything down significantly. Therefore, we split the library just like we already did for `Float`s.


samuelchassot:
Results of my experiment with combined library or split, for the record:

```
  ┌────────────────────────────────────────────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────────────────────────┐
  │                                           Config                                           │                                        Result                                         │
  ├────────────────────────────────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────┤
  │ Split: LibrarySuite (z3-only, 30s) + AdvancedLibrarySuite (3 solvers + Bland, LArray only) │ ✅ 5m57s total (3m46s + 2m11s), all green                                             │
  ├────────────────────────────────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────┤
  │ Combined, 3 solvers + Bland                                                                │ ❌ crashes at 4m0s (recs is not empty in Core.convertPattern)                         │
  ├────────────────────────────────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────┤
  │ Combined, 3 solvers + OL                                                                   │ ❌ crashes at 51s (same bug — OL/Bland share the broken Core pattern-conversion code) │
  ├────────────────────────────────────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────┤
  │ Combined, 3 solvers + Vanilla (no lattice simplifier)                                      │ ❌ 15m14s, same 13 LArray VCs time out as the original bug report                     │
  └────────────────────────────────────────────────────────────────────────────────────────────┴───────────────────────────────────────────────────────────────────────────────────────┘
```

PR #1788, opened 2026-09-03
Update to 3.10.1-RC1-bin-20260903-e1f9361-NIGHTLY

mbovel:
Companion of epfl-lara/inox#269.

- Bump the Scala nightly in `build.sbt`, the packaging scripts and the sbt-plugin scripted tests, and point the inox submodule to the merged update.
- Adapt the Dotty frontend to API changes: `AbstractFile.absolute` and `SourceFile.apply(file, codec)` were removed.
- Discard `module-info.class` in the assembly merge strategy: the compiler now depends on upstream ASM 9.10.1 modules instead of scala-asm.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DZ2xybept98w22oE7cp3Hg

PR #1789, opened 2026-09-03
Update to 3.10.1-RC1-bin-20260903-e1f9361-NIGHTLY

mbovel:
Companion of epfl-lara/inox#269. Replaces #1788, which was opened from a fork and did not trigger CI.

- Bump the Scala nightly in `build.sbt`, the packaging scripts and the sbt-plugin scripted tests, and point the inox submodule to the merged update.
- Adapt the Dotty frontend to API changes: `AbstractFile.absolute` and `SourceFile.apply(file, codec)` were removed.
- Discard `module-info.class` in the assembly merge strategy: the compiler now depends on upstream ASM 9.10.1 modules instead of scala-asm.
- Fix the extraction of library Tasty units, which failed with "missing dependencies" errors:
  - The root tree of a Tasty unit is now wrapped in a `PackageDef` for the empty package with nested `PackageDef`s for the unit's packages. The filter that skips units under `scala` only looked at the outermost symbol and dropped every library unit; it now looks at the outermost non-empty package.
  - `->` is now an inline extension method defined directly in `Predef` (`ArrowAssoc` is deprecated). Calls from source files are inlined, but calls from Tasty units are plain method calls, which the tuple extractor did not recognize.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DZ2xybept98w22oE7cp3Hg


mbovel:
Note the nightly failed on this PR; I'll investigate.

samuelchassot:
> Note the nightly failed on this PR; I'll investigate.

I checked and it is unrelated. It's a problem we have with @sankalpgambhir with killing solvers in portfolio. This is a flaky behavior.

mbovel:
Okay, thanks for the insight!

Turns out daily now also fails, and these seem like real problems?

https://github.com/epfl-lara/stainless/actions/runs/33848121116/job/100965853912?pr=1789#step:11:1559

Issue #1522, opened 2024-05-08
Treat the `erased` soft keyword as Stainless  `@ghost` 


vkuncak:
Scala 3 has erased definitions: https://dotty.epfl.ch/docs/reference/experimental/erased-defs.html

We can presumably extract `erased` parameter keyword and map it to Stainless `@ghost`. 

First, introduce ErasedChecks  that has require, ensuring, assert, etc. all expecting `erased` parameters.
We need to keep StaticChecks for legacy reasons.

We should start by writing some code that uses erased instead of ghost and compiles with Scala 3 (even without Stainless). 

There are differences between erased, including the fact that `var`-s are possibly not marked by `erased`. Still, being able to remove immutable parameters of `def`s and case classes is already useful, as it allows to not rely on our own ghost code elimination.

vkuncak:
This example is accepted by Scala 3.5.0, so those `erased` locals are what we would like to accept as ghost.

```scala
import scala.language.experimental.erasedDefinitions

@scala.annotation.experimental
object ErasedTerms1 {

  def takeErased(x: BigInt)(erased y: BigInt): BigInt = {
    x + 1
  }

  def valid1(y: BigInt) = {
    erased val z: BigInt = y + 1
    takeErased(y)(z)
 }.ensuring { _ > y }

  def valid2(y: BigInt) = {
    val z: BigInt = y + 1
    takeErased(y)(z)
 }.ensuring { _ > y }
}
```

samuelchassot:
From my case studies, we need:

ghost vals and vars in code blocks:
```scala
def f() = {
  @ghost val x = ???
  @ghost var y = ???
}
```

ghost function parameters:

```scala
def f(x: BigInt, @ghost y: BigInt) = ???
```

ghost defs 
```scala
@ghost def f() = ???
```

class/case class parameters (vals and vars)
```scala
case class C(x: BigInt, @ghost y: BigInt, @ghost var z: BigInt)
```

samuelchassot:
First, we need to add `erased var` to Dotty. @kasiaMarek is doing it

samuelchassot:
Eventually, we will replace `@ghost` by `erased`. Therefore, `erased` must be extracted to the same internal representation of `ghost` so that the ghost checking machinery works for both.

PR #1790, opened 2026-09-04
Extract `erased` as ghost

samuelchassot:
Implements #1522

### DIFF
--- a/frontends/dotty/src/main/scala/stainless/frontends/dotc/FragmentChecker.scala
+++ b/frontends/dotty/src/main/scala/stainless/frontends/dotc/FragmentChecker.scala
@@ -97,10 +97,24 @@ class FragmentChecker(inoxCtx: inox.Context)(using override val dottyCtx: DottyC
 
     object GhostManagement {
       private val ghostAnnotation = getClassIfDefinedOrNone("stainless.annotation.ghost")
+      private val externAnnotation = getClassIfDefinedOrNone("stainless.annotation.extern")
       extension (sym: Symbol) {
-        private def hasGhostAnnotation(using DottyContext): Boolean = ghostAnnotation.exists(ghostClassSymbol => sym.hasAnnotation(ghostClassSymbol))
+        // An `erased` symbol (Scala's native soft keyword) is treated exactly like a `@ghost` symbol,
+        // mirroring the equivalence made when extracting flags in ASTExtractors.getAnnotations.
+        private def hasGhostAnnotation(using DottyContext): Boolean =
+          (sym.isEffectivelyErased && !(sym `is` Inline)) ||
+          ghostAnnotation.exists(ghostClassSymbol => sym.hasAnnotation(ghostClassSymbol))
         private def addGhostAnnotation()(using DottyContext): Unit = ghostAnnotation.foreach(ghostClassSymbol => sym.addAnnotation(ghostClassSymbol))
         private def removeGhostAnnotation()(using DottyContext): Unit = ghostAnnotation.foreach(ghostClassSymbol => sym.removeAnnotation(ghostClassSymbol))
+        // @extern code is not verified/extracted by Stainless, so ghost-access restrictions
+        // (which only make sense for code Stainless actually checks) should not apply inside it.
+        // Externally-erased symbols (e.g. macro-only stdlib methods like StringContext.f) are thus
+        // still usable from @extern code, same as from ghost code. Propagates from enclosing owners,
+        // mirroring how `propagatedAnnotations` propagates @extern during extraction.
+        private def hasExternAnnotation(using DottyContext): Boolean =
+          externAnnotation.exists(externClassSymbol =>
+            sym.hasAnnotation(externClassSymbol) || sym.ownersIterator.drop(1).exists(_.hasAnnotation(externClassSymbol))
+          )
       }
       /**
         * Tree traverser that propagates ghost annotations on synthesized members:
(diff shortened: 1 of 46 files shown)


frontends/dotty/src/main/scala/stainless/frontends/dotc/FragmentChecker.scala
```diff
@@ -97,10 +97,28 @@ class FragmentChecker(inoxCtx: inox.Context)(using override val dottyCtx: DottyC
...
         private def addGhostAnnotation()(using DottyContext): Unit = ghostAnnotation.foreach(ghostClassSymbol => sym.addAnnotation(ghostClassSymbol))
         private def removeGhostAnnotation()(using DottyContext): Unit = ghostAnnotation.foreach(ghostClassSymbol => sym.removeAnnotation(ghostClassSymbol))
+        // @extern code is not verified/extracted by Stainless, so ghost-access restrictions
+        // (which only make sense for code Stainless actually checks) should not apply inside it.
+        // Externally-erased symbols (e.g. macro-only stdlib methods like StringContext.f) are thus
+        // still usable from @extern code, same as from ghost code. Propagates from enclosing owners,
+        // mirroring how `propagatedAnnotations` propagates @extern during extraction.
+        private def hasExternAnnotation(using DottyContext): Boolean =
+          externAnnotation.exists(externClassSymbol =>
+            sym.hasAnnotation(externClassSymbol) || sym.ownersIterator.drop(1).exists(_.hasAnnotation(externClassSymbol))
```
mbovel: Why `drop(1)`? Does that drop the current symbol? If so couldn't we just write `sym.ownersIterator.exists(_.hasAnnotation(externClassSymbol))` instead of the disjunction?

frontends/dotty/src/main/scala/stainless/frontends/dotc/FragmentChecker.scala
```diff
@@ -276,8 +294,10 @@ class FragmentChecker(inoxCtx: inox.Context)(using override val dottyCtx: DottyC
             case m: tpd.MemberDef  =>
               // We consider some synthetic methods values as being inside ghost
               // but don't auto-annotate as such because we don't want all code to be removed.
-              // They are synthetic case class methods that are harmless if they see some ghost nulls
-              if (m.symbol.hasGhostAnnotation || effectivelyGhost(sym))
+              // They are synthetic case class methods that are harmless if they see some ghost nulls.
+              // @extern members are not verified or extracted by Stainless, so ghost-access
+              // restrictions (meant for code Stainless actually checks) don't apply inside them either.
+              if (m.symbol.hasGhostAnnotation || effectivelyGhost(sym) || m.symbol.hasExternAnnotation)
```
mbovel: Why is that needed now and wasn't before?

mbovel (commented):
Apart from my 2 questions, looks great to me! 👍

mbovel:
And I don't think copilot's review is relevant. The change in `PackageDef` is linked to a change in Dotty, so it should be fine.

frontends/dotty/src/main/scala/stainless/frontends/dotc/FragmentChecker.scala
samuelchassot: I showed up in an extern main that was using `println(f"asdf")` and apparently `f"..."` is implemented with `erased` so this fails our ghost checks. As it is in extern anyway, this should be the responsibility of the programmer right?

mbovel: Yes. I am just wondering if this is the right place.

There is already:

https://github.com/epfl-lara/stainless/blob/e070b73a5e324c173820741272ba21e6e048b0e6/frontends/dotty/src/main/scala/stainless/frontends/dotc/FragmentChecker.scala#L666-L677

Called from:

https://github.com/epfl-lara/stainless/blob/e070b73a5e324c173820741272ba21e6e048b0e6/frontends/dotty/src/main/scala/stainless/frontends/dotc/FragmentChecker.scala#L541-L547

Could we and should we instead modify this function to walk the ancestors as well?

frontends/dotty/src/main/scala/stainless/frontends/dotc/FragmentChecker.scala
samuelchassot: good point! I removed it!

samuelchassot:
@mbovel wdyt?

mbovel (approved):
Looks good, I like that the "ignore symbol" logic is now shared.

frontends/dotty/src/main/scala/stainless/frontends/dotc/FragmentChecker.scala
```diff
@@ -351,6 +360,24 @@ class FragmentChecker(inoxCtx: inox.Context)(using override val dottyCtx: DottyC
       }
     }
 
+
+  def ignoredSymbols(sym: Symbol, additionalCases: Boolean = false): Boolean = {
```
mbovel: Minor: I'd call it `ignoredSymbol` (singular) or even `isIgnoredSymbol`.

samuelchassot: sounds good!

samuelchassot: done

PR #1793, opened 2026-09-07
Substitute the inliner's opaque proxies in inlined code

mbovel:
Another attempt at fixing the bug #1782 tries to fix.

It's not exactly simpler than #1782, but I think it's closer to being the right thing. It substitute proxies generated by Dotty in inlined code back to their definitions.

| Case | Shape | main | #1782 | this PR |
|---|---|---|---|---|
| `MonadicTry2` | library `Try.flatMap`, `opaque type` in `package object lang` | n/a (no opaque in library) | 48/48 valid | 48/48 valid |
| `V1` | same as the library, in `object Lib` | crash `IndexOutOfBoundsException: -1` | crash `IndexOutOfBoundsException: -1` | 17/17 valid |
| `V2` | `inline def`s directly in `object Refined`, `this.h(x)`, opaque param/result | crash `IndexOutOfBoundsException: -1` | crash `IndexOutOfBoundsException: -1` | 2/2 valid |
| `V3` | `opaque type Id[A] = A`, `inline def` with a lambda and a by-name argument | `Stainless does not support type [A] =>> A` | `missing dependencies: Method Lib3, Class Lib3` | 1/1 valid |
| `V4` | `package object` with `opaque type`, `inline def viaThis(v) = this.helper(v)` | `Stainless does not support type [A] =>> A` | `missing dependencies: Method package$_this` | 1/1 valid |
| `V5` | nested inline calls across two objects with opaque types | crash `IndexOutOfBoundsException: -1` | crash `IndexOutOfBoundsException: -1` | 1/1 valid |

## Detailed description

When the enclosing object `O` of an inlined method declares an opaque type alias, Dotty's inliner does not elide `O.this` (`Inliner.canElideThis`) and instead binds the object to a proxy exposing the aliases (`Inliner.addOpaqueProxies`):

    val $proxy1: O.type{type T = R} = O.$asInstanceOf[O.type{type T = R}]
    val O$_this: ($proxy1 : O.type{type T = R}) = $proxy1

Types in the inlined code are then expressed relative to these proxies, members of `O` are selected on them, and value arguments whose type mentions `O` are cast to `A & A[O := $proxy1]`. None of this can be extracted as is: the proxy bindings have refined object types, and the intersections are unsupported. This affects any object mixing `opaque type` and `inline def`, and crashed extraction with an `IndexOutOfBoundsException` in the `RefinedType` case.

Objects have no instances in Stainless, so at each `Inlined` node the proxies (recognised with `Inliner.OpaqueProxy`) are substituted by `O` itself with a `TreeTypeMap`, their bindings are dropped and the redundant intersections are simplified with `glb`, mirroring what the inliner does for the result type (`Inliner.mapBackToOpaques`).

`ExCall` now looks through the binding-less `Inlined` markers that wrap references coming from the call site when testing whether a receiver is a module, and the `RefinedType` case reports an error instead of crashing when the refined member is not a type parameter of the parent.

### DIFF
--- a/frontends/benchmarks/dotty-specific/valid/OpaqueTypesInline1.scala
+++ b/frontends/benchmarks/dotty-specific/valid/OpaqueTypesInline1.scala
@@ -1,0 +1,33 @@
+import stainless.lang._
+
+// When the enclosing object of an inlined method declares an opaque type alias,
+// Dotty's inliner materializes that object as local proxies in the inlined code
+// (`val $proxy1: Lib.type{type Positive = BigInt} = Lib.$asInstanceOf[...]`,
+// `val Lib$_this = $proxy1`) and casts value arguments to `A & A[Lib := $proxy1]`.
+// This is the shape of `Try.flatMap` in the Stainless library, in a user object.
+object OpaqueTypesInline1 {
+  object Lib {
+    opaque type Positive = BigInt
+
+    sealed abstract class Try[T] {
+      inline def flatMap[U](f: T => Try[U]): Try[U] = this match {
+        case Success(t) => f(t)
+        case Failure(exc) => Failure(exc)
+      }
+    }
+    case class Success[T](t: T) extends Try[T]
+    case class Failure[T](exc: BigInt) extends Try[T]
+  }
+
+  import Lib._
+
+  def g(x: Try[BigInt]): Try[BigInt] = {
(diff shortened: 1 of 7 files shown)


samuelchassot (approved):
That looks sensible to me. I don't have any comments!

